PR #113, opened 2015-07-15
Updates toolkit to use 10.2.6

anttikajanus:
@rexhansen please review.


anttikajanus:
@rexhansen removing since it seems you already merged changes

Issue #123, opened 2016-03-13
Support for UWP


dimesv:
I have included a new project for UWP.
Repo here: 
https://github.com/dimesv/arcgis-toolkit-dotnet

dotMorten:
Thank you for this amazing piece of work. Unfortunately the .NET SDK currently does not fully support UWP (there are several known compatibility issues), so we can't pull this in. Also you don't really need to rebuild the toolkit for UWP - you can continue to use the 8.1 SDK in your 10.0 app (but beware of the compat issues).

We'll have full support for UWP with the Quartz release, and a toolkit will be provided with it.

Thank you for your hard work. I hate to have to reject it, but keep up the good work! Please keep your repo around though - some might find it handy, and if anyone asks for something like this, we'll make sure to point to your fork.

dimesv:
Hi Morten,

Thanks for your reply.
I knew that you were going to reject it due to the compability issues but I was just trying to make sure that I'm doing things correctly including copyrights.

I only modified the Toolkit because I didn't knew you were going to release a new one so this approach seemed right at the time. It would have allowed me to change the reference to the Quartz Runtime once it was released and have a Toolkit running UWP. Since a new Toolkit will come out then this does not make much sense now.

In regards to the viewer is essentially the same. I don't know if a new one will come out so I am hoping that just changing the reference to the Quartz Runtime and Toolkit will do the trick.

Thanks again for your comment.

Cheers,
Jose

PR #124, opened 2016-03-13
New project for UWP

dimesv:
I have created a new project for UWP to support the viewer: 
https://github.com/dimesv/arcgis-portalviewer-dotnet

This addresses issue: #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # arcgis-toolkit-dotnet
 
+Includes a new project for UWP.
+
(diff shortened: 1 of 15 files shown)


dotMorten:
Thank you for this amazing piece of work. Unfortunately the .NET SDK currently does not fully support UWP (there are several known compatibility issues), so we can't pull this in. Also you don't really need to rebuild the toolkit for UWP - you can continue to use the 8.1 SDK in your 10.0 app (but beware of the compat issues).

We'll have full support for UWP with the Quartz release, and a toolkit will be provided with it.

Thank you for your hard work. I hate to have to reject it, but keep up the good work!

PR #136, opened 2017-08-02
TableOfContents fixes - See comments with JH:

ramtechjoe:
The Layer order was backwards - bases layer in bottom and opposite order added
Removing layers threw an exception


ramtechjoe:
At this time, the ObservableLayerContentList  is only used by the TableOfContent control so I think that it is valid to change there.

I do agree that changing in the TableOfContents control would be a preferred approach.  However, this is not possible because the ObservableLayerContentList is an IReadOnlyList and adding/removing is done by responding to the Layer events in the map.  Unless I am missing something, I do not see a way I could change the behavior without changing ObservableLayerContentList.  Adding behavior to order the list based on a dependency property night be a nice approach, but this still does require changes to the ObservableLayerContentList.

I will put in PR for the bug when I have time, but unfortunately, we will have to use our own one off of the TOC to provide the required behavior

dotMorten:
Closing this PR, since it's pointing to an old branch. We have some major rework coming for the ToC though, and part of that design is reversing layer ordering.

PR #137, opened 2017-08-03
iOS Scalebar

zwaap:
Includes changes to add a Xamarin iOS version of the scalebar.  Includes:

- New head project for building a Xamarin iOS Toolkit assembly
- Compiler conditionals and logic in ScaleLine.cs to cross-compile on Xamarin iOS
- iOS-specific layout logic (in ScaleLine.iOS.cs)
- Compiler conditionals in several shared classes to opt out of compiling them on Xamarin platforms

The API surface and implementation parallels that for Windows platforms.  The view has a default width of 200, which can be controlled via the `TargetWidth` property.  Design-time support is also included.


src/Esri.ArcGISRuntime.Toolkit/iOS/Esri.ArcGISRuntime.Toolkit.iOS/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Esri.ArcGISRuntime" version="100.1.0" targetFramework="xamarinios10" />
+  <package id="Esri.ArcGISRuntime.Xamarin.iOS" version="100.1.0" targetFramework="xamarinios10" />
```
dotMorten: I think the first one should be enough?

src/Esri.ArcGISRuntime.Toolkit/iOS/Esri.ArcGISRuntime.Toolkit.iOS/UI/RectangleView.cs
```diff
@@ -0,0 +1,80 @@
...
+                InvalidateIntrinsicContentSize();
+                OnPropertyChanged();
+            }
+        }
+
+        private CGSize _size = CGSize.Empty;
+
+        public override CGSize IntrinsicContentSize => _size;
+
+        /// <inheritdoc />
```
dotMorten: Inheritdoc on interfaces must include a cref: `cref="INotifyPropertyChanged.PropertyChanged"`

src/Esri.ArcGISRuntime.Toolkit/iOS/Esri.ArcGISRuntime.Toolkit.iOS/UI/Controls/ScaleLine.iOS.cs
```diff
@@ -0,0 +1,258 @@
...
+        private RectangleView _firstUsTickLine;
+        private RectangleView _secondUsTickLine;
+        private RectangleView _combinedScaleLine;
+
+#pragma warning disable SA1642 // Constructor summary documentation must begin with standard text
+        /// <summary>
+        /// Internal use only.  Invoked by the Xamarin iOS designer.
+        /// </summary>
+        /// <param name="handle">A platform-specific type that is used to represent a pointer or a handle.</param>
+#pragma warning restore SA1642 // Constructor summary documentation must begin with standard text
```
dotMorten: Hide from intellisense?

src/Esri.ArcGISRuntime.Toolkit/iOS/Esri.ArcGISRuntime.Toolkit.iOS/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,52 @@
...
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
dotMorten: Version should be 100.1.Perhaps we should move this to a buildnum.cs for all the platforms.

src/Esri.ArcGISRuntime.Toolkit/Shared/UI/Controls/ScaleLine/ScaleLine.cs
```diff
@@ -32,23 +37,34 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
...
         private TextBlock _usUnit;
         private TextBlock _metricValue;
         private TextBlock _metricUnit;
         private Rectangle _metricScaleLine;
         private Rectangle _usScaleLine;
+        private static double _dpi =
+#if __IOS__
+            96 * UIKit.UIScreen.MainScreen.Scale;
+#else
+            96;
```
dotMorten: Always 96. The value is calculated based on a dips value already, so this doesn't change with screenscale. Probably just revert all _dpi changes back to the fixed value.

zwaap:
@dotMorten please have another look.  I've addressed all your feedback items.  Additionally, as discussed, I eliminated nearly all the compiler conditionals in the shared `ScaleLine.cs` file by moving platform-specific logic into partial classes.  Changes for that specifically include:

- One new partial class file for Xamarin platforms (`ScaleLine.Xamarin.cs`) and one for Windows (`ScaleLine.Windows.cs`)
- One new shared project for code that is shared only between Xamarin platforms (`Esri.ArcGISRuntime.Toolkit.Xamarin.Shared`) and one for code shared between Windows platforms (`Esri.ArcGISRuntime.Toolkit.Windows.Shared`)
  - Alternatives here would have been to keep the files in the master shared project or link these files directly into the head projects.  I thought quite a bit about this, and in the end came to think that having dedicated shared projects is really the best approach.  Due mainly to having DependencyProperties on Windows platforms and not Xamarin (native), it's going to be very common to have swaths of code that applies just to UWP and WPF or just to iOS and Android.
- Moved the "master" shared project (`Esri.ArcGISRuntime.Toolkit.Shared`) from the root of the `Shared` folder into `Shared\All`.  So the `Shared` folder now has three sub-folders - `All`, `Windows`, and `Xamarin`.
- Public cross-platform properties still have their signatures defined in the shared `ScaleLine.cs` file, but the getters and setters redirect to private platform-specific implementations, like so:

    ```cs
        /// <summary>
        /// Gets or sets the width that will be used to
        /// calculate the length of the ScaleLine
        /// </summary>
        public double TargetWidth
        {
            get => TargetWidthImpl;
            set => TargetWidthImpl = value;
        }
    ```

  - This seemed the best way of ensuring that the API surface and XML summary doc stays in sync across all the platforms.
  - I've used a naming convention of `<MemberName>Impl` for the platform-specific implementations.  I also considered `<MemberName>Internal` or similar.  Definitely open to suggestions.

src/Esri.ArcGISRuntime.Toolkit/iOS/Esri.ArcGISRuntime.Toolkit.iOS/packages.config
zwaap: Surprisingly, no.  When you install the Esri.ArcGISRuntime package, the Esri.ArcGISRuntime.Xamarin.iOS reference also gets added.  I tested building with that removed, and it fails.

zwaap:
@dotMorten please have another look.  The code-sharing approach has platform-specific code moved into partial classes.  I've put the partial classes into the main shared project and done away with the Windows and Xamarin shared projects.

PR #144, opened 2017-09-12
Scalebar: Add support for Android and Xamarin Forms

zwaap:
Adds implementations of the `ScaleLine` component for Xamarin Android and Xamarin Forms.  Changes include:

- Addition of Xamarin Android, Xamarin Forms iOS, Xamarin Forms Android, Xamarin Forms UWP, and Xamarin Forms shared toolkit projects
- Xamarin Android implementation of `ScaleLine` logic
- Xamarin Forms implementation of `ScaleLine` that leverages native implementations
- Tweaks to native Xamarin iOS `ScaleLine` implementation to support Forms

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,6 @@ project.lock.json
 
 # UWP auto-generated NuGet file
 *.nuget.targets
+
+# Xamarin Android auto-generated resource file
+*Resource.Designer.cs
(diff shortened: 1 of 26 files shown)


zwaap:
> At some point we need to update nuget generation for these new targets as well

Good point.  I'll drop an issue in for it to make sure it doesn't get overlooked.

PR #147, opened 2017-09-28
VS2017 Upgrade: Downgrade Android target framework

zwaap:
Downgrades Android target framework to `monoandroid60` and adds ignore rule for auto-generated VS2017 file location.


zwaap:
@dotMorten please go ahead and merge into your branch if you want these tweaks.

PR #159, opened 2018-01-30
Time Slider: Add UWP implementation

zwaap:
Adds the TimeSlider control on UWP


zwaap:
> might want to consider getting rid of the data string converter, and just use function x-binding instead

@dotMorten looks like that's not possible in a control template.  Build fails with the following:



I wasn't able to find an effective workaround.  Let me know if you know of any tricks to get this working and I'll update it.

PR #170, opened 2018-03-02
Added a simple Xamarin.Forms sample app

dotMorten:
FYI @jNery @zwaap


src/Samples/Toolkit.Samples.Forms/Toolkit.Samples.Forms.iOS/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Xamarin.Forms" version="2.4.0.266-pre1" targetFramework="xamarinios10" />
```
jNery: Could this be causing this error? `{System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.TypeInitializationException: The type initializer for 'Esri.ArcGISRuntime.Mapping.Basemap' threw an exception. ---> System.EntryPointNotFoundException: CoreRT_ArcGISRuntimeEnvironment_setInstallDirectory`

Do we need to match API Xamarin.Forms reference 2.4.0.74863? I think I had that problem in my sample before too.

PR #232, opened 2018-08-24
Android Time Slider: Rename .targets file to match package ID

zwaap:
Changes the name of the `Esri.ArcGISRuntime.Android.Toolkit.targets` file to `Esri.ArcGISRuntime.Toolkit.targets` to match the NuGet package's ID.  This is a requirement for the targets file to be added when the NuGet package is referenced.  The file has also been moved into a directory named `build` so it's not at the root level of the project.


zwaap:
> Should it be moved to an Android subfolder below \build\ then?

Sure, would make sense for it to be in case we ever want to add one for another platform.  Done.

Issue #277, opened 2019-10-04
Compass clicked Event


kilimnik:
It would be nice to have an event that is called wyhen the compass gets clicked.
And also an option to disable the resetting rotation when it is clicked.

dotMorten:
Thank you for this and the PR.

Since this is introducing new APIs we'd have to go through an API review with the other Toolkits first. To help that, could you describe some real-world scenarios where this is needed:

> It would be nice to have an event that is called when the compass gets clicked.

Also things to consider: Resetting rotation only happens if you bind it to the geoview. You can instead bind the heading to the MapView's MapRotation property and it wouldn't be able to reset (as it doesn't have a mapview to act on). I admit it's a little subtle difference, and not obvious, but the ability is there today. On WPF and UWP you can also set `IsHitTestVisible="false"` to disable any clicking.

Wrt a clicked event, you can use the built-in platform events for detecting clicks as well (like MouseDown/Tapped etc).

DK99:
First of all thanks for the quick response.

> could you describe some real-world scenarios where this is needed:

I am currently working on an App where I wanty to give the user a choice setting the heading of the Map fixed to the north or relative to the heading of the phone. And I thought the compass would be a great way to represent this toggeling. The same functionality is also used in the GoogleMaps App.

> Resetting rotation only happens if you bind it to the geoview

I haven't thought about that but thats a great idea.

> you can use the built-in platform events for detecting clicks

I have to test this. If so you didn't need my proposed PR.

DK99:
I was testing your proposal to use the built-in platform events for detecting clicks but I think that wouldn't work in Xamarin.Forms

dotMorten:
With .NET MAUI you should be able to use the gesture recognizers.

PR #278, opened 2019-10-04
Compass clicked Event

kilimnik:
An event that fires when the Compass is pressed.
See Issue #277.

### DIFF
--- a/src/Toolkit.Forms/Compass/Compass.cs
+++ b/src/Toolkit.Forms/Compass/Compass.cs
@@ -31,10 +31,32 @@ namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
         /// </summary>
         public Compass()
         {
+            ResetOnClick = true;
+
             HorizontalOptions = LayoutOptions.End;
             VerticalOptions = LayoutOptions.Start;
             WidthRequest = 30;
             HeightRequest = 30;
+        }
+
+        /// <summary>
+        /// Gets called, when the Compass is clicked
+        /// </summary>
+        public event EventHandler CompassClicked;
+
+        /// <summary>
+        /// Identifies the <see cref="ResetOnClick"/> bindable property.
+        /// </summary>
+        public static readonly BindableProperty ResetOnClickProperty =
+            BindableProperty.Create(nameof(ResetOnClick), typeof(bool), typeof(Compass), false, BindingMode.OneWay, null);
+
+        /// <summary>
+        /// Enables or disables the rotation of the <see cref="GeoView"/> if the compass is clicked.
+        /// </summary>
+        public bool ResetOnClick
+        {
+            get { return (bool)GetValue(ResetOnClickProperty); }
+            set { SetValue(ResetOnClickProperty, value); }
         }
 
         /// <summary>
@@ -82,5 +104,10 @@ namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
         /// </summary>
         public static readonly BindableProperty GeoViewProperty =
             BindableProperty.Create(nameof(Compass.GeoView), typeof(GeoView), typeof(Compass), null, BindingMode.OneWay, null);
+
+        public void RaiseCompassClickedEvent()
+        {
+            CompassClicked?.Invoke(this, EventArgs.Empty);
+        }
     }
 }
--- a/src/Toolkit.Forms/Compass/Compass.cs
+++ b/src/Toolkit.Forms/Compass/Compass.cs
@@ -105,6 +105,9 @@ namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
         public static readonly BindableProperty GeoViewProperty =
             BindableProperty.Create(nameof(Compass.GeoView), typeof(GeoView), typeof(Compass), null, BindingMode.OneWay, null);
 
+        /// <summary>
+        /// Raises the <see cref="CompassClicked"/> event
+        /// </summary>
         public void RaiseCompassClickedEvent()
         {
             CompassClicked?.Invoke(this, EventArgs.Empty);
--- a/src/Toolkit/Toolkit.Android/UI/Controls/Compass/Compass.Android.cs
+++ b/src/Toolkit/Toolkit.Android/UI/Controls/Compass/Compass.Android.cs
@@ -14,6 +14,7 @@
 //  *   limitations under the License.
 //  ******************************************************************************/
 
+using System;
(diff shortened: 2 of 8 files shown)


src/Toolkit.Forms/Compass/Compass.cs
```diff
@@ -31,12 +31,34 @@ public class Compass : View
         /// </summary>
         public Compass()
         {
+            ResetOnClick = true;
```
dotMorten: Set to true in the dependency property instead

src/Toolkit.Forms/Compass/Compass.cs
```diff
@@ -31,12 +31,34 @@ public class Compass : View
...
+        /// <summary>
+        /// Gets called, when the Compass is clicked
+        /// </summary>
+        public event EventHandler CompassClicked;
+
+        /// <summary>
+        /// Identifies the <see cref="ResetOnClick"/> bindable property.
+        /// </summary>
+        public static readonly BindableProperty ResetOnClickProperty =
+            BindableProperty.Create(nameof(ResetOnClick), typeof(bool), typeof(Compass), false, BindingMode.OneWay, null);
```
dotMorten: Default value should be true

src/Toolkit.Forms/Compass/Compass.cs
```diff
@@ -82,5 +104,13 @@ public GeoView GeoView
...
         public static readonly BindableProperty GeoViewProperty =
             BindableProperty.Create(nameof(Compass.GeoView), typeof(GeoView), typeof(Compass), null, BindingMode.OneWay, null);
+
+        /// <summary>
+        /// Raises the <see cref="CompassClicked"/> event
+        /// </summary>
+        public void RaiseCompassClickedEvent()
+        {
+            CompassClicked?.Invoke(this, EventArgs.Empty);
+        }
```
dotMorten: This shouldn't be public

src/Toolkit/Toolkit.Android/UI/Controls/Compass/Compass.Android.cs
```diff
@@ -36,16 +37,24 @@ public partial class Compass
         /// Initializes a new instance of the <see cref="Compass"/> class.
         /// </summary>
         /// <param name="context">The Context the view is running in, through which it can access resources, themes, etc</param>
-        public Compass(Context context)
-            : base(context) => Initialize();
+        public Compass(Context context) : base(context)
+        {
+            ResetOnClick = true;
```
dotMorten: Initialize `_resetOnClick` to true instead

src/Toolkit/Toolkit.iOS/UI/Controls/Compass/Compass.iOS.cs
```diff
@@ -74,7 +74,7 @@ private void Initialize()
 
         private void OnTapped()
         {
-            ResetRotation();
+            CompassClicked?.Invoke(this, EventArgs.Empty);
```
dotMorten: This should still call ResetRotation. Perhaps better to rename ResetRotation to a common internal "OnCompassClicked" method, that resets (if enabled) and raises the public event, rather than subscribing to an event this control already owns (that also avoids raising events when no one externally is listening)

src/Toolkit/Toolkit/UI/Controls/Compass/Compass.Windows.cs
```diff
@@ -56,17 +56,55 @@ public override void OnApplyTemplate()
...
+                {
+
+                    compass.CompassClicked += compass.OnResetRotation;
+                }
+                else
+                {
+                    compass.CompassClicked -= compass.OnResetRotation;
+                }
+            }
+        }
```
dotMorten: I don't think this is necessary if we don't rely on an event but just use a handler (that in turn raises a public event).

src/Toolkit/Toolkit/UI/Controls/Compass/Compass.Xamarin.cs
```diff
@@ -24,6 +24,31 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
...
+                {
+                    if (value)
+                    {
+
+                        CompassClicked += OnResetRotation;
+                    }
+                    else
+                    {
+                        CompassClicked -= OnResetRotation;
+                    }
```
dotMorten: Shouldn't be necessary with a common handler

src/Toolkit/Toolkit/UI/Controls/Compass/Compass.Xamarin.cs
```diff
@@ -24,6 +24,31 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
...
+
+                    _resetOnClick = value;
+                    OnPropertyChanged(nameof(ResetOnClick));
```
dotMorten: Only raise this for ios/android (we don't raise INPC for dependency properties)

src/Toolkit/Toolkit/UI/Controls/Compass/Compass.cs
```diff
@@ -44,10 +45,25 @@ public partial class Compass : Control
...
         {
+            ResetOnClick = true;
             Initialize();
         }
 #endif
 
+        /// <summary>
+        /// Gets called, when the Compass is clicked
+        /// </summary>
+        public event EventHandler CompassClicked;
```
dotMorten: Should it just be `Clicked` ?
Also should it be a cancellable RoutedEvent (and if cancelled won't trigger reset)

src/Toolkit/Toolkit/UI/Controls/Compass/Compass.cs
```diff
@@ -103,7 +119,12 @@ private void UpdateCompassFromGeoView(GeoView view)
             _headingSetByGeoView = false;
         }
 
-        private void ResetRotation()
+        /// <summary>
+        /// The method called when the Compass gets clicked and the <see cref="ResetOnClick"/> property is set to true
+        /// </summary>
+        /// <param name="sender">Sender of the event</param>
+        /// <param name="args">Arguments of the event</param>
+        public void OnResetRotation(object sender, EventArgs args)
```
dotMorten: Shouldn't be public

mbranscomb:
@kilimnik Closing this PR due to lack of activity. We appreciate your contribution - feel free to re-open this PR if you would like to follow up on the review feedback and see this feature included in the Toolkit.

Thanks

PR #281, opened 2019-11-06
[Bug Fix] Fix OriginCamera for Xamarin.Forms AR

ZackAllen:
Change made to ARSceneViewRenderer.cs to make sure the OriginCamera property is updated.
ARSceneView.cs has been updated to prevent an event cycle when updating the camera.

I have tested this on Xamarin.iOS, Xamarin.Android, and Xamarin.Forms.


src/ARToolkit/ARSceneView.cs
```diff
@@ -253,8 +253,10 @@ public Mapping.Camera OriginCamera
             get => _controller?.OriginCamera ?? throw new InvalidOperationException();  // Only null in design mode
             set
             {
-                if (_controller != null) 
+                if (_controller != null && !value.IsEqual(OriginCamera))
```
dotMorten: Can't you just use `(value != OriginCamera)` ?
We're not as concerned about equality as we are same reference.

PR #282, opened 2019-11-20
Scene clipping distance implementation

nCastle1:
Moves to 100.7 Runtime and adds support for scene clipping distance.

### DIFF
--- a/src/ARToolkit.Forms/ARSceneView.cs
+++ b/src/ARToolkit.Forms/ARSceneView.cs
@@ -37,6 +37,25 @@ namespace Esri.ArcGISRuntime.ARToolkit.Forms
         }
 
         /// <summary>
+        /// Identifies the <see cref="ClippingDistance"/> bindable property.
+        /// </summary>
+        public static readonly BindableProperty ClippingDistanceProperty =
+            BindableProperty.Create(nameof(ClippingDistance), typeof(double), typeof(ARSceneView), 0.0, BindingMode.OneWay, null);
+
+        /// <summary>
+        /// Gets or sets the clipping distance from the origin camera, beyond which data will not be displayed.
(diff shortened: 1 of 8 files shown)


src/ARToolkit.Forms/ARSceneView.cs
```diff
@@ -36,6 +36,25 @@ public ARSceneView()
             OriginCamera = new Mapping.Camera(0, 0, 15E6, 0, 0, 0);
         }
 
+        /// <summary>
+        /// Identifies the <see cref="ClippingDistance"/> bindable property.
+        /// </summary>
+        public static readonly BindableProperty ClippingDistanceProperty =
+            BindableProperty.Create(nameof(ClippingDistance), typeof(double), typeof(ARSceneView), 0.0, BindingMode.OneWay, null);
```
dotMorten: Confirming: Is the default 0.0 and not NaN or something else?

src/Directory.Build.props
```diff
@@ -35,7 +35,7 @@
     <UWPTargetFramework>uap10.0.16299</UWPTargetFramework>
     <NetCoreTargetFramework>netcoreapp3.0</NetCoreTargetFramework>
 
-    <ArcGISRuntimePackageVersion Condition="'$(ArcGISRuntimePackageVersion)'==''">100.6.0</ArcGISRuntimePackageVersion>
+    <ArcGISRuntimePackageVersion Condition="'$(ArcGISRuntimePackageVersion)'==''">100.7.0</ArcGISRuntimePackageVersion>
```
fragmer: Nice catch!

src/ARToolkit.Forms/ARSceneView.cs
nCastle1: Default is 0, and zero should mean no clipping

nCastle1:
Thanks for the reviews!

PR #317, opened 2020-02-25
Update default legend templates for UWP & WPF

nCastle1:
Summary of changes:

* Changes font styling slightly to more closely match default behavior from other apps, including the new Online map viewer
    * Layer names slightly smaller
    * Bolded layer names
* Wrap text instead of horizontally scrolling - more closely matches other UIs and works better out of the box (imho - this is debatable)
* Removes the (imho) distracting hover/reveal effect in the UWP listview; the legend isn't interactive, so it shouldn't have the reveal effect (per my interpretation of [the guidelines](https://docs.microsoft.com/en-us/windows/uwp/design/style/reveal#dos-and-donts))
* Removes the hover effect in WPF - legend isn't interactive so there shouldn't be hover effects
* Updates the UWP sample to show a more complex map (needed for testing wrapping behavior and scrolling)

### DIFF
--- a/src/Toolkit/Toolkit.WPF/UI/Controls/Legend/Legend.Theme.xaml
+++ b/src/Toolkit/Toolkit.WPF/UI/Controls/Legend/Legend.Theme.xaml
@@ -9,14 +9,18 @@
     <Setter Property="LayerItemTemplate">
       <Setter.Value>
         <DataTemplate>
-          <TextBlock Text="{Binding Content.Name}" FontSize="18" />
+          <TextBlock Text="{Binding Content.Name}" 
+                     FontSize="14" 
+                     TextWrapping="Wrap" 
+                     FontWeight="Semibold"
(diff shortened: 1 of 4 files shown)


src/Toolkit/Toolkit.WPF/UI/Controls/Legend/Legend.Theme.xaml
```diff
@@ -25,23 +29,28 @@
...
           </StackPanel>
         </DataTemplate>
       </Setter.Value>
     </Setter>
     <Setter Property="Template">
       <Setter.Value>
         <ControlTemplate TargetType="controls:Legend">
-            <ListView x:Name="List"
+          <ScrollViewer>
+            <ItemsControl x:Name="List"
```
dotMorten: Undo this. This would get rid of virtualization support

PR #338, opened 2020-08-30
Introduces ItemContainerStyle to BookmarksView

nCastle1:
While working on the refreshed UI for the Data Collection app, I had a need to customize the appearance of the bookmarks view item container. I was able to move forward without the changes in this PR by setting a list view item container style globally, but this approach may not be convenient for our customers.


src/Toolkit/Toolkit/UI/Controls/BookmarksView/BookmarksView.Windows.cs
```diff
@@ -141,6 +141,21 @@ public DataTemplate ItemTemplate
...
+        {
+            get { return (Style)GetValue(ItemContainerStyleProperty); }
+            set { SetValue(ItemContainerStyleProperty, value); }
+        }
+
+        /// <summary>
+        /// Identifies the <see cref="ItemContainerStyle"/> dependency property.
+        /// </summary>
+        public static readonly DependencyProperty ItemContainerStyleProperty =
+            DependencyProperty.Register(nameof(ItemContainerStyle), typeof(Style), typeof(BookmarksView), new PropertyMetadata(null));
```
dotMorten, line 158: Instead of `new PropertyMetadata(null)` you can just say `null`

dotMorten (approved):
Looks great, but does Xamarin.Forms need this new property too?

nCastle1:
I checked and it doesn't look like Xamarin.Forms has this concept.

PR #339, opened 2020-09-01
Improves the appearance of the popupviewer for iOS

nCastle1:
Previously, the popup viewer didn't work well with dynamic type. Now it adapts to any system font settings, including larger accessibility sizes.

| Smallest text | Largest text |
|-----------|-----------|
| updated_smalltext | updated_largetext |

The new appearance is a little different than the old appearance on the smallest text size, but I think that is ok because there was some clipping of descenders previously (e.g. y, g)


nCastle1:
Note there are some appearance issues with the modal and the Xamarin.Forms sample appears to be in a non-working state, but that is out of scope for this PR - both issues also affect master.

PR #350, opened 2021-03-18
Fixed initial heading getting overwritten in Initial and Continuous tracking modes on Android

kamszyc:
In my testing, when `ARSceneView` is configured with `NorthAlign` set to `true` and tracking mode set to `Initial` or `Continuous` then aligning using compass doesn't work properly on Android.
`OrientationHelper_OrientationChanged` gets called first setting `OriginCamera` with initial heading, then `LocationDataSource_LocationChanged` gets called, overwriting initial heading with zero.

In this pull request I propose a simple fix to this problem. I'm not entirely sure this is the best solution, maybe some kind of boolean flag and lock preventing initial heading from overwriting would do better.

Tested on POCO X3 NFC phone, running Android 11.

### DIFF
--- a/src/ARToolkit/ARSceneView.cs
+++ b/src/ARToolkit/ARSceneView.cs
@@ -156,8 +156,14 @@ namespace Esri.ArcGISRuntime.ARToolkit
                 var locationPoint = e.Position;
                 if (_locationTrackingMode == ARLocationTrackingMode.Initial || !_initialLocationSet)
                 {
+#if __ANDROID__
+                    double heading = _initialHeading ?? 0;
+#else
+                    double heading = 0;
+#endif
+
                     // if location has altitude, use that else use a default value
-                    var newCamera = new Mapping.Camera(locationPoint.Y, locationPoint.X, locationPoint.HasZ ? locationPoint.Z : 1, 0, 90, 0);
+                    var newCamera = new Mapping.Camera(locationPoint.Y, locationPoint.X, locationPoint.HasZ ? locationPoint.Z : 1, heading, 90, 0);
                     OriginCamera = newCamera;
                     _initialLocationSet = true;
                 }
--- a/src/ARToolkit/ARSceneView.cs
+++ b/src/ARToolkit/ARSceneView.cs
@@ -156,12 +156,7 @@ namespace Esri.ArcGISRuntime.ARToolkit
                 var locationPoint = e.Position;
                 if (_locationTrackingMode == ARLocationTrackingMode.Initial || !_initialLocationSet)
                 {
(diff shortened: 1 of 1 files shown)


src/ARToolkit/ARSceneView.cs
```diff
@@ -156,8 +156,14 @@ private void LocationDataSource_LocationChanged(object sender, Location.Location
                 var locationPoint = e.Position;
                 if (_locationTrackingMode == ARLocationTrackingMode.Initial || !_initialLocationSet)
                 {
+#if __ANDROID__
+                    double heading = _initialHeading ?? 0;
```
dotMorten, line 160: I think this should only be done if `NorthAlign` and `IsUsingARCore` is true. ie
`double heading = NorthAlign && IsUsingARCore && _initialHeading.HasValue ? _initialHeading.Value : 0;`
In addition I'm still not entirely convinced we should be using the initial value, but rather the current value of the compass heading. Or rather it should be the value at the point in time where ARCore started up as started tracking, as that is what ARCore considers "0"
An even simpler solution might be to just grab the heading from the current OriginCamera.
ie `var heading = OriginCamera.Heading;`

dotMorten (commented):
Thank you for this. The current implementation matches the native Android implementation here: https://github.com/Esri/arcgis-runtime-toolkit-android/blob/2ac0e76793ef07887ad3bf234981a2a89424f9af/arcgis-android-toolkit/src/main/java/com/esri/arcgisruntime/toolkit/ar/ArcGISArView.kt#L292
Perhaps @mrnelson6 or @puneet-pdx can comment on whether they've seen this happen in the Android SDK

src/ARToolkit/ARSceneView.cs
kamszyc: I updated pull request to use heading from OriginCamera instead.
Seems to also work well when I tested it.

mrnelson6:
The change looks good to me. I can't comment on the existence of the issue on native android unfortunately.

nCastle1 (approved):
I tested with a Samsung Galaxy Note 10 and confirmed the behavior has improved with this PR. I also tested on iPad to confirm that iOS would not be adversely affected. Everything seems to be in order and I do not have any concerns about moving forward.

Great find, thanks for contributing! :)

dotMorten (approved):
👍

kamszyc:
So, if you approve my changes, @dotMorten @nCastle1, can you merge this pull request?
I'm interested in building and using toolkit straight from repository main branch and merge would make this easier for me.

nCastle1:
@kamszyc Thank you for the PR!

PR #364, opened 2021-06-13
OverviewMap

nCastle1:
Implementation of OverviewMap for WPF, UWP, and Xamarin.Forms.


src/Toolkit/Toolkit/UI/Controls/OverviewMap/OverviewMapController.cs
```diff
@@ -0,0 +1,212 @@
...
+            else if (_connectedView is MapView)
+            {
+                _extentOverlay.Renderer = new SimpleRenderer(
+                    new SimpleFillSymbol(SimpleFillSymbolStyle.Null, System.Drawing.Color.Transparent,
+                    new SimpleLineSymbol(SimpleLineSymbolStyle.Solid, System.Drawing.Color.Red, 1)));
+            }
+            else if (_connectedView is SceneView)
+            {
+                _extentOverlay.Renderer = new SimpleRenderer(
+                    new SimpleMarkerSymbol(SimpleMarkerSymbolStyle.Cross, System.Drawing.Color.Red, 16));
```
dotMorten, line 207: Consider making the default symbols a private static variable for reuse (unless you start exposing both as properties, when this wouldn't be needed)

src/Toolkit/Toolkit/Esri.ArcGISRuntime.Toolkit.csproj
```diff
@@ -18,6 +18,14 @@
...
     <GenerateLibraryLayout>true</GenerateLibraryLayout>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == '$(AndroidTargetFramework)'">
+    <DefineConstants>$(DefineConstants);XAMARIN;_XAMARIN_ANDROID_</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == '$(iOSTargetFramework)'">
+    <DefineConstants>$(DefineConstants);XAMARIN;_XAMARIN_IOS_</DefineConstants>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
nCastle1: I need to exclude native Xamarin.iOS and native Xamarin.Android without excluding Xamarin.Forms on Android and iOS.

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,219 @@
...
+        {
+            get => (double)GetValue(ScaleFactorProperty);
+            set => SetValue(ScaleFactorProperty, value);
+        }
+
+        /// <summary>
+        /// Identifies the <see cref="AreaSymbol"/> bindable property.
+        /// </summary>
+        public static readonly BindableProperty AreaSymbolProperty =
+            BindableProperty.Create(nameof(AreaSymbol), typeof(Symbol), typeof(OverviewMap), new SimpleFillSymbol(SimpleFillSymbolStyle.Null, System.Drawing.Color.Transparent, new SimpleLineSymbol(SimpleLineSymbolStyle.Solid, System.Drawing.Color.Red, 1)), propertyChanged: OnAreaSymbolChanged);
```
dotMorten, line 153: You shouldn't have mutable reference types as default values in dependency properties, since they are static and shared among all. Either specify them in the template or the constructor, and make the default here `null`.

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,219 @@
...
+        /// Identifies the <see cref="Map"/> bindable property.
+        /// </summary>
+        public static readonly BindableProperty MapProperty =
+            BindableProperty.Create(nameof(Map), typeof(Map), typeof(OverviewMap), null, propertyChanged: OnMapPropertyChanged);
+
+        /// <summary>
+        /// Identifies the <see cref="PointSymbol"/> bindable property.
+        /// </summary>
+        public static readonly BindableProperty PointSymbolProperty =
+            BindableProperty.Create(nameof(PointSymbol), typeof(Symbol), typeof(OverviewMap), new SimpleMarkerSymbol(SimpleMarkerSymbolStyle.Cross, System.Drawing.Color.Red, 16), propertyChanged: OnPointSymbolChanged);
```
dotMorten, line 171: Same. Don't define default reference type values

src/Toolkit/Toolkit/Esri.ArcGISRuntime.Toolkit.csproj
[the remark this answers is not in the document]
dotMorten: I'm not following? This project isn't using Forms. These are standard defines. If you need to do something special for forms, define `XAMARIN_FORMS` in the forms project, and use that. `XAMARIN_IOS_` is equivalent to `__IOS__` so it isn't clear how these are different. By having a forms specific conditional, the code would much more clearly read "this is for forms only"

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,219 @@
...
+            ControlTemplate = DefaultControlTemplate;
+        }
+
+        /// <summary>
+        /// <inheritdoc />
+        /// </summary>
+        protected override void OnApplyTemplate()
+        {
+            base.OnApplyTemplate();
+
```
dotMorten, line 68: If suggest "cleaning up" the old controller here if it isn't null (set AttachedView to null to unhook event handlers, as well as clear out the weak references - perhaps just add a dispose that does this?)

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,219 @@
...
+            if (_overviewMapView != null)
+            {
+                _controller = new OverviewMapController(_overviewMapView)
+                {
+                    ScaleFactor = ScaleFactor,
+                    PointSymbol = PointSymbol,
+                    AreaSymbol = AreaSymbol,
+                    AttachedView = GeoView,
+                };
+                _overviewMapView.IsAttributionTextVisible = false;
```
dotMorten, line 80: Do this in the template

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,219 @@
...
+
+                if (Map == null)
+                {
+                    Map = new Map(BasemapStyle.ArcGISTopographic);
+                }
```
dotMorten, line 85: This should be done in the constructor instead. Or set directly on the `_overviewMapView.Map` property so the public property stays `null`. Also you should probably set _overviewMapView.Map regardless, or it won't get applied if you set the property before applying the template

src/Toolkit/Toolkit/UI/Controls/OverviewMap/OverviewMapController.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        public double ScaleFactor
+        {
+            get => _scaleFactor;
+            set
+            {
+                if (_scaleFactor != value)
+                {
+                    _scaleFactor = value;
```
dotMorten, line 115: Should there be a range check here? For instance not 1 and less?

src/Toolkit/Toolkit.WPF/Themes/Generic.xaml
```diff
@@ -1,6 +1,4 @@
-﻿<ResourceDictionary
-    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
```
dotMorten, line 1: This seems like an unnecessary change?

src/Toolkit/Toolkit/Internal/LoadStatusToVisibilityConverter.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    return Visibility.Visible;
+                }
+                else if (parameterString == "FailedToLoad" && status == LoadStatus.FailedToLoad)
+                {
+                    return Visibility.Visible;
+                }
+            }
+
+            return Visibility.Collapsed;
```
dotMorten, line 60: Should this return visible during design time?

src/Toolkit/Toolkit/UI/Controls/OverviewMap/OverviewMap.Windows.cs
```diff
@@ -0,0 +1,209 @@
...
+#endif
+
+namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
+{
+    /// <summary>
+    /// Defines a small "overview" (or "inset") map displaying a representation of the attached <see cref="GeoView"/>'s current viewpoint.
+    /// </summary>
+    [TemplatePart(Name = "PART_MapView", Type = typeof(MapView))]
+    public class OverviewMap : Control
+    {
```
dotMorten, line 37: Pretty much all comments for the Forms version applies to this as well

src/Samples/Toolkit.SampleApp.WPF_Net5/Samples/OverviewMap/OverviewMapSample.xaml
```diff
@@ -0,0 +1,56 @@
...
+    <Grid>
+        <esri:MapView x:Name="MyMapView" />
+        <esri:SceneView x:Name="MySceneView" Visibility="Collapsed" />
+        <toolkit:OverviewMap
+            x:Name="SampleOverview"
+            Width="100"
+            Height="100"
+            Margin="8"
+            HorizontalAlignment="Right"
+            VerticalAlignment="Top" />
```
dotMorten, line 21: I see you define 100x100 and upper-right a lot, and these are defaults. Do we really need them here? Since this is mainly a functional test app this also helps validating those defaults work.

dotMorten (changes requested):

docs/overview-map.md
```diff
@@ -0,0 +1,89 @@
...
+             xmlns:esriUI="clr-namespace:Esri.ArcGISRuntime.Xamarin.Forms;assembly=Esri.ArcGISRuntime.Xamarin.Forms">
+    <ContentPage.Content>
+        <Grid>
+            <esriUI:SceneView x:Name="sceneView" />
+            <StackLayout Margin="4"
+                         HorizontalOptions="Start"
+                         VerticalOptions="Start">
+                <esriTK:OverviewMap GeoView="{Binding Source={x:Reference sceneView}}"
+                                    HeightRequest="100"
+                                    WidthRequest="100" />
```
dotMorten, line 83: Aren't these default, so they can be removed? (also good to have the functional test not set these so we know the default works). The non-form samples doesn't appear to be setting these.

src/Samples/Toolkit.Samples.Forms/Toolkit.Samples.Forms/Samples/OverviewMapSample.xaml
```diff
@@ -0,0 +1,34 @@
...
+                <Button Clicked="ToggleModelClick" Text="Apply new Map/Scene" />
+                <Button Clicked="ToggleSymbolsClick" Text="Toggle symbols" />
+                <Label Text="Map rotation (map only):" />
+                <Slider Maximum="360" ValueChanged="Slider_ValueChanged" />
+                <Label Text="Scale factor:" />
+                <Slider Maximum="50" Minimum="10" Value="{Binding Source={x:Reference SampleOverview}, Path=ScaleFactor, Mode=TwoWay}" />
+                <esriTK:OverviewMap
+                    x:Name="SampleOverview"
+                    HeightRequest="100"
+                    WidthRequest="100" />
```
dotMorten, line 30: Remove?

src/Toolkit.Forms/OverviewMap/LoadStatusToVisibilityConverter.cs
```diff
@@ -0,0 +1,77 @@
...
+using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms.Internal
+{
+    /// <summary>
+    /// *FOR INTERNAL USE* Returns a boolean representing visibility for a given load status.
+    /// </summary>
+    [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
```
dotMorten, line 26: This is an internal class, so you can remove that

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,226 @@
...
+
+        private MapView? _overviewMapView;
+
+        static OverviewMap()
+        {
+            string template = @"<ControlTemplate xmlns=""http://xamarin.com/schemas/2014/forms""
+                                                 xmlns:x=""http://schemas.microsoft.com/winfx/2009/xaml""
+                                                 xmlns:esri=""clr-namespace:Esri.ArcGISRuntime.Xamarin.Forms;assembly=Esri.ArcGISRuntime.Xamarin.Forms""
+                                                 xmlns:internal=""clr-namespace:Esri.ArcGISRuntime.Toolkit.Xamarin.Forms.Internal;assembly=Esri.ArcGISRuntime.Toolkit.Xamarin.Forms"">
+                                    <Frame Padding=""1"" WidthRequest=""100"" HeightRequest=""100"" CornerRadius=""0"" HasShadow=""False"" BorderColor=""Black"" BackgroundColor=""White"">
```
dotMorten, line 42: Should this size be hardcoded here? Wouldn't that mean a larger size around it won't make it stretch?

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OverviewMap"/> class.
+        /// </summary>
+        public OverviewMap()
+        {
+            ControlTemplate = DefaultControlTemplate;
+            AreaSymbol = new SimpleFillSymbol(SimpleFillSymbolStyle.Null, System.Drawing.Color.Transparent, new SimpleLineSymbol(SimpleLineSymbolStyle.Solid, System.Drawing.Color.Red, 1));
+            PointSymbol = new SimpleMarkerSymbol(SimpleMarkerSymbolStyle.Cross, System.Drawing.Color.Red, 16);
+            Map = new Map(BasemapStyle.ArcGISTopographic);
+        }
```
dotMorten, line 65: Define default size and alignment here like WPF/UWP?

src/Toolkit.Forms/OverviewMap/OverviewMap.cs
```diff
@@ -0,0 +1,226 @@
...
+        {
+            base.OnApplyTemplate();
+
+            _overviewMapView = GetTemplateChild("PART_MapView") as MapView;
+
+            if (_overviewMapView != null)
+            {
+                _overviewMapView.Map = Map;
+
+                _controller?.Dispose();
```
dotMorten, line 80: Do this _before `_overviewMapView` is assigned in line 74.
Ie before line 74, you clean up from the old round (also important to unhook the old map, since it can only be used on one mapview):
```cs
if (_overviewMapView != null)
{
    _overviewMapView.Map = null;
    _controller?.Dispose();
}
```

src/Toolkit/Toolkit.UWP/UI/Controls/OverviewMap/OverviewMap.Theme.xaml
```diff
@@ -0,0 +1,46 @@
...
+                    </Border>
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>
+        <Setter Property="HorizontalAlignment" Value="Right" />
+        <Setter Property="VerticalAlignment" Value="Top" />
+        <Setter Property="BorderBrush" Value="Black" />
+        <Setter Property="BorderThickness" Value="1" />
+        <Setter Property="Width" Value="100" />
+        <Setter Property="Height" Value="100" />
```
dotMorten, line 44: I seems to me Forms are missing the equivalent of this?

src/Toolkit/Toolkit/Internal/LoadStatusToVisibilityConverter.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Windows;
+using System.Windows.Data;
+#endif
+
+namespace Esri.ArcGISRuntime.Toolkit.Internal
+{
+    /// <summary>
+    /// *FOR INTERNAL USE* Returns visibility for the given load status.
+    /// </summary>
+    [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
```
dotMorten, line 34: This is an internal class, so not needed

src/Toolkit/Toolkit/UI/Controls/OverviewMap/OverviewMap.Windows.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        /// <inheritdoc/>
+#if NETFX_CORE
+        protected override void OnApplyTemplate()
+#else
+        public override void OnApplyTemplate()
+#endif
+        {
+            base.OnApplyTemplate();
```
dotMorten, line 60: Same comment around clean up as in Forms.

dotMorten (changes requested):
Looking great! Main (minor) concerns left are default sizing/placement of Forms + cleanup during applying templates.

dotMorten (approved):

PR #370, opened 2021-07-31
BasemapGallery implementation

nCastle1:


dotMorten (changes requested):
Just some initial comments - mostly around overall design.
Also, when using `await` and you don't have to come back to the UI thread in a given method, add `.ConfigurateAwait(false)` so we reduce number of context switches needed.

PR #371, opened 2021-08-01
Update NuGet tags

mbranscomb:


mbranscomb:
@dotMorten @nCastle1 Added the new set of package tags from the UI/API packages.

dotMorten (approved):

PR #372, opened 2021-08-10
UWP: force include custom renderer assemblies

gdestigter:
Fix for Xamarin.Forms UWP linker removing custom renderer assemblies from the samples app.


mbranscomb (approved):
@gdestigter LGTM

dotMorten (approved):

PR #384, opened 2021-11-27
Fix BasemapGallery for Forms UWP release

nCastle1:
Certification revealed that BasemapGallery as previously implemented did not work with the .NET Native compiler in release. .NET Native didn't like the internal interface pattern, and it was faster to change it than to figure out how to make .NET Native accept it.

Summary of changes:

- Remove the IBasemapGallery interface. Instead of the controller calling into the view via that interface, the controller uses `INotifyPropertyChanged` and the view listens for changes.
- Fixes to some warnings that were missed previously, mostly in the area of nullability (BasemapGallery code only)
- Disable un-addressable warnings around nullable arrays (stylecop rules conflict)
- Documentation link error fix


src/Toolkit.Forms/BasemapGallery/BasemapGallery.Appearance.cs
```diff
@@ -42,22 +42,22 @@ static BasemapGallery()
 
             DefaultGridDataTemplate = new DataTemplate(() =>
             {
-                Grid outerScrimContainer = new Grid();
```
nCastle1, line 45: All of these changes address the 'new expression can be simplified' build message.

dotMorten: We should disable that message instead, as it isn't a requirement for our coding guidelines. I would remove these changes from this PR and we can disable those warnings in a separate PR

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
```diff
@@ -17,35 +17,124 @@
...
 using System.Collections.ObjectModel;
 using System.Collections.Specialized;
+using System.ComponentModel;
 using System.Linq;
 using System.Threading.Tasks;
 using Esri.ArcGISRuntime.Mapping;
 using Esri.ArcGISRuntime.Portal;
-using Esri.ArcGISRuntime.Toolkit.UI.Controls;
+
+[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("Esri.ArcGISRuntime.Toolkit.Xamarin.Forms")]
```
dotMorten, line 26: Shouldn't this be removed?

dotMorten (approved):
See comments. Also check whether we can remove all `internalsvisibleto`

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
nCastle1: I did try removing that, but that would require making more API public on `BasemapGalleryItem` than we're comfortable with.

dotMorten (approved):

PR #388, opened 2021-12-11
BasemapGallery WPF: fix loading scrim

nCastle1:
Previously, the loading scrim, which obscures the listview while the collection is being updated, had a padding that exposed the control beneath it. This PR removes that padding, and adds a margin to the progress bar to preserve the original intended appearance.

image


mstefarov (approved):
Makes sense 👍

PR #389, opened 2021-12-11
UWP: fix missing glyph in FeatureDataField

nCastle1:
This change fixes a missing icon in the FeatureDataField textbox.

After:

image

Before:

image

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,12 +1,9 @@
 <Project>
 
   <PropertyGroup>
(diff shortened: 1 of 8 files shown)


mstefarov (approved):
Looks like the old codepoint came from Segoe UI Symbol font, which was default symbol font for Windows 8.x store apps.

For Windows 10+, SymbolThemeFontFamily maps to the **Segoe MDL2** font instead, which uses the new codepoint.

Because we [no longer support](https://developers.arcgis.com/net/reference/system-requirements/) UWP on Windows 8.x, this is an appropriate change to make 👍

Issue #373, opened 2021-08-18
iOS: Compass does not redraw after resizing


mstefarov:
As seen in the native iOS sample for Compass:
Resizing the compass does not trigger a redraw. That means going from small to large size causes the compass to appear very blurry -- until the view is rotated.

![abeae600-feeb-11eb-93fd-9f454ca09bbd](https://user-images.githubusercontent.com/587809/129855128-c35c054d-de43-41cf-ad26-5235eb22c197.gif)

So we should probably be listening to viewDidLayoutSubviews event (called when compass is resized by its parent) and call SetNeedsDisplay() to redraw.

nCastle1:
Issue resolved by #396

PR #396, opened 2021-12-20
iOS - Redraw compass on resize

nCastle1:
As reported in #373, the compass was not re-drawing after resize on iOS, resulting in a blurry appearance. Because Forms uses the underlying platform view, Forms on iOS was also affected.

This change ensures that the view is redrawn when the size changes.


mstefarov (approved):
Looks good 👍

dotMorten (approved):

PR #402, opened 2021-12-28
SymbolDisplay - Remove async void

nCastle1:
This change makes the previously async void Refresh method return Task. The Windows implementation and other parts of the shared implementation are updated to discard the task.

Testing with Xamarin.Forms confirms that the previously-discovered crash no longer repros.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/SymbolDisplay/SymbolDisplay.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/SymbolDisplay/SymbolDisplay.cs
@@ -68,17 +68,17 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                 {
                     OnEventAction = (instance, source, eventArgs) =>
                     {
-                        _ = Refresh();
+                        Refresh();
                     },
                     OnDetachAction = (instance, weakEventListener) => instance.PropertyChanged -= weakEventListener.OnEvent,
                 };
                 newValue.PropertyChanged += _inpcListener.OnEvent;
             }
 
-            _ = Refresh();
+            Refresh();
         }
 
-        private async Task Refresh()
+        private async void Refresh()
         {
             try
             {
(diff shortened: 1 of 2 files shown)


src/Toolkit/Toolkit/UI/Controls/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -67,33 +68,40 @@ private void OnSymbolChanged(Symbology.Symbol? oldValue, Symbology.Symbol? newVa
...
-            _currentUpdateTask = null;
-            if (_isRefreshRequired)
+                _isRefreshRequired = false;
+                var task = _currentUpdateTask = UpdateSwatchAsync();
+                await task;
+                _currentUpdateTask = null;
+                if (_isRefreshRequired)
+                {
+                    await Refresh();
+                }
```
mstefarov, line 100: Alternative solution -- functionally same but avoids recursion:
```suggestion
                do
                {
                    _isRefreshRequired = false;
                    var task = _currentUpdateTask = UpdateSwatchAsync();
                    await task;
                } while (_isRefreshRequired);
                _currentUpdateTask = null;
```

mstefarov (approved):

src/Toolkit/Toolkit/UI/Controls/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -67,33 +68,39 @@ private void OnSymbolChanged(Symbology.Symbol? oldValue, Symbology.Symbol? newVa
                 {
                     OnEventAction = (instance, source, eventArgs) =>
                     {
-                        Refresh();
+                        _ = Refresh();
```
dotMorten, line 71: Since you're not awaiting the refresh anywhere, there's not much point to this change. You're still creating a potentially unobserved task exception.
async-void is ok, as long as there's a try-catch to catch any exceptions. Adding a try-catch should be sufficient.

dotMorten (approved):

PR #405, opened 2021-12-31
TimeSlider improvements

nCastle1:
This PR takes a first step towards improving the TimeSlider and the developer experience.

Changes include:

- Improved WPF sample to better exercise the capabilities of the control
- Added samples for UWP, Xamarin.Forms, iOS, and Android
- Updated the markdown doc to have a page on time slider, which highlights useful API
- Updated other markdown documentation to include a picture
- Fixed some API reference content for the Xamarin.Forms implementation
- Fixed the issue where a white rectangle was shown behind the play/pause button on WPF
- Fixed a layout issue where the thumb and right-most part of the slider track were being clipped on WPF
- Adds a check for time step counts less than one, which would previously cause strange behavior.
- ⚠️ Partially fixed an issue where time handling code was causing some time intervals to be left off of the slider. These changes are in TimeSlider.cs and TimeExtensions.cs and are the riskiest changes in the PR. To see the benefit, you can manually set the time steps; before the change, with the time-aware layer in the sample, step counts of 7, 11, 13, 17, 19, 21, and 29  would result in steps being left off the slider. With this change, all steps are present.
- Fixes an issue where initializing the time slider from a GeoView wouldn't work if the map or scene wasn't already loaded.
- Fixes an issue where initializing the time slider from a GeoView would not update the slider's current extent if the GeoView's current extent was already null.

Follow up issues will be logged to address design issues, further sample improvements, and testing needs.

### DIFF
--- a/src/Toolkit.Forms/TimeSlider/TimeSlider.cs
+++ b/src/Toolkit.Forms/TimeSlider/TimeSlider.cs
@@ -29,7 +29,8 @@ using Windows.UI.Xaml.Media;
 namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
 {
     /// <summary>
-    /// The Compass Control showing the heading on the map when the rotation is not North up / 0.
+    /// The TimeSlider is a utility Control that emits TimeExtent values typically for use with the Map Control
+    /// to enhance the viewing of geographic features that have attributes based upon Date/Time information.
     /// </summary>
     public class TimeSlider : View
     {
@@ -220,7 +221,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
             BindableProperty.Create(nameof(FullExtent), typeof(TimeExtent), typeof(TimeSlider), null, BindingMode.OneWay, null, OnFullExtentPropertyChanged);
 
         /// <summary>
-        /// Gets or sets the <see cref="TimeExtent" /> associated with the visual thumbs(s) displayed on the TimeSlider.
+        /// Gets or sets the <see cref="TimeExtent" /> that specifies the overall start and end time of the time slider instance.
         /// </summary>
         public TimeExtent? FullExtent
         {
@@ -251,7 +252,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
             BindableProperty.Create(nameof(TimeStepInterval), typeof(TimeValue), typeof(TimeSlider), null, BindingMode.OneWay, null, OnTimeStepIntervalPropertyChanged);
 
         /// <summary>
-        /// Gets or sets the <see cref="TimeValue" /> associated with the visual thumbs(s) displayed on the TimeSlider.
+        /// Gets or sets the time step intervals for the time slider.  The slider thumbs will snap to and tick marks will be shown at this interval.
         /// </summary>
         public TimeValue? TimeStepInterval
         {
--- a/src/Toolkit.Forms/TimeSlider/TimeSlider.cs
+++ b/src/Toolkit.Forms/TimeSlider/TimeSlider.cs
@@ -965,7 +965,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
         /// <summary>
         /// Updates the time slider to have the specified number of time steps.
         /// </summary>
-        /// <param name="count">The number of time steps.</param>
+        /// <param name="count">The number of time steps. Values less than one are ignored.</param>
         /// <remarks>This method divides the TimeSlider instance's <see cref="FullExtent"/> into the number of steps specified,
         /// updating the <see cref="TimeStepInterval"/> and <see cref="TimeSteps"/> properties.  The method will attempt to set
         /// the interval to a TimeValue with the smallest duration and largest time unit that will fit evenly (i.e. without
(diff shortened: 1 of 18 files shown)


src/Toolkit.Forms/TimeSlider/TimeSlider.cs
```diff
@@ -29,7 +29,8 @@
 namespace Esri.ArcGISRuntime.Toolkit.Xamarin.Forms
 {
     /// <summary>
-    /// The Compass Control showing the heading on the map when the rotation is not North up / 0.
+    /// The TimeSlider is a utility Control that emits TimeExtent values typically for use with the Map Control
+    /// to enhance the viewing of geographic features that have attributes based upon Date/Time information.
```
dotMorten, line 33: 😱

src/Toolkit/Toolkit.WPF/UI/Controls/TimeSlider/TimeSlider.Theme.xaml
```diff
@@ -289,6 +276,13 @@
                             <ColumnDefinition Width="Auto" />
                         </Grid.ColumnDefinitions>
 
+                        <!-- Corrects layout issues that would otherwise happen. -->
```
dotMorten, line 279: Remove. The comment isn't helpful / clear what will happen anyway.

src/Toolkit/Toolkit.WPF/UI/Controls/TimeSlider/TimeSlider.Theme.xaml
```diff
@@ -161,19 +161,6 @@
...
-                                                            <Rectangle.Fill>
-                                                                <LinearGradientBrush EndPoint=".7,1" StartPoint=".7,0">
-                                                                    <GradientStop Color="#FFFFFFFF" Offset="0"/>
-                                                                    <GradientStop Color="#F9FFFFFF" Offset="0.375"/>
-                                                                    <GradientStop Color="#E5FFFFFF" Offset="0.625"/>
-                                                                    <GradientStop Color="#C6FFFFFF" Offset="1"/>
-                                                                </LinearGradientBrush>
-                                                            </Rectangle.Fill>
-                                                        </Rectangle>
-                                                    </Grid>
```
dotMorten, line 176: Could you explain why this part was removed?

src/Toolkit/Toolkit/Internal/TimeExtensions.cs
```diff
@@ -213,21 +213,21 @@ public static DateTimeOffset AddTimeValue(this DateTimeOffset startTime, TimeVal
...
+                else if (millisecondsPerTimeStep / millisecondsPerHour > 1)
                 {
-                    return new TimeValue(millisecondsPerTimeStep / millisecondsPerMinute, TimeUnit.Minutes);
+                    return new TimeValue((int)(millisecondsPerTimeStep / millisecondsPerHour), TimeUnit.Hours);
                 }
-                else if (millisecondsPerTimeStep / millisecondsPerSecond > 1)
+                else if (millisecondsPerTimeStep / millisecondsPerDay > 1)
                 {
-                    return new TimeValue(millisecondsPerTimeStep / millisecondsPerSecond, TimeUnit.Seconds);
+                    return new TimeValue((int)(millisecondsPerTimeStep / millisecondsPerDay), TimeUnit.Days);
```
dotMorten, line 230: I don't fully understand this reversal. It seems to me the first expression would always be true now, as it is more likely to be greater than one when you divide by a large number.

dotMorten (commented):
See comments. It's really hard to review though, as this PR addresses multiple different issues, so hard to understand which change fixes what.

src/Toolkit/Toolkit.WPF/UI/Controls/TimeSlider/TimeSlider.Theme.xaml
nCastle1: This specific part was causing an unexpected white background to be drawn behind the play button:

image

It is the first checkbox in https://github.com/Esri/arcgis-toolkit-dotnet/issues/193

src/Toolkit/Toolkit/Internal/TimeExtensions.cs
nCastle1: Yeah that's probably what will happen. The initial behavior of choosing the larger intervals wasn't leading to good results.

src/Toolkit/Toolkit/UI/Controls/TimeSlider/TimeSlider.cs
```diff
@@ -857,6 +857,12 @@ private void CalculateTimeSteps()
 
             for (var nextStep = startTime.AddTimeValue(timeStep); nextStep <= endTime; nextStep = nextStep.AddTimeValue(timeStep))
             {
+                // TODO = come up with a more rigorous definition or expose API
+                if (endTime.Subtract(nextStep).TotalMinutes < 10)
+                {
+                    nextStep = endTime;
+                }
```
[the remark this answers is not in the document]
nCastle1: The intent is to snap the end of the last interval to the end of the time extent in cases where imprecision in time manipulation leads to the combined length of time steps being slightly loo long. 

e.g. 10years/6 intervals = 1.67 year intervals; 1.67 years*6 intervals = 10.2 years; previously in cases like that, the last time interval would just be skipped.

Other changes in TimeExtensions are intended to make the time manipulation as precise as possible (milliseconds). So ideally any error is limited to milliseconds rather than years. This seems to make things work for some common cases, but it really doesn't solve the bigger problems with the control.

dotMorten (approved):
👍

PR #412, opened 2022-02-09
AR Project Templates

dotMorten:
Moves the project templates from [the template repo](https://github.com/Esri/arcgis-runtime-templates-dotnet) to here.
Only difference from old repo: Version parameters are parameterized, and digital signing support added.


nCastle1 (commented):
I'm not sure what's going on, but when I build the template project and run the installer, only the iOS template is installed into VS 2022.

nCastle1 (approved):
I was able to build the template vsix locally.

PR #432, opened 2022-08-01
Samples improvements - reliability

nCastle1:
Improvements to the simplicity and reliability of samples as functional tests:

- Replace deprecated basemap constructors, replace with newer services
- Replace use of sampleserver6 with ArcGIS Online-hosted services where possible
- Simplify popupviewer sample by removing callout with image button
- BasemapGallery samples now add real basemaps when manually modifying the list
- UWP updated community toolkit and removed deprecated functionality
- Removed Data grid sample, because the grid is not part of toolkit. The sample is misleading and prone to bugs caused by the sample implementation rather than the toolkit component.


src/Samples/Toolkit.SampleApp.Android/Samples/PopupSample.cs
```diff
@@ -8,45 +8,26 @@
...
-            // Used to demonstrate display of EditSummary in PopupViewer
-            // Provides credentials to token-secured layer that has editor-tracking enabled
-            AuthenticationManager.Current.ChallengeHandler = new ChallengeHandler(async (info) =>
-            {
-                return await AuthenticationManager.Current.GenerateCredentialAsync(info.ServiceUri, "user1", "user1");
-            });
-
             mapView = FindViewById<MapView>(Resource.Id.mapView);
-            mapView.Map = new Map(Basemap.CreateLightGrayCanvasVector());
+            mapView.Map = new Map(BasemapStyle.ArcGISLightGray);
```
dotMorten, line 26: This would prevent functional tests to work out of the box.

PR #436, opened 2022-08-02
Correct Popup placement on WPF

### LINKED ISSUE #428: Fix WPF popups on left-handed PCs
See: https://stackoverflow.com/questions/20363100/wpf-popup-location-issue

image

nCastle1:
Resolves #428

By default, WPF will change the placement of popups on touch screen devices configured for left handed use. This change resets that behavior if necessary and ensures that it stays reset if screen configuration changes.

See https://stackoverflow.com/questions/20363100/wpf-popup-location-issue for details

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/SearchView/SearchView.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/SearchView/SearchView.cs
@@ -29,6 +29,7 @@ using Esri.ArcGISRuntime.UI.Controls;
 #if !NETFX_CORE
 using System.Windows;
 using System.Windows.Controls;
+using System.Reflection;
 #else
 using System.Collections;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


src/Toolkit/Toolkit/UI/Controls/SearchView/SearchView.cs
```diff
@@ -79,6 +80,32 @@ public SearchView()
             RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
         }
 
+        // Ensure popup appears in correct location on touch-enabled PCs configured for left-handed use.
+#if !NETFX_CORE
```
dotMorten, line 84: Should this read `#if WPF` instead?

src/Toolkit/Toolkit/UI/Controls/SearchView/SearchView.cs
```diff
@@ -79,6 +80,32 @@ public SearchView()
...
+        private static readonly FieldInfo _popupAlignmentField;
+
+        static SearchView()
+        {
+            _popupAlignmentField = typeof(SystemParameters).GetField("_menuDropAlignment", BindingFlags.NonPublic | BindingFlags.Static);
```
dotMorten, line 89: We shouldn't be relying on reflection into the internals of .NET

dotMorten (changes requested):

PR #440, opened 2022-08-05
Place validation error in popup on UWP

nCastle1:
Resolves 390 and improves layout when an error message is being shown.


dotMorten:
Would setting the tooltip property  + adding red outline be better? Perhaps force-opening the tooltip on field focus.

PR #443, opened 2022-08-13
Revise weak event listener

### LINKED ISSUE #365: Update WeakEventListener
There is a problem with current WeakEventListener pattern: Every time its OnEventAction/OnDetachAction lambdas called an instance method or referenced something from the outer context (a local, a field, a parameter), a strong reference to "this" was captured by the lambda closure.

To fix it:
- Store a reference to "source" (object to which we subscribed) on the WeakEventListener and pass it to OnDetachAction lambdas (to unsubscribe).
- When using WeakEventListener, call methods on the "instance" parameter instead of calling on "this".

See dotnet-api issue 8679 for original report.

nCastle1:
Updates the weak event listener internal API to be clearer to consumers. Also updates usage of weak event listener. 

Resolves #365

### DIFF
--- a/src/Toolkit/Toolkit/UI/LayerContentDataSource.cs
+++ b/src/Toolkit/Toolkit/UI/LayerContentDataSource.cs
@@ -221,7 +221,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
             {
                 _documentListener = new Internal.WeakEventListener<LayerContentDataSource<T>, INotifyPropertyChanged, object?, PropertyChangedEventArgs>(this, document)
                 {
-                    OnEventAction = static (instance, source, eventArgs) => instance?.DocumentPropertyChanged(instance, eventArgs.PropertyName),
+                    OnEventAction = static (instance, source, eventArgs) => instance.DocumentPropertyChanged(instance, eventArgs.PropertyName),
                     OnDetachAction = static (instance, source, weakEventListener) => source.PropertyChanged -= weakEventListener.OnEvent,
                 };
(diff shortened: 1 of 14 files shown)


src/Toolkit/Toolkit/UI/LayerContentDataSource.cs
```diff
@@ -362,10 +362,10 @@ private IEnumerable<Action> TrackLayerContentsRecursive(IEnumerable<ILayerConten
                 {
                     if (layer is INotifyPropertyChanged inpc)
                     {
-                        var listener = new Internal.WeakEventListener<INotifyPropertyChanged, object?, PropertyChangedEventArgs>(inpc)
+                        var listener = new Internal.WeakEventListener<LayerContentDataSource<T>, INotifyPropertyChanged, object?, PropertyChangedEventArgs>(this, inpc)
                         {
-                            OnEventAction = (instance, source, eventArgs) => Layer_PropertyChanged((ILayerContent)instance, eventArgs.PropertyName),
-                            OnDetachAction = (instance, weakEventListener) => instance.PropertyChanged -= weakEventListener.OnEvent,
+                            OnEventAction = static (instance, source, eventArgs) => instance?.Layer_PropertyChanged(source as ILayerContent, eventArgs.PropertyName),
+                            OnDetachAction = static (instance, source, weakEventListener) => (source as INotifyPropertyChanged).PropertyChanged -= weakEventListener.OnEvent,
```
mstefarov, line 368: Is this cast necessary? I expect `source` to already be of type INotifyPropertyChanged here.

PR #444, opened 2022-08-16
Update versions 100.15

nCastle1:


nCastle1:
Build appears to have failed due to https://stackoverflow.com/questions/73374893/nuget-restore-issue-xamarin-ios-csharp-targets-was-not-found

PR #445, opened 2022-08-17
100.15 with workflow change

nCastle1:
Build for original PR failed due to https://stackoverflow.com/questions/73374893/nuget-restore-issue-xamarin-ios-csharp-targets-was-not-found

This PR updates the github actions build workflow to install the Xamarin VS component, because that is no longer installed on the Windows-latest build runner by default.

### DIFF
--- a/.github/workflows/CIBuild.yml
+++ b/.github/workflows/CIBuild.yml
@@ -24,7 +24,10 @@ jobs:
     
     - name: Setup Visual Studio Command Prompt
       uses: microsoft/setup-msbuild@v1.0.2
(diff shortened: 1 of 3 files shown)


pMaske (approved):
@nCastle1 Thanks. Let's go with the fix until Microsoft fixes Github VMs to accommodate this change.

nCastle1:
@pMaske Microsoft has apparently fixed the problem in the last ~5 hours. I retested the original branch and now it builds without issue. I'm closing this PR in favor of the original PR.

PR #450, opened 2022-09-12
MAUI v200 fixes

nCastle1:
Fixes several issues that weren't addressed in the initial port from Forms to MAUI. See individual commits for individual issues fixed


src/Samples/Toolkit.SampleApp.Maui/Resources/Styles/Colors.xaml
```diff
@@ -17,8 +17,8 @@
     <Color x:Key="Gray600">#404040</Color>
     <Color x:Key="Gray900">#212121</Color>
     <Color x:Key="Gray950">#141414</Color>
-    <SolidColorBrush x:Key="PrimaryBrush" Color="{StaticResource Primary}"/>
-    <SolidColorBrush x:Key="SecondaryBrush" Color="{StaticResource Secondary}"/>
+    <SolidColorBrush x:Key="PrimaryBrush" Color="{AppThemeBinding Dark={StaticResource Tertiary}, Light={StaticResource Primary}}"/>
+    <SolidColorBrush x:Key="SecondaryBrush" Color="{AppThemeBinding Dark={StaticResource Tertiary}, Light={StaticResource Primary}}"/>
```
dotMorten, line 21: Isn't it better to declare the Primary color by apptheme instead so anything using 'Primary" color will use it?

src/Toolkit/Toolkit/UI/Controls/BookmarksView/BookmarksViewDataSource.cs
```diff
@@ -338,7 +338,11 @@ private void OnCollectionChanged(NotifyCollectionChangedEventArgs args)
         {
             RunOnUIThread(() =>
             {
+#if MAUI && __IOS__
+                CollectionChanged?.Invoke(this, new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
```
dotMorten, line 342: I'm not quite following what this is doing? Why a full reset?

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -137,7 +137,7 @@ private async Task UpdateSwatchAsync()
             }
         }
 
-        private double GetScaleFactor() => Window?.DisplayDensity ?? 1d;
```
dotMorten, line 140: Is this right? MainDisplayInfo is the main screen. What you want is the density of the window, which could be on any display

src/Toolkit/Toolkit.Maui/BasemapGallery/ByteArrayToImageSourceConverter.cs
```diff
@@ -30,7 +30,7 @@ internal class ByteArrayToImageSourceConverter : IValueConverter
                 return ImageSource.FromStream(() => new MemoryStream(rawBuffer));
             }
 
-            return null;
```
dotMorten, line 33: This seems like an odd fallback for a generic bytes to image source converter.

src/Toolkit/Toolkit/UI/Controls/BookmarksView/BookmarksViewDataSource.cs
nCastle1: There's a bug in the collectionview implementation on iOS, if the event is raised with an insert at the end of the list, e.g. at index 4, the collectionview implementation tries to access the collection at index 5, resulting in an exception. Doing a full reset each time the list changes fixes the problem.

src/Toolkit/Toolkit.Maui/BasemapGallery/ByteArrayToImageSourceConverter.cs
nCastle1: It is internal and only used in BasemapGallery. There are MAUI issues when converters are used in collection views. Previous to this change I was seeing the wrong thumbnail shown for items after scrolling. With this change, the issue appears to be fixed.

src/Samples/Toolkit.SampleApp.Maui/Resources/Styles/Colors.xaml
nCastle1: Maybe, but that would require that the value for the color is a bindable property, and I don't think it is.

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
nCastle1: Yes, but in testing, Window is always null. I will update it to check Window first then fall back to main display.

src/Toolkit/Toolkit.Maui/BasemapGallery/ByteArrayToImageSourceConverter.cs
dotMorten: Could you add a comment there for future reference then?

src/Toolkit/Toolkit/UI/Controls/BookmarksView/BookmarksViewDataSource.cs
dotMorten: That happens if your notification change events doesn't match exactly what is actually in the collection. iOS is more picky about this, and it becomes trickier to handle with RunOnUIThread because to collection might change more.

dotMorten: Can we at least add a //todo or an issue to address this correctly later?

dotMorten (approved):

PR #459, opened 2022-10-03
Fix assets for WinUI, UWP

nCastle1:
This PR corrects problems with the visual appearance of the MeasureToolbar and FloorFilter components on WinUI and UWP.

- Adds font to WinUI asset folder
- Adds asset folder to UWP via linking
- Adds platform-specific XAML includes files for UWP and WinUI, referenced by shared XAML file.


src/Toolkit/Toolkit.UWP/UI/Controls/FloorFilter/FloorFilter.Includes.xaml
```diff
@@ -0,0 +1,8 @@
+﻿<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:controls="using:Esri.ArcGISRuntime.Toolkit.UI.Controls"
+    xmlns:internal="using:Esri.ArcGISRuntime.Toolkit.Internal"
+    xmlns:toolkit="using:Esri.ArcGISRuntime.Toolkit.UI.Controls">
+    <FontFamily x:Key="CalciteIconFont">ms-appx:///Esri.ArcGISRuntime.Toolkit.UWP/Assets/calcite-ui-icons-24.ttf#calcite-ui-icons-24</FontFamily>
```
dotMorten, line 7: Include this once in generic.xaml instead, and let all styles be able to reuse it.

src/Toolkit/Toolkit.UWP/UI/Controls/MeasureToolbar/MeasureToolbar.Includes.xaml
```diff
@@ -0,0 +1,22 @@
+﻿<ResourceDictionary
```
dotMorten, line 1: Same here. I'd declare these are general resources in generic.xaml.

dotMorten (changes requested):
I'm not really a fan of the `.Includes` pattern added here. These are just resources. Make resources file and reference it first from Generic.xaml

src/Toolkit/Toolkit.WinUI/UI/Controls/FloorFilter/FloorFilter.Theme.xaml
```diff
@@ -4,6 +4,9 @@
     xmlns:controls="using:Esri.ArcGISRuntime.Toolkit.UI.Controls"
     xmlns:internal="using:Esri.ArcGISRuntime.Toolkit.Internal"
     xmlns:toolkit="using:Esri.ArcGISRuntime.Toolkit.UI.Controls">
+    <ResourceDictionary.MergedDictionaries>
+        <ResourceDictionary Source="../../../Themes/SharedResources.xaml" />
+    </ResourceDictionary.MergedDictionaries>
```
dotMorten, line 9: Right approach, but this should be in the first part of Generic.xaml (before FloorFilter.Theme.xaml or anything else relying on the resources is referenced). The ensures resources aren't imported twice by any xaml files.

src/Toolkit/Toolkit.WinUI/UI/Controls/MeasureToolbar/MeasureToolbar.Theme.xaml
```diff
@@ -1,32 +1,10 @@
...
-                         UriSource="ms-appx:///Esri.ArcGISRuntime.Toolkit/Assets/MeasureFeatureDark.png" />
-        </ResourceDictionary>
-    </ResourceDictionary.ThemeDictionaries>
+﻿<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:controls="using:Esri.ArcGISRuntime.Toolkit.UI.Controls">
+    <ResourceDictionary.MergedDictionaries>
+        <ResourceDictionary Source="../../../Themes/SharedResources.xaml" />
+    </ResourceDictionary.MergedDictionaries>
```
dotMorten, line 7: Same.

dotMorten (commented):
Nice work. Just a small tweak needed.

src/Toolkit/Toolkit.WinUI/UI/Controls/FloorFilter/FloorFilter.Theme.xaml
nCastle1: I tried that approach a few different ways, but I see crashes for missing keys at run time.

dotMorten: In that case these resources aren't shared, and should be added to `FloorFilter.Theme.xaml`.

src/Toolkit/Toolkit.WinUI/Themes/SharedResources.xaml
```diff
@@ -0,0 +1,20 @@
...
+            <BitmapImage x:Key="MeasureLengthImage" UriSource="ms-appx:///Esri.ArcGISRuntime.Toolkit.WinUI/Assets/MeasureLength.png" />
+            <BitmapImage x:Key="MeasureAreaImage" UriSource="ms-appx:///Esri.ArcGISRuntime.Toolkit.WinUI/Assets/MeasureArea.png" />
+            <BitmapImage x:Key="MeasureFeatureImage" UriSource="ms-appx:///Esri.ArcGISRuntime.Toolkit.WinUI/Assets/MeasureFeature.png" />
+        </ResourceDictionary>
+        <ResourceDictionary x:Key="Dark">
+            <BitmapImage x:Key="MeasureLengthImage" UriSource="ms-appx:///Esri.ArcGISRuntime.Toolkit.WinUI/Assets/MeasureLengthDark.png" />
+            <BitmapImage x:Key="MeasureAreaImage" UriSource="ms-appx:///Esri.ArcGISRuntime.Toolkit.WinUI/Assets/MeasureAreaDark.png" />
+            <BitmapImage x:Key="MeasureFeatureImage" UriSource="ms-appx:///Esri.ArcGISRuntime.Toolkit.WinUI/Assets/MeasureFeatureDark.png" />
+        </ResourceDictionary>
+    </ResourceDictionary.ThemeDictionaries>
```
dotMorten, line 19: Aren't these only used by the measure tool? If so, only declare those in its xaml resource file.

src/Toolkit/Toolkit.WinUI/UI/Controls/FloorFilter/FloorFilter.Resources.xaml
```diff
@@ -0,0 +1,8 @@
+﻿<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:controls="using:Esri.ArcGISRuntime.Toolkit.UI.Controls"
+    xmlns:internal="using:Esri.ArcGISRuntime.Toolkit.Internal"
+    xmlns:toolkit="using:Esri.ArcGISRuntime.Toolkit.UI.Controls">
+    <FontFamily x:Key="CalciteIconFont">ms-appx:///Esri.ArcGISRuntime.Toolkit.WinUI/Assets/calcite-ui-icons-24.ttf#calcite-ui-icons-24</FontFamily>
```
dotMorten, line 7: Just to nitpick: Why not just put this single line in `FloorFilter.Theme.xaml` instead of adding an entire file? (like you had it?) Also all the namespace declarations are probably not necessary.
The original idea of moving common resources to a common file really only applies if we could put it in a common resource file, but you said that didn't actually work, and if resources aren't really common after all (as in resources used by multiple controls), there's not much point in splitting things out into separate files. I was under the impression that the font file was used in multiple controls, and therefore should be in a single common place before any of the other theme files gets imported. Is floorfilter the only one using it?
Similar comment for the measure toolbar.

Might be simpler just to fix the URIs and not start moving a bunch of files around at this point.

dotMorten (approved):
Ignore my last review. My bad. Looks good!

PR #461, opened 2022-10-10
Handle DPI changes in SymbolDisplay

### LINKED ISSUE #451: MAUI Symbol Display - always detects display as low resolution
`private double GetScaleFactor() => Window?.DisplayDensity ?? 1d;` is used in SymbolDisplay to determine the scale factor when rendering the symbol. On MAUI, as currently implemented, Window is always null, so SymbolDisplay renders blurry images on high-DPI devices.

nCastle1:
Resolves #451

### DIFF
--- a/src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
+++ b/src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
@@ -14,6 +14,7 @@
 //  *   limitations under the License.
 //  ******************************************************************************/
 
+using System;
 using Esri.ArcGISRuntime.Symbology;
 using Esri.ArcGISRuntime.Toolkit.Internal;
 
@@ -25,12 +26,12 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui;
 public partial class SymbolDisplay : TemplatedView
 {
     private WeakEventListener<System.ComponentModel.INotifyPropertyChanged, object?, System.ComponentModel.PropertyChangedEventArgs>? _inpcListener;
-    private WeakEventListener<System.ComponentModel.INotifyPropertyChanged, object?, System.ComponentModel.PropertyChangedEventArgs>? _thisWindowInpcListener;
+    private WeakEventListener<Window, object, DisplayDensityChangedEventArgs>? _displayDensityChangedListener;
     private Task? _currentUpdateTask;
     private bool _isRefreshRequired;
     private static readonly ControlTemplate DefaultControlTemplate;
     private Image? image;
-    private double _lastScaleFactor = 1;
+    private double _lastScaleFactor = double.NaN;
 
     static SymbolDisplay()
     {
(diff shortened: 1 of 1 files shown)


src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -176,4 +178,28 @@ private async void Refresh()
     /// </summary>
     [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
     public event System.EventHandler? SourceUpdated;
+
+    private void SymbolDisplay_PropertyChanged(object? sender, System.ComponentModel.PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(Window) && Window != null)
+        {
+            Refresh();
+            Window.DisplayDensityChanged += Window_DisplayDensityChanged;
```
dotMorten, line 187: This needs to be a weak event, or this control will stay resident until the window goes out of scope (which usually is never).
Also if window is changing, you'd want to unhook the old handler - that's easier to do with the weak handler which has a reference to the old one.

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -176,4 +178,28 @@ private async void Refresh()
     /// </summary>
     [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
     public event System.EventHandler? SourceUpdated;
+
+    private void SymbolDisplay_PropertyChanged(object? sender, System.ComponentModel.PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(Window) && Window != null)
+        {
+            Refresh();
```
dotMorten, line 186: I'd suggest only doing this if the DPI is different from earlier.

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -176,4 +178,28 @@ private async void Refresh()
...
+    }
+
+    private void SymbolDisplay_Unloaded(object? sender, EventArgs e)
+    {
+        if (Window != null)
+        {
+            Window.DisplayDensityChanged -= Window_DisplayDensityChanged;
+        }
+        this.Unloaded -= SymbolDisplay_Unloaded;
+        this.PropertyChanged -= SymbolDisplay_PropertyChanged;
```
dotMorten, line 198: You should only unhook here if you hook up in loaded. Since you're doing it in the constructor though, this would mean load->unload->load would not work. I don't think you need to unhook though - the event handler goes out of scope with the control.

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -176,4 +178,28 @@ private async void Refresh()
...
+            Window.DisplayDensityChanged += Window_DisplayDensityChanged;
+        }
+    }
+
+    private void SymbolDisplay_Unloaded(object? sender, EventArgs e)
+    {
+        if (Window != null)
+        {
+            Window.DisplayDensityChanged -= Window_DisplayDensityChanged;
+        }
```
dotMorten, line 196: Better to rely on weak event handler, and then here releasing the weak handler, rather than relying on Window still being around

dotMorten (changes requested):

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -47,6 +49,18 @@ static SymbolDisplay()
...
         ControlTemplate = DefaultControlTemplate;
+        _thisWindowInpcListener = new WeakEventListener<System.ComponentModel.INotifyPropertyChanged, object?, System.ComponentModel.PropertyChangedEventArgs>(this)
+        {
+            OnEventAction = (instance, source, eventArgs) =>
+            {
+                if (eventArgs.PropertyName == nameof(Window) && Window != null && _lastScaleFactor != GetScaleFactor())
+                {
+                    Refresh();
+                }
+            },
```
dotMorten, line 60: One more issue here. The reference to `Window`, `_lastScaleFactor` and `Refresh` creates hard references to `this`, so this doesn't actually act as a weak reference.  You'll want to get at these through the `instance`/`source` property.
Otherwise looking good

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -137,7 +152,6 @@ private async Task UpdateSwatchAsync()
         }
     }
 
-    // Known issue - see toolkit issue #451
     private double GetScaleFactor() => Window?.DisplayDensity ?? 1d;
```
dotMorten, line 155: I wonder if this code could be called before we have a window, and then will be called again later. Of so, perhaps we should just return NaN here and not generate a swatch until the window gets established?

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -120,6 +148,12 @@ private async Task UpdateSwatchAsync()
             try
             {
                 var scale = GetScaleFactor();
+                _lastScaleFactor = scale;
+
+                if (double.IsNaN(scale))
+                {
+                    throw new Exception();
+                }
```
dotMorten, line 156: Since this is not exceptional, just return here instead of throwing (and generally we don't throw the base exception type and with no message)

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -47,6 +50,31 @@ static SymbolDisplay()
...
+    private void SymbolDisplay_PropertyChanged(object? sender, System.ComponentModel.PropertyChangedEventArgs e)
+    {
+        if (e.PropertyName == nameof(Window) && Window != null)
+        {
+            if (_lastScaleFactor != GetScaleFactor())
+            {
+                Refresh();
+            }
+
+            _displayDensityChangedListener = new WeakEventListener<Window, object, DisplayDensityChangedEventArgs>(Window)
```
dotMorten, line 65: Add `_displayDensityChangedListener?.Detach()` before creating a new one, which ensures immediate and full cleanup of the previous listener, and avoid risk of firing multiple times.

dotMorten (approved):
Just a couple minor things left to tweak - otherwise good to go.

PR #462, opened 2022-10-21
Revise weak event listener

### LINKED ISSUE #365: Update WeakEventListener

nCastle1:
Updates the weak event listener internal API to be clearer to consumers. Also updates usage of weak event listener.

Resolves #365

### DIFF
--- a/src/Toolkit/Toolkit/WeakEventListener.cs
+++ b/src/Toolkit/Toolkit/WeakEventListener.cs
@@ -22,56 +22,64 @@ namespace Esri.ArcGISRuntime.Toolkit.Internal
     /// Implements a weak event listener that allows the owner to be garbage
     /// collected if its only remaining link is an event handler.
     /// </summary>
-    /// <typeparam name="TInstance">Type of instance listening for the event.</typeparam>
-    /// <typeparam name="TSource">Type of source for the event.</typeparam>
+    /// <typeparam name="TListeningInstance">Type of instance listening for the event.</typeparam>
+    /// <typeparam name="TEventRaisingSource">Type of source for the event.</typeparam>
+    /// <typeparam name="TEventSender">Type of sender for the event handler.</typeparam>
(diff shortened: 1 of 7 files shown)


src/Toolkit/Toolkit/WeakEventListener.cs
```diff
@@ -22,56 +22,64 @@ namespace Esri.ArcGISRuntime.Toolkit.Internal
...
-    /// <typeparam name="TSource">Type of source for the event.</typeparam>
+    /// <typeparam name="TListeningInstance">Type of instance listening for the event.</typeparam>
+    /// <typeparam name="TEventRaisingSource">Type of source for the event.</typeparam>
+    /// <typeparam name="TEventSender">Type of sender for the event handler.</typeparam>
     /// <typeparam name="TEventArgs">Type of event arguments for the event.</typeparam>
     [SuppressMessage("Microsoft.Performance", "CA1812:AvoidUninstantiatedInternalClasses", Justification = "Used as link target in several projects.")]
 #pragma warning disable SA1402 // File may only contain a single class
-    internal class WeakEventListener<TInstance, TSource, TEventArgs>
+    internal class WeakEventListener<TListeningInstance, TEventRaisingSource, TEventSender, TEventArgs>
 #pragma warning restore SA1402 // File may only contain a single class
```
mstefarov, line 32: You can get rid of this `#pragma` -- there is only one class left in this file.

mstefarov (approved):
Looks good! I'm glad the leaks are getting fixed 👍

PR #486, opened 2023-01-19
Update OnApplyTemplate() in BasemapGallery.Appearance.cs 

soba-noodle:
Under NetFramework 472, BasemapGallery does not work when started as Collapsed.   This update will allow BasemapGallery starts as Collapsed and when Visibility changed to Visible  basemap items will be shown.


dotMorten:
Thank you for this. Are you not seeing this happen on .NET 6? If so I think this behavior might actually be due to a .NET Framework bug we just had to add a workaround for, and not a toolkit issue.

soba-noodle:
I confirmed that this won't happen in NET6.  My fix I submitted works for
both Framework and Net6.

Tosa

dotMorten:
@soba-noodle Thanks for confirming. This is actually caused by a .NET Framework bug, that we have worked around in 200.1, so other tools (like this one) won't need to work around it.

soba-noodle:
Can you tell when 200.1 is released?   We like to release our app ASAP.

Thanks.

Tosa

PR #488, opened 2023-02-10
Basemap URI null check

williambohrmann3:
Two basemaps both with null URI's were incorrectly returning as true. Different basemaps can both have null URI's.


mstefarov (approved):

williambohrmann3:
Addresses #489

PR #522, opened 2023-09-11
popup: Implement TextPopupElement view for MAUI

mstefarov:
<table><tr><td>

</td><td>

</td></tr><tr><td>

</td><td>

</td></tr></table>

TextElementPopupView is coming to MAUI!

- Text maps to Labels with formatted Spans.
- Blocks (e.g. paragraphs, table cells, list items) map to StackLayouts.
- Lists map to grids (one column for markers, another for content).
- Tables map to grids (with some extra logic to handle row/column assignment and rowspans/colspans).
- Images are loaded through RuntimeImage to take advantage of our authentication and caching stack.

### Known limitations
- Text occasionally has extra space between lines or words.  This is due to MAUI implementation being more sensitive to inconsistent newlines/whitespace than WPF's FlowDocument-based viewer.  I am a preparing a follow-up PR that adds a document tree simplification pass to the HTML parser that will address this.
- Text links are not clickable on some platforms due to https://github.com/dotnet/maui/issues/4734 (might be fixed before NET 8 GA)
- Subscript (`<sub>`) and superscript (`<sup>`) do not have any visual effect, because MAUI does not expose any API to control this font feature.
- Table columns are always equally-spaced, which might inefficiently distribute space between columns with different amounts of content. "Auto" column width does not work well in MAUI because it prevents line-wrapping and causes the table to stretch out-of-bounds.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.Maui.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.Maui.cs
@@ -17,6 +17,11 @@
 #if MAUI
 using Microsoft.Maui.Controls.Internals;
 using Esri.ArcGISRuntime.Mapping.Popups;
+using Esri.ArcGISRuntime.Toolkit.Internal;
+using Microsoft.Maui.ApplicationModel;
+using Esri.ArcGISRuntime.UI;
+using Grid = Microsoft.Maui.Controls.Grid;
+using RuntimeImageExtensions = Esri.ArcGISRuntime.Maui.RuntimeImageExtensions;
 
 namespace Esri.ArcGISRuntime.Toolkit.Maui.Primitives
 {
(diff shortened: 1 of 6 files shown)


src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.Maui.cs
```diff
@@ -49,17 +55,422 @@ private static object BuildDefaultTemplate()
...
+
+                case MarkupType.Divider:
+                    return new BoxView { HeightRequest = 1, Color = Colors.Gray }; // TODO: Do we need to set a color?
+
+                case MarkupType.Table:
+                    return ConvertTableToGrid(node);
+
+                case MarkupType.Image:
+                    var imageElement = new Image();
+                    if (Uri.TryCreate(node.Content, UriKind.Absolute, out var imgUri))
```
dotMorten, line 194: Note that his can also begin with `data:image/`. See https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/1e8ad8a249796bea954778bcb5360465b1ae430e/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupMediaView.cs#L89-L92

dotMorten (commented):
Looking great. Really only found one minor oversight

src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.Maui.cs
mstefarov: Looks like I missed this from both MAUI and WPF text element views.  Fixed!

dotMorten (approved):

PR #528, opened 2023-10-12
Adds a GeoView controller to help with common MVVM challenges

dotMorten:
Helper class for allowing you to perform view operations on the MapView from your ViewModel, through an attached proxy-object that ensures you keep VM and View separated.

Design choices:
- Basic controller for most common operations like setting viewpoint, performing identify and showing callouts. No specific mapview or sceneview operation.
- Extensible, so you can add your own custom map and scene operations, or interface for allowing testing (see WPF sample)
- No events exposed - use `EventToCommand` or similar already established approaches to forward view events back to the view model (see samples)

### DIFF
--- a/src/Toolkit/Toolkit/GeoViewController.cs
+++ b/src/Toolkit/Toolkit/GeoViewController.cs
@@ -168,11 +168,13 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
         public virtual Task<IReadOnlyList<IdentifyLayerResult>> IdentifyLayersAsync(Point screenPoint, double tolerance, bool returnPopupsOnly = false, CancellationToken cancellationToken = default) =>
             ConnectedView?.IdentifyLayersAsync(screenPoint, tolerance, returnPopupsOnly, cancellationToken) ??
                 Task.FromResult<IReadOnlyList<IdentifyLayerResult>>(Array.Empty<IdentifyLayerResult>());
-        
+
         /// <inheritdoc cref="GeoView.IdentifyLayerAsync(Layer, Point, double, bool, CancellationToken)"/>
-        public virtual Task<IdentifyLayerResult> IdentifyLayerAsync(Layer layer, Point screenPoint, double tolerance, bool returnPopupsOnly = false, CancellationToken cancellationToken = default) =>
-            ConnectedView?.IdentifyLayerAsync(layer, screenPoint, tolerance, returnPopupsOnly, cancellationToken) ??
-                Task.FromResult<IdentifyLayerResult>(null!);
+        public virtual async Task<IdentifyLayerResult?> IdentifyLayerAsync(Layer layer, Point screenPoint, double tolerance, bool returnPopupsOnly = false, CancellationToken cancellationToken = default)
+        {
+            if (ConnectedView is null) return null;
+            return await ConnectedView.IdentifyLayerAsync(layer, screenPoint, tolerance, returnPopupsOnly, cancellationToken).ConfigureAwait(false);
+        }
 
         /// <inheritdoc cref="GeoView.IdentifyGraphicsOverlaysAsync(Point, double, bool, long)" />
         public virtual Task<IReadOnlyList<IdentifyGraphicsOverlayResult>> IdentifyGraphicsOverlaysAsync(Point screenPoint, double tolerance, bool returnPopupsOnly = false, long maximumResultsPerOverlay = 1) =>
@@ -180,9 +182,12 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
                 Task.FromResult<IReadOnlyList<IdentifyGraphicsOverlayResult>>(Array.Empty<IdentifyGraphicsOverlayResult>());
 
         /// <inheritdoc cref="GeoView.IdentifyGraphicsOverlaysAsync(Point, double, bool, long)" />
-        public virtual Task<IdentifyGraphicsOverlayResult> IdentifyGraphicsOverlayAsync(GraphicsOverlay overlay, Point screenPoint, double tolerance, bool returnPopupsOnly = false, long maximumResults = 1) =>
-            ConnectedView?.IdentifyGraphicsOverlayAsync(overlay, screenPoint, tolerance, returnPopupsOnly, maximumResults) ??
-                Task.FromResult<IdentifyGraphicsOverlayResult>(null!);
+        public virtual async Task<IdentifyGraphicsOverlayResult?> IdentifyGraphicsOverlayAsync(GraphicsOverlay overlay, Point screenPoint, double tolerance, bool returnPopupsOnly = false, long maximumResults = 1)
+        {
+            if (ConnectedView is null) return null;
+            return await ConnectedView.IdentifyGraphicsOverlayAsync(overlay, screenPoint, tolerance, returnPopupsOnly, maximumResults).ConfigureAwait(false);
+        }
+
         #endregion Identify
 
         #region Callouts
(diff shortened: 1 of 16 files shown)


src/Toolkit/Toolkit/GeoViewController.cs
```diff
@@ -0,0 +1,279 @@
...
+        /// <param name="geoView">The <see cref="GeoView"/> from which to read the property value.</param>
+        /// <returns>The value of the <see cref="GeoViewController.GeoViewController"/> XAML attached property on the target element.</returns>
+        public static GeoViewController? GetGeoViewController(GeoViewDPType geoView) => geoView.GetValue(GeoViewControllerProperty) as GeoViewController;
+
+        /// <summary>
+        /// Sets the value of the <see cref="GeoViewController.GeoViewController"/> XAML attached property on the specified object.
+        /// </summary>
+        /// <param name="geoView">The target <see cref="GeoView"/> on which to set the <see cref="GeoViewController.GeoViewController"/> XAML attached property.</param>
+        /// <param name="value">The property value to set.</param>
+        public static void SetGeoViewController(GeoViewDPType geoView, GeoViewController? value) => geoView.SetValue(GeoViewControllerProperty, value);
```
mstefarov, line 256: Does `geoView` parameter need a null-check in these methods?  Maybe just something to throw ArgumentNullException instead of NRE.

src/Toolkit/Toolkit/GeoViewController.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        /// <inheritdoc cref="GeoView.ShowCalloutForGeoElement(GeoElement, Point, CalloutDefinition)" />
+        public virtual void ShowCalloutForGeoElement(GeoElement element, Point tapPosition, CalloutDefinition definition) => ConnectedView?.ShowCalloutForGeoElement(element, tapPosition, definition);
+        #endregion Callouts
+
+#if MAUI
+        /// <summary>
+        /// Identifies the <see cref="GeoViewController"/> attached property.
+        /// </summary>
+        public static BindableProperty GeoViewControllerProperty =
```
mstefarov, line 204: This BindableProperty field can be readonly.
```suggestion
        public static readonly BindableProperty GeoViewControllerProperty =
```

src/Toolkit/Toolkit/GeoViewController.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        private void GeoView_Unloaded(GeoView sender)
+        {
+            Debug.Assert(sender == ConnectedView && ConnectedView is not null);
+            OnGeoViewUnloaded(sender);
+        }
+
+        #region Viewpoints
+
+        /// <inheritdoc cref="GeoView.GetCurrentViewpoint(ViewpointType)"/>
```
mstefarov, line 151: I didn't know we can just inherit docs across unrelated classes. Neat!

src/Toolkit/Toolkit/GeoViewController.cs
```diff
@@ -0,0 +1,279 @@
...
+        public virtual void SetViewpoint(Viewpoint viewpoint) => ConnectedView?.SetViewpoint(viewpoint);
+
+        #endregion Viewpoints
+
+        #region Identify
+        
+        /// <inheritdoc cref="GeoView.IdentifyLayersAsync(Point, double, bool, CancellationToken)"/>
+        public virtual Task<IReadOnlyList<IdentifyLayerResult>> IdentifyLayersAsync(Point screenPoint, double tolerance, bool returnPopupsOnly = false, CancellationToken cancellationToken = default) =>
+            ConnectedView?.IdentifyLayersAsync(screenPoint, tolerance, returnPopupsOnly, cancellationToken) ??
+                Task.FromResult<IReadOnlyList<IdentifyLayerResult>>(new List<IdentifyLayerResult>().AsReadOnly());
```
mstefarov, line 170: There's a simpler way:
```suggestion
                Task.FromResult<IReadOnlyList<IdentifyLayerResult>>(Array.Empty<IdentifyLayerResult>());
```

src/Toolkit/Toolkit/GeoViewController.cs
```diff
@@ -0,0 +1,279 @@
...
+        
+        /// <inheritdoc cref="GeoView.IdentifyLayerAsync(Layer, Point, double, bool, CancellationToken)"/>
+        public virtual Task<IdentifyLayerResult> IdentifyLayerAsync(Layer layer, Point screenPoint, double tolerance, bool returnPopupsOnly = false, CancellationToken cancellationToken = default) =>
+            ConnectedView?.IdentifyLayerAsync(layer, screenPoint, tolerance, returnPopupsOnly, cancellationToken) ??
+                Task.FromResult<IdentifyLayerResult>(null!);
+
+        /// <inheritdoc cref="GeoView.IdentifyGraphicsOverlaysAsync(Point, double, bool, long)" />
+        public virtual Task<IReadOnlyList<IdentifyGraphicsOverlayResult>> IdentifyGraphicsOverlaysAsync(Point screenPoint, double tolerance, bool returnPopupsOnly = false, long maximumResultsPerOverlay = 1) =>
+            ConnectedView?.IdentifyGraphicsOverlaysAsync(screenPoint, tolerance, returnPopupsOnly, maximumResultsPerOverlay) ??
+                Task.FromResult<IReadOnlyList<IdentifyGraphicsOverlayResult>>(new List<IdentifyGraphicsOverlayResult>().AsReadOnly());
```
mstefarov, line 180:
```suggestion
                Task.FromResult<IReadOnlyList<IdentifyGraphicsOverlayResult>>(Array.Empty<IdentifyGraphicsOverlayResult>());
```

src/Toolkit/Toolkit/GeoViewController.cs
```diff
@@ -0,0 +1,279 @@
...
+        
+        /// <inheritdoc cref="GeoView.IdentifyLayersAsync(Point, double, bool, CancellationToken)"/>
+        public virtual Task<IReadOnlyList<IdentifyLayerResult>> IdentifyLayersAsync(Point screenPoint, double tolerance, bool returnPopupsOnly = false, CancellationToken cancellationToken = default) =>
+            ConnectedView?.IdentifyLayersAsync(screenPoint, tolerance, returnPopupsOnly, cancellationToken) ??
+                Task.FromResult<IReadOnlyList<IdentifyLayerResult>>(new List<IdentifyLayerResult>().AsReadOnly());
+        
+        /// <inheritdoc cref="GeoView.IdentifyLayerAsync(Layer, Point, double, bool, CancellationToken)"/>
+        public virtual Task<IdentifyLayerResult> IdentifyLayerAsync(Layer layer, Point screenPoint, double tolerance, bool returnPopupsOnly = false, CancellationToken cancellationToken = default) =>
+            ConnectedView?.IdentifyLayerAsync(layer, screenPoint, tolerance, returnPopupsOnly, cancellationToken) ??
+                Task.FromResult<IdentifyLayerResult>(null!);
```
mstefarov, line 175: It's a shame we have to return null for non-nullable task.  But I'm not sure if adding `IdentifyLayerResult.CreateEmpty(ILayerContent)` to the public API is worth it just for MVVM purity.

Maybe this method (and IdentifyGraphicsOverlayAsync) can return a task with a nullable result instead?

src/Toolkit/Toolkit/GeoViewController.cs
dotMorten: It works for intellisense - however I don't think docfx likes it, so be careful.

mstefarov (approved):

PR #539, opened 2024-01-03
`MeasureToolbar`: modify `ArgumentException` message

williambohrmann3:
When registering the `DependencyProperty`s for `LineSketchEditor` and `AreaSketchEditor`, they are both using the same method for the `PropertyMetadata` `PropertyChangedCallback` parameter. This is fine, but currently the exception message when the new value is null doesn't make sense for `AreaSketchEditor`. To remedy this, I would like to remove the unused method and modify the currently used method to make sense in both cases.

Also, I removed some redundant "using" statements and corrected a typo.


williambohrmann3:
Any idea why the job might fail if the solution is building for me locally?

williambohrmann3:
All resolved - updated the certificate for UWP, changes included in this PR.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -597,15 +591,15 @@ private static void OnMapViewPropertyChanged(DependencyObject d, DependencyPrope
         /// Identifies the <see cref="LineSketchEditor"/> dependency property.
         /// </summary>
         public static readonly DependencyProperty LineSketchEditorProperty =
-            DependencyProperty.Register(nameof(LineSketchEditor), typeof(SketchEditor), typeof(MeasureToolbar), new PropertyMetadata(null, OnLineSketchEditorPropertyChanged));
+            DependencyProperty.Register(nameof(LineSketchEditor), typeof(SketchEditor), typeof(MeasureToolbar), new PropertyMetadata(null, OnSketchEditorPropertyChanged));
 
-        private static void OnLineSketchEditorPropertyChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+        private static void OnSketchEditorPropertyChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
```
dotMorten, line 596: Why this change? It's literally the 'LineSketchEditor' property. Please undo

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -631,26 +625,7 @@ private static void OnLineSketchEditorPropertyChanged(DependencyObject d, Depend
         /// Identifies the <see cref="AreaSketchEditor"/> dependency property.
         /// </summary>
         public static readonly DependencyProperty AreaSketchEditorProperty =
-            DependencyProperty.Register(nameof(AreaSketchEditor), typeof(SketchEditor), typeof(MeasureToolbar), new PropertyMetadata(null, OnLineSketchEditorPropertyChanged));
```
dotMorten, line 634: This is the only bug. It should call  OnAreaSketchEditorPropertyChanged instead, and don't delete the code below.

dotMorten (changes requested):

dotMorten (approved):

PR #543, opened 2024-01-10
BookmarksViewSample fixes

williambohrmann3:
- Fixes the WinUI app crash
- Replaces the scene two online item
- Adds comments about the Web Scene specifications

### DIFF
--- a/src/Samples/Toolkit.SampleApp.UWP/Samples/BookmarksView/BookmarksViewSample.xaml.cs
+++ b/src/Samples/Toolkit.SampleApp.UWP/Samples/BookmarksView/BookmarksViewSample.xaml.cs
@@ -126,14 +126,18 @@ namespace Esri.ArcGISRuntime.Toolkit.SampleApp.Samples.BookmarksView
             BookmarksView.ItemTemplate = template;
         }
 
-        private void AddSelectionListener_Click(object sender, RoutedEventArgs e)
+        private void SelectionListenerToggle_Click(object sender, RoutedEventArgs e)
         {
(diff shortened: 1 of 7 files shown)


src/Samples/Toolkit.SampleApp.UWP/Samples/BookmarksView/BookmarksViewSample.xaml.cs
```diff
@@ -136,7 +138,13 @@ private void RemoveSelectionListener_Click(object sender, RoutedEventArgs e)
 
         private async void BookmarkSelected(object sender, Bookmark bookmark)
         {
+#if WINUI
+            var dialog = new MessageDialog($"{bookmark.Name} Selected!");
+            WinRT.Interop.InitializeWithWindow.Initialize(dialog, App.CurrentWindowHandle);
+            await dialog.ShowAsync();
```
mstefarov, line 144: Maybe we should use **ContentDialog** instead of **MessageDialog**, as [Microsoft suggests](https://learn.microsoft.com/en-us/uwp/api/windows.ui.popups.messagedialog?view=winrt-22621)?

[Setting the XAML root](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.contentdialog?view=windows-app-sdk-1.4#set-the-xamlroot) for ContentDialog looks a lot neater than dealing with window handles.

dotMorten (approved):

mstefarov (approved):

PR #546, opened 2024-01-31
FeatureFormsView

dotMorten:
Adds a WPF control to render FeatureForms.

TODO:
 - Barcode scanner support
 - [x] ~~weak listeners on INPC~~

image
image
image

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/ComboboxFormInputView.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/ComboboxFormInputView.cs
@@ -1,6 +1,7 @@
 ﻿#if WPF
 using Esri.ArcGISRuntime.Data;
 using Esri.ArcGISRuntime.Mapping.FeatureForms;
+using Esri.ArcGISRuntime.Toolkit.Internal;
 using System.ComponentModel;
 
 namespace Esri.ArcGISRuntime.Toolkit.Primitives
@@ -11,6 +12,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Primitives
     [TemplatePart(Name ="Selector", Type = typeof(System.Windows.Controls.Primitives.Selector))]
     public class ComboBoxFormInputView : Control
     {
+        private WeakEventListener<ComboBoxFormInputView, INotifyPropertyChanged, object?, PropertyChangedEventArgs>? _elementPropertyChangedListener;
         private System.Windows.Controls.Primitives.Selector? _selector;
 
         /// <summary>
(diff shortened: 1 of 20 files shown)


src/Toolkit/Toolkit/UI/Controls/FeatureForm/ComboboxFormInputView.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        private void Element_PropertyChanged(object? sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(FieldFormElement.Value))
+            {
+                if (Dispatcher.CheckAccess())
+                    UpdateSelection();
+                else
+                    Dispatcher.Invoke(UpdateSelection);
```
mstefarov, line 83: Doesn't Dispatcher.Invoke already do this check?  I'm pretty sure it executes immediately if we are on the right thread.  Or did you mean to do immediate-or-BeginInvoke behavior, to avoid blocking the INPC-raising thread?

Same question for all the other implementations of Element_PropertyChanged in this PR.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/RadioButtonsFormInputView.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        /// <inheritdoc />
+        protected override void OnSelectionChanged(SelectionChangedEventArgs e)
+        {
+            base.OnSelectionChanged(e);
+            if (Element is null) return;
+            var value = (SelectedItem as CodedValue);
+            foreach(var item in GetItemContainers(this))
+            {
+                item.IsChecked = (item.DataContext == SelectedItem); // || item.DataContext is RadioButtonNullValue && value is null);
```
mstefarov, line 167: The `value` variable is unused. Did you mean to keep this code commented-out?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/DatePickerFormInputView.cs
```diff
@@ -0,0 +1,130 @@
...
+        private void DatePicker_SelectedDateChanged(object? sender, SelectionChangedEventArgs e) => UpdateValue();
+
+        private void UpdateValue()
+        { 
+            if (Element?.Input is DateTimePickerFormInput input && _datePicker != null)
+            {
+                var date = _datePicker.SelectedDate;
+                if(date.HasValue && input.IncludeTime && _timePicker != null && _timePicker.Time.HasValue)
+                {
+                    date = date.Value.Subtract(date.Value.TimeOfDay).Add(_timePicker.Time.Value);
```
mstefarov, line 54: Optional: Another slightly-simpler-way of doing this, using [DateTime.Date](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.date?view=netstandard-2.0):
```suggestion
                    date = date.Value.Date.Add(_timePicker.Time.Value);
```

src/Toolkit/Toolkit/UI/Controls/FeatureForm/TimePicker.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            if (_timeText is null)
+                return;
+            if(DateTime.TryParse(_timeText.Text, null, DateTimeStyles.NoCurrentDateDefault, out DateTime result))
+            {
+                Time = result.TimeOfDay;
+            }
+            else
+            {
+                UpdateText(); //Reset text
```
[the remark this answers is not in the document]
dotMorten: Interesting idea, but the DatePicker has the same reset behavior, and it doesn't "ding"

src/Toolkit/Toolkit/UI/Controls/FeatureForm/ComboboxFormInputView.cs
dotMorten: When performing async evaluation of expressions, core might be coming back on a different thread, so sometimes we need to dispatch back to the UI thread, and but many other times we do not. You're correct [there's some logic](https://github.com/dotnet/wpf/blob/873db676363ce1777e763948481b24ded3513746/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs#L579) to avoid extra work, but this is the pattern we have followed everywhere.

mstefarov (approved):

PR #552, opened 2024-02-22
utilnet: localize trace tool

jnery:
Localize trace tool

- Updated strings in `Text="`, `Content="`, `Header="`, and error messages since all get displayed on the tool.
- There were some subtle string differences in WinUI, MAUI, WPF; for example, "Not enough starting points..", "Networks", "Delete result", "Click on the map.." that are now resolved.
- Text that use symbols 'X', '<', '>', (i.e. `&#xE10A;`) are not localized.
- Text to "Cancel" though different buttons are shared by add/identify starting points and running trace.
- Passenger bugs: 
    - Rename "(Named) trace configurations" => "Trace types". Older toolkit design and Javascript widget were all updated to use "Trace types"
    - Remove null check on `traceParameters` because null check for `SelectedTraceType` guarantees a new traceParameters is always created.
- To test, I ran Maui, WinUI, WPF, UWP `UtilityNetworkTraceTool` sample and visually checked appropriate labels, warnings and error messages are picked up from resx.

### DIFF
--- a/src/Toolkit/Toolkit.Maui/UtilityNetworkTraceTool/UtilityNetworkTraceTool.Appearance.cs
+++ b/src/Toolkit/Toolkit.Maui/UtilityNetworkTraceTool/UtilityNetworkTraceTool.Appearance.cs
@@ -69,6 +69,26 @@ public partial class UtilityNetworkTraceTool
     {
         const string backgroundColor = "{AppThemeBinding Dark=#353535, Light=#F8F8F8}";
         const string foregroundColor = "{AppThemeBinding Dark=#ffffff, Light=#151515}";
+
+        var noUtilityNetworks = Properties.Resources.GetString("UtilityNetworkTraceToolNoUtilityNetworks");
+        var utilityNetworks = Properties.Resources.GetString("UtilityNetworkTraceToolUtilityNetworks");
+        var traceTypes = Properties.Resources.GetString("UtilityNetworkTraceToolTraceTypes");
+        var addStartingPoint = Properties.Resources.GetString("UtilityNetworkTraceToolAddStartingPoint");
+        var cancel = Properties.Resources.GetString("UtilityNetworkTraceToolCancel");
+        var notEnoughStartingPoints = Properties.Resources.GetString("UtilityNetworkTraceToolNotEnoughStartingPoints");
+        var moreThanRequiredStartingPoints = Properties.Resources.GetString("UtilityNetworkTraceToolMoreThanRequiredStartingPoints");
+        var duplicateTrace = Properties.Resources.GetString("UtilityNetworkTraceToolDuplicateTrace");
+        var runTrace = Properties.Resources.GetString("UtilityNetworkTraceToolRunTrace");
+        var noResults = Properties.Resources.GetString("UtilityNetworkTraceToolNoResults");
+        var noFeatureResults = Properties.Resources.GetString("UtilityNetworkTraceToolNoFeatureResults");
+        var noFunctionResults = Properties.Resources.GetString("UtilityNetworkTraceToolNoFunctionResults");
+        var featureResults = Properties.Resources.GetString("UtilityNetworkTraceToolFeatureResults");
+        var functionResults = Properties.Resources.GetString("UtilityNetworkTraceToolFunctionResults");
+        var visualizationOptions = Properties.Resources.GetString("UtilityNetworkTraceToolVisualizationOptions");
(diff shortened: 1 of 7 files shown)


src/Toolkit/Toolkit/UI/Controls/UtilityNetworkTraceTool/UtilityNetworkTraceToolController.cs
```diff
@@ -326,7 +326,7 @@ private void OnGeoModelLoaded(object? sender, EventArgs e)
                     IsLoadingNetwork = true;
                     UtilityNetworks.Clear();
 
-                    var utilityNetworks = (sender is Map map ? map.UtilityNetworks : null) ?? throw new ArgumentException("No UtilityNetworks found.");
+                    var utilityNetworks = (sender is Map map ? map.UtilityNetworks : null) ?? throw new ArgumentException(Properties.Resources.GetString("UtilityNetworkTraceToolNoUtilityNetworks"));
```
dotMorten, line 329: Exception messages shouldn't be localized - only UI displayed to the end user. Goes for everything below here as well.
I'm also a little confused about throwing here. It looks to me like this is an `async void`, and async voids should never be allowed to throw

dotMorten (changes requested):
Looks good except for the exception resources, which don't see to be shown to the end user (or even the end developer since it's in an async void)

src/Toolkit/Toolkit/UI/Controls/UtilityNetworkTraceTool/UtilityNetworkTraceToolController.cs
jnery: Ah you're right exception here is not part of UI (only [Trace](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/main/src/Toolkit/Toolkit/UI/Controls/UtilityNetworkTraceTool/UtilityNetworkTraceToolController.cs#L689) errors are displayed to user) . It seems this one is wrapped in a try-catch that only marks readiness of UI. I can update this to null/empty check instead.

jnery: I left the resource key because it's used in all XAML, say UN collection is cleared or empty.
image

jnery: Actually, is it okay to put back the catch all? Should utility network fail to load when no credential for example, then the try-catch handle it and mark this control as not ready.

dotMorten: I think fixing how we handle exceptions here is a bit out of scope for the specific issue, but something to look into - the main thing to change for now is to not add localizations for exceptions.

dotMorten (approved):

src/Toolkit/Toolkit.Maui/UtilityNetworkTraceTool/UtilityNetworkTraceTool.Appearance.cs
```diff
@@ -86,22 +106,22 @@ static UtilityNetworkTraceTool()
         <RowDefinition Height=""*"" />
     </Grid.RowDefinitions>
     <Frame x:Name=""{nameof(PART_NoNetworksWarning)}"" BorderColor=""{{AppThemeBinding Light=#D83020, Dark=#FE583E}}"" CornerRadius=""4"" Margin=""4"" Grid.RowSpan=""3"" IsVisible=""false"">
-        <Label Text=""No utility networks found."" />
+        <Label Text=""{noUtilityNetworks}"" />
```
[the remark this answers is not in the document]
jnery: Actually, I am not sure what symbols will be used in the translation and sounds like a good idea to replace `""` with `&quot;`. I can't seem to find `cs` file with `&quot;` in toolkit repo but our targets file use it.

@dotMorten,  is this something we'll do for the other MAUI XAML code? Should I address this now in this PR or next round? 

I found a few .NET Maui toolkit implementation issues that I'd like to address in the future because it affected my dev summit demo and I ended up writing my own to match WPF implementation.

- previous traces are cleared
- no color palette to compare aggregated geometry results
- no terminal picker or fraction along slider
- theme issues

If this feedback can wait til I get around to fixing MAUI trace tool issues, I can tackle it then.

mstefarov (approved):
I am okay with merging this now, and addressing _potential_ escaping problems later.

PR #567, opened 2024-04-06
wpf: Improve keyboard navigation for FeatureForms

mstefarov:
I had a go using the WPF FeatureForms sample with keyboard navigation, and found that we had MANY extra tab stops between the actual inputs: on content controls, on items controls, on containers, etc.  Here is the problem that this PR attempts to solve: it took 16 TAB presses to navigate through 5 fields:

2024-04-05_225008 Toolkit SampleApp WPF

This PR removes the useless intermediate tab stops, so that keyboard focus switches directly between inputs.  Now it take 5 TAB presses to navigate 5 fields:

2024-04-05_224823 Toolkit SampleApp WPF


dotMorten (approved):
Nice find!

PR #568, opened 2024-04-06
A helper method to simplify creating dependency properties on all platforms

dotMorten:
Before 

```cs
#if MAUI
        public static readonly BindableProperty ElementProperty =
            BindableProperty.Create(nameof(Element), typeof(TextPopupElement), typeof(TextPopupElementView), null, propertyChanged: (s,o,n) => ((TextPopupElementView)s).OnElementPropertyChanged());
#else
        public static readonly DependencyProperty ElementProperty =
            DependencyProperty.Register(nameof(Element), typeof(TextPopupElement), typeof(TextPopupElementView), new PropertyMetadata(null, (s, e) => ((TextPopupElementView)s).OnElementPropertyChanged()));
#endif
```
after:
```cs
 public static readonly DependencyProperty ElementProperty =
        PropertyHelper.CreateProperty<TextPopupElement, TextPopupElementView>(nameof(Element), null, (s, oldValue, newValue) => s.OnElementPropertyChanged());
```

New helper applied to PopupViewer. More controls to do afterwards


mstefarov (changes requested):
I see a build failure in Esri.ArcGISRuntime.Toolkit.Maui project when this is merged into `main`:
> src\Toolkit\Toolkit\UI\Controls\FeatureForm\FieldFormElementView.cs(26,7,26,25): error CS1537: The using alias 'DependencyProperty' appeared previously in this namespace

mstefarov (approved):

PR #570, opened 2024-04-11
Maui compass

dotMorten:
Uses MAUI UI components to render the compass and gets rid of native views. This creates more consistent rendering, better resizing, and also breaks the dependency on the WinUI Toolkit


src/Toolkit/Toolkit.Maui/Compass/Compass.cs
```diff
@@ -78,19 +173,79 @@ public bool AutoHide
     /// Identifies the <see cref="GeoView"/> Dependency Property.
     /// </summary>
     public static readonly BindableProperty GeoViewProperty =
-        BindableProperty.Create(nameof(Compass.GeoView), typeof(GeoView), typeof(Compass), null, BindingMode.OneWay, propertyChanged: OnGeoViewPropertyChanged);
+        BindableProperty.Create(nameof(Compass.GeoView), typeof(GeoView), typeof(Compass), null, BindingMode.OneWay, propertyChanged: (s, oldValue, newValue) => ((Compass)s).OnGeoViewPropertyChanged(oldValue as GeoView, newValue as GeoView));
+
+    private void OnGeoViewPropertyChanged(GeoView oldValue, GeoView newValue)
```
mstefarov, line 178: I spy nullability warnings!
```suggestion
    private void OnGeoViewPropertyChanged(GeoView? oldValue, GeoView? newValue)
```

src/Toolkit/Toolkit.Maui/Compass/ICompass.cs
```diff
@@ -19,6 +19,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui;
 /// <summary>
 /// Defines the public API for the MAUI compass control.
 /// </summary>
+[Obsolete("No longer in use")]
```
mstefarov, line 22: The interface _is_ still in use, isn't it? The new view implements it.
Perhaps only CompassHandler should be deprecated &mdash; or even removed?

dotMorten: it's not a new view - its the same view - I kept the interface on it for backwards compat reasons, but it isn't "in use"

mstefarov (approved):
I would completely delete CompassHandler.  What's the compatibility benefit of leaving an empty implementation that throws if you use it?  If someone out there is relying on the old  CompassHandler implementation, this doesn't help them -- it just conceals the problem.

Other than that, looks great!

PR #574, opened 2024-06-03
Fix AllDisplayLevelsSelecteded typo

williambohrmann3:
Deprecates `AllDisplayLevelsSelecteded` and replaces with `AllDisplayLevelsSelected`.

### DIFF
--- a/src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
@@ -659,6 +659,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         /// <remarks>
         /// This is used for showing an entire facility in 3D.
         /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
         [Obsolete("This property is deprecated. Use AllDisplayLevelsSelected instead.")]
         public bool AllDisplayLevelsSelecteded
         {
--- a/src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
@@ -341,6 +341,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                     _autoVisibilityWrapper?.SetValue(VisibilityProperty, _controller.ShouldDisplayFloorPicker ? Visibility.Visible : Visibility.Collapsed);
                     break;
                 case nameof(_controller.AllDisplayLevelsSelected):
+                    OnPropertyChanged(nameof(AllDisplayLevelsSelecteded));
                     OnPropertyChanged(nameof(AllDisplayLevelsSelected));
                     break;
             }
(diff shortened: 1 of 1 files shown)


src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
```diff
@@ -673,9 +674,31 @@ public bool AllDisplayLevelsSelecteded
                 }
```
dotMorten, line 674: Duplicate implementation. The old property should forward to the new one.
```cs
public bool AllDisplayLevelsSelecteded
{
        get => AllDisplayLevelsSelected;
        set => AllDisplayLevelsSelected = value;
}
```

src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
```diff
@@ -341,7 +341,7 @@ private void HandleControllerPropertyChanges(object? sender, PropertyChangedEven
                     _autoVisibilityWrapper?.SetValue(VisibilityProperty, _controller.ShouldDisplayFloorPicker ? Visibility.Visible : Visibility.Collapsed);
                     break;
                 case nameof(_controller.AllDisplayLevelsSelected):
-                    OnPropertyChanged(nameof(AllDisplayLevelsSelecteded));
```
dotMorten, line 344: The old property is still changing too.

dotMorten (commented):
Some minor comments. Also note that this binding should now work:
- https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/a6ee2c2e8b217bb9eb6f62f3e3897e3b95608be8/src/Toolkit/Toolkit.WPF/UI/Controls/FloorFilter/FloorFilter.Theme.xaml#L341
- https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/a6ee2c2e8b217bb9eb6f62f3e3897e3b95608be8/src/Toolkit/Toolkit.WinUI/UI/Controls/FloorFilter/FloorFilter.Theme.xaml#L556

It would be good to verify that it acts like expected

src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
williambohrmann3: Do you mean we want the old property to keep changing? So undo the deletion of line 344? I'm not sure I follow.

mstefarov: For the sake of backwards compatibility, for all the software that has worked around our old misspelling, we should keep firing a PCE for that old property name too &mdash; in addition to the new correctly-spelled PCE.

src/Toolkit/Toolkit.UI.Controls/FloorFilter/FloorFilter.cs
```diff
@@ -658,7 +659,20 @@ public void SetSelectedLevelWithoutZoom(FloorLevel newLevel)
         /// <remarks>
         /// This is used for showing an entire facility in 3D.
         /// </remarks>
+        [Obsolete("This property is deprecated. Use AllDisplayLevelsSelected instead.")]
         public bool AllDisplayLevelsSelecteded
```
mstefarov, line 663: I would add `[EditorBrowsable(EditorBrowsableState.Never)]` to hide this from IntelliSense

mstefarov (approved):
I suggest one more attribute, but otherwise this PR looks good to me

williambohrmann3:
@dotMorten verified bindings

dotMorten (approved):

PR #579, opened 2024-06-11
BugFix: BasemapGallery loading demo data in production release

prathameshnarkhede:
### Description

This pull request includes several updates to the `BasemapGallery` and `BasemapGalleryItem` classes. Here are the key changes:
1.	Update to `BasemapGallery.Appearance.cs`: The visibility of the `PART_LoadingScrim` Grid has been set to False. This change will hide the loading scrim by default, providing a cleaner user interface while the basemaps are loading.
2.	Refactoring of `LoadBasemaps` method: The `LoadBasemaps` method previously present in both `BasemapGallery.cs` and `BasemapGallery.Windows.cs` has been removed. Its functionality has been integrated into the `BasemapGallery_Loaded` method. This change streamlines the loading process and reduces code duplication.
3.	Enhancement to `BasemapGalleryItem.cs`: The Equals method in this class has been updated. In addition to the existing checks, it now also verifies the equality of the `BaseLayers` count and checks for the existence of all layers in `Basemap` in the `other` object. This enhancement ensures a more thorough comparison of `BasemapGalleryItem` instances.

These updates aim to improve the efficiency of the basemap loading process and the accuracy of basemap item comparisons. Please review the changes and let me know if you have any questions or feedback.

### DIFF
--- a/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
+++ b/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
@@ -26,10 +26,13 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui;
 /// If connected to a GeoView, changing the basemap selection will change the connected Map or Scene's basemap.
 /// Only basemaps whose spatial reference matches the map or scene's spatial reference can be selected for display.
 /// </remarks>
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable
 public partial class BasemapGallery
+#pragma warning restore CA1001 // Types that own disposable fields should be disposable
 {
     private CollectionView? _listView;
     private readonly BasemapGalleryController _controller;
+    private CancellationTokenSource? _loadCancellationTokenSource;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BasemapGallery"/> class.
@@ -44,11 +47,17 @@ public partial class BasemapGallery
         Loaded += BasemapGallery_Loaded;
     }
 
-    private void BasemapGallery_Loaded(object? sender, EventArgs e)
+    private async void BasemapGallery_Loaded(object? sender, EventArgs e)
     {
         if (AvailableBasemaps is null)
         {
-            _ = _controller.LoadFromDefaultPortal();
+            _loadCancellationTokenSource = new CancellationTokenSource();
+            try
+            {
+                await _controller.LoadFromDefaultPortal(_loadCancellationTokenSource.Token);
+            }
+            catch (OperationCanceledException)
+            { }
         }
     }
 
@@ -193,7 +202,11 @@ public partial class BasemapGallery
     public IList<BasemapGalleryItem>? AvailableBasemaps
     {
         get => GetValue(AvailableBasemapsProperty) as IList<BasemapGalleryItem>;
-        set => SetValue(AvailableBasemapsProperty, value);
+        set
+        {
+            SetValue(AvailableBasemapsProperty, value);
+            _loadCancellationTokenSource?.Cancel();
+        }
     }
 
     /// <summary>
--- a/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
+++ b/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
@@ -46,21 +46,10 @@ public partial class BasemapGallery
 
     private void BasemapGallery_Loaded(object? sender, EventArgs e)
     {
-        LoadBasemaps();
-    }
-
-    private void LoadBasemaps()
-    {
-        _controller.IsLoading = true;
-        if (AvailableBasemaps == null)
+        if (AvailableBasemaps is null)
         {
             _ = _controller.LoadFromDefaultPortal();
         }
-        else
-        {
-            _controller.AvailableBasemaps = AvailableBasemaps;
-        }
-        _controller.IsLoading = false;
     }
 
     private void HandleControllerPropertyChanged(object? sender, PropertyChangedEventArgs e)
--- a/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
+++ b/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
@@ -41,7 +41,26 @@ public partial class BasemapGallery
         ListItemTemplate = DefaultListDataTemplate;
         GridItemTemplate = DefaultGridDataTemplate;
         ControlTemplate = DefaultControlTemplate;
-        _ = _controller.LoadFromDefaultPortal();
+        Loaded += BasemapGallery_Loaded;
+    }
+
+    private void BasemapGallery_Loaded(object? sender, EventArgs e)
+    {
+        LoadBasemaps();
+    }
+
+    private void LoadBasemaps()
+    {
+        _controller.IsLoading = true;
+        if (AvailableBasemaps == null)
+        {
+            _ = _controller.LoadFromDefaultPortal();
+        }
+        else
+        {
+            _controller.AvailableBasemaps = AvailableBasemaps;
+        }
+        _controller.IsLoading = false;
     }
 
     private void HandleControllerPropertyChanged(object? sender, PropertyChangedEventArgs e)
--- a/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
+++ b/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
@@ -49,6 +49,9 @@ public partial class BasemapGallery
 
     private async void BasemapGallery_Loaded(object? sender, EventArgs e)
     {
+        // Unsubscribe from the Loaded event to ensure this only runs once.
+        Loaded -= BasemapGallery_Loaded;
+
         if (AvailableBasemaps is null)
         {
             _loadCancellationTokenSource = new CancellationTokenSource();
--- a/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
+++ b/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
@@ -26,13 +26,10 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui;
 /// If connected to a GeoView, changing the basemap selection will change the connected Map or Scene's basemap.
 /// Only basemaps whose spatial reference matches the map or scene's spatial reference can be selected for display.
 /// </remarks>
-#pragma warning disable CA1001 // Types that own disposable fields should be disposable
 public partial class BasemapGallery
-#pragma warning restore CA1001 // Types that own disposable fields should be disposable
 {
     private CollectionView? _listView;
     private readonly BasemapGalleryController _controller;
-    private CancellationTokenSource? _loadCancellationTokenSource;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BasemapGallery"/> class.
@@ -54,13 +51,7 @@ public partial class BasemapGallery
 
         if (AvailableBasemaps is null)
         {
-            _loadCancellationTokenSource = new CancellationTokenSource();
-            try
-            {
-                await _controller.LoadFromDefaultPortal(_loadCancellationTokenSource.Token);
-            }
-            catch (OperationCanceledException)
-            { }
+            await _controller.LoadFromDefaultPortal();
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
```diff
@@ -41,9 +41,17 @@ public BasemapGallery()
...
         GridItemTemplate = DefaultGridDataTemplate;
         ControlTemplate = DefaultControlTemplate;
-        _ = _controller.LoadFromDefaultPortal();
-    }
-
+        Loaded += BasemapGallery_Loaded;
+    }
+
+    private void BasemapGallery_Loaded(object? sender, EventArgs e)
+    {
```
dotMorten, line 48: We need some code here to ensure this only triggers the first time loaded is fired (the null check isn't enough since it could return null or occur before this call finishes).

src/Toolkit/Toolkit.UI.Controls/BasemapGallery/BasemapGallery.Windows.cs
```diff
@@ -45,7 +45,15 @@ public BasemapGallery()
             SizeChanged += BasemapGallerySizeChanged;
             AvailableBasemaps = new ObservableCollection<BasemapGalleryItem>();
             _controller.PropertyChanged += HandleControllerPropertyChanged;
-            _ = _controller.LoadFromDefaultPortal();
+            Loaded += BasemapGallery_Loaded;
+        }
+
+        private void BasemapGallery_Loaded(object? sender, RoutedEventArgs e)
+        {
```
dotMorten, line 52: Same here.

src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
```diff
@@ -41,9 +41,17 @@ public BasemapGallery()
...
+        if (AvailableBasemaps is null)
+        {
+            _ = _controller.LoadFromDefaultPortal();
```
dotMorten, line 51: We might need to consider making this cancellable in case AvailableBasemaps does get set after this code was called.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -246,7 +246,8 @@ internal bool EqualsBasemap(Basemap? other)
                 return false;
             }
 
-            return other == Basemap || other.Name == Basemap?.Name
+            return other == Basemap || (Basemap.BaseLayers.Count == other.BaseLayers.Count
+                                        && Basemap.BaseLayers.All(layer => other?.BaseLayers.FirstOrDefault(l => l.Name == layer.Name) is not null))
                                     || (other.Item?.ItemId != null && other.Item?.ItemId == Basemap?.Item?.ItemId)
                                     || (other.Uri != null && other.Uri == Basemap?.Uri);
```
dotMorten, line 252: We should do early exit on these simple more rigid checks first. If if itemid is the same, just return right away, rather than doing all the layer comparison up-front.
Don't check Layer.Name but instead use the Source URI to compare as well as type comparisons. This entire expression is basically gonna be a lot more complicated than before.
You also forgot to check the ReferenceLayers collection on the basemap.

prathameshnarkhede: I thought we are already doing URI comparisons with `(other.Uri != null && other.Uri == Basemap?.Uri)`. Would it be helpful to check `Layer.Id`?

dotMorten: I mean the URL of the layers. A basemap doesn't have a URI when it's constructed manually like you were doing in your test

src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
```diff
@@ -185,7 +205,11 @@ private static void AvailableBasemapsChanged(BindableObject sender, object oldVa
     public IList<BasemapGalleryItem>? AvailableBasemaps
     {
         get => GetValue(AvailableBasemapsProperty) as IList<BasemapGalleryItem>;
-        set => SetValue(AvailableBasemapsProperty, value);
+        set
+        {
+            SetValue(AvailableBasemapsProperty, value);
+            _loadCancellationTokenSource?.Cancel();
```
dotMorten, line 211: I think it would be cleaner to move all the cancellation logic into the `BasemapControllerGallery`, so it's is shared by all users of it.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -246,9 +246,10 @@ internal bool EqualsBasemap(Basemap? other)
                 return false;
             }
 
-            return other == Basemap || other.Name == Basemap?.Name
-                                    || (other.Item?.ItemId != null && other.Item?.ItemId == Basemap?.Item?.ItemId)
-                                    || (other.Uri != null && other.Uri == Basemap?.Uri);
+            return other == Basemap || (other.Item?.ItemId != null && other.Item?.ItemId == Basemap?.Item?.ItemId)
+                                    || (other.Uri != null && other.Uri == Basemap?.Uri)
+                                    || (Basemap?.BaseLayers.Count == other.BaseLayers.Count
+                                        && Basemap.BaseLayers.All(layer => other?.BaseLayers.FirstOrDefault(l => l.Name == layer.Name) is not null));
```
dotMorten, line 252: Still don't rely on names - rely on the actual content of the BaseLayers and also just as importantly ReferenceLayers. You'll need to compare each layer by type and Source uri.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
prathameshnarkhede: [Layer](https://developers.arcgis.com/net/api-reference/api/netwin/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Mapping.Layer.html) type doesn't seems to have URL property until it is casted to specific Layer Type.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
prathameshnarkhede:
```suggestion
                                        && Basemap.BaseLayers.Zip(other.BaseLayers, (a, b) => new { LayerA = a, LayerB = b })
                                               .All(pair => pair.LayerA.GetType() == pair.LayerB.GetType()
                                                     && (pair.LayerA as ArcGISSceneLayer)?.Source == (pair.LayerB as ArcGISSceneLayer)?.Source
                                                     && (pair.LayerA as ArcGISTiledLayer)?.Source == (pair.LayerB as ArcGISTiledLayer)?.Source
                                                     && (pair.LayerA as ArcGISVectorTiledLayer)?.Source == (pair.LayerB as ArcGISVectorTiledLayer)?.Source)
                                        && Basemap.ReferenceLayers.Zip(other.ReferenceLayers, (a, b) => new { LayerA = a, LayerB = b })
                                               .All(pair => pair.LayerA.GetType() == pair.LayerB.GetType()
                                                     && (pair.LayerA as ArcGISSceneLayer)?.Source == (pair.LayerB as ArcGISSceneLayer)?.Source
                                                     && (pair.LayerA as ArcGISTiledLayer)?.Source == (pair.LayerB as ArcGISTiledLayer)?.Source
                                                     && (pair.LayerA as ArcGISVectorTiledLayer)?.Source == (pair.LayerB as ArcGISVectorTiledLayer)?.Source));
```

How about this?

dotMorten: Reflection isn't an option since that won't work in some release builds.
It might be cleaner to make an layer-comparer method you can use to simplify the overall expression

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -246,9 +246,38 @@ internal bool EqualsBasemap(Basemap? other)
...
+        {
+            // This method can be extended to handle more specific layer comparisons if needed
+            if (layer1.GetType() != layer2.GetType()) return false;
+
+            return layer1 switch
+            {
+                ArcGISSceneLayer sceneLayer => sceneLayer.Source == ((ArcGISSceneLayer)layer2).Source,
+                ArcGISTiledLayer tiledLayer => tiledLayer.Source == ((ArcGISTiledLayer)layer2).Source,
+                ArcGISVectorTiledLayer vectorTiledLayer => vectorTiledLayer.Source == ((ArcGISVectorTiledLayer)layer2).Source,
+                _ => false,
```
dotMorten, line 279: This is great. There's a few missing though. 
Here's a list of layers you can expect to see in a basemap: https://developers.arcgis.com/web-map-specification/objects/baseMapLayer/ (note that these are the webmapspec names, and the maps sdk names might be slightly different)

dotMorten (commented):
Really close! Just need those missing layer types.

dotMorten (approved):

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -246,9 +246,54 @@ internal bool EqualsBasemap(Basemap? other)
...
+            return LayersEqual(basemap1.BaseLayers, basemap2.BaseLayers)
+                && LayersEqual(basemap1.ReferenceLayers, basemap2.ReferenceLayers);
+        }
+
+        private static bool LayersEqual(LayerCollection layers1, LayerCollection layers2)
+        {
+            return layers1.Count == layers2.Count
+                && layers1.Zip(layers2, (layer1, layer2) => new { Layer1 = layer1, Layer2 = layer2 })
+                    .All(pair => LayerEquals(pair.Layer1, pair.Layer2));
+        }
```
mstefarov, line 267: Optional:
You can pass LayersEquals directly to Zip, then just check that all returned bools are true.

```cs
bool LayersEqual(LayerCollection layers1, LayerCollection layers2)
{
    return layers1.Count == layers2.Count 
        && layers1.Zip(layers2, LayerEquals).All(equal => equal);
}
```

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -246,9 +246,54 @@ internal bool EqualsBasemap(Basemap? other)
...
+
+        private static bool LayerEquals(Layer layer1, Layer layer2)
+        {
+            // This method can be extended to handle more specific layer comparisons if needed
+            if (layer1.GetType() != layer2.GetType()) return false;
+
+            return layer1 switch
```
mstefarov, line 274: How about DynamicEntityLayer, FeatureLayer, and RasterLayer?

mstefarov (commented):
@dotMorten All the comparison logic in this PR is making a good case for us to expose RT_Layer_getHash and RT_Layer_equals in the public API 😉

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
prathameshnarkhede:
```suggestion
        private static bool LayersEqual(LayerCollection layers1, LayerCollection layers2)
        {
            return layers1.Count == layers2.Count
                && layers1.Zip(layers2, LayerEquals).All(equal => equal);
        }
```

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
prathameshnarkhede: What could be the properties to compare in `DynamicEntityLayer` and `RasterLayer`?

dotMorten: @mstefarov those aren't valid basemap layers according to [the official webmap spec](https://developers.arcgis.com/web-map-specification/objects/baseMapLayer/).

mstefarov (approved):

Issue #576, opened 2024-06-06
PopupViewer: Allow overriding URL clicks


dotMorten:
Currently if you click a link in the popup viewer it'll open the default browser. In some cases you might want to just open the link in a webview inside the app.
We should provide a [similar approach as with attachment clicks](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/ae34e71ba4bf81ff6c57d178c26cf53dd89bf8a9/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.cs#L239-L273), allowing users to cancel the default operation and perform their own.

Follow-up to issue #555 suggested by @bsundsbo

prathameshnarkhede:
This feature will be available in v200.5

PR #584, opened 2024-07-16
Added support to override URL clicks in PopupViewer

prathameshnarkhede:
## Related Issue #576 

Added `HyperlinkClicked` Event handler on PopupViewer Element to allow URL handling customization.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
@@ -59,7 +59,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
 
         internal static PopupViewer? GetPopupViewerParent(DependencyObject? child) => GetParent<PopupViewer>(child);
 
-        private static T? GetParent<T>(DependencyObject? child) where T : FrameworkElement
+        private static T? GetParent<T>(DependencyObject? child) where T : DependencyObject
         {
             // Traverse up the visual tree to find the PopupViewer parent
             while (child is not null)
--- a/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
@@ -57,7 +57,9 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
 #pragma warning restore CS0618 // Type or member is obsolete
         }
 
-        internal static PopupViewer? GetPopupViewerParent(DependencyObject? child)
+        internal static PopupViewer? GetPopupViewerParent(DependencyObject? child) => GetParent<PopupViewer>(child);
+
+        private static T? GetParent<T>(DependencyObject? child) where T : FrameworkElement
         {
             // Traverse up the visual tree to find the PopupViewer parent
             while (child is not null)
@@ -72,12 +74,12 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                     child = VisualTreeHelper.GetParent(child);
                 }
 
-                if (child is PopupViewer popupViewer)
+                if (child is T popupViewer)
                 {
                     return popupViewer;
                 }
             }
-            return null; // No PopupViewer parent found
+            return default; // No parent found
         }
 
     }
--- a/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
@@ -69,9 +69,9 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                     // Use the logical parent for FrameworkContentElement instances
                     child = contentElement.Parent ?? VisualTreeHelper.GetParent(contentElement);
                 }
-                else
+                else if (child is Visual visual)
                 {
-                    child = VisualTreeHelper.GetParent(child);
+                    child = VisualTreeHelper.GetParent(visual);
                 }
 
                 if (child is T popupViewer)
@@ -81,7 +81,6 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             }
             return default; // No parent found
         }
-
     }
 }
 #endif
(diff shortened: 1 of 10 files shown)


src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.cs
```diff
@@ -70,6 +74,41 @@ public override void OnApplyTemplate()
...
+                        return;
+                }
+
+                try
+                {
+#if WPF
+                    await Launcher.LaunchUriAsync(uri);
+#elif MAUI
+                    await Browser.OpenAsync(uri, BrowserLaunchMode.SystemPreferred);
+#endif
```
dotMorten, line 103: This needs to use `Esri.ArcGISRuntime.Toolkit.Internal.Launcher.LaunchUri` which already has the cross-platform code to deal with launching. Also Line 100 doesn't work on .NET Framework.

src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.cs
```diff
@@ -70,6 +74,41 @@ public override void OnApplyTemplate()
...
+                }
+                catch (Exception ex)
+                {
+                    System.Diagnostics.Trace.WriteLine($"Failed to open URL: " + ex.Message);
+                }
+            }
+
```
dotMorten, line 110: Extra linefeed

src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.Maui.cs
```diff
@@ -455,6 +451,16 @@ private static bool MapsToBlock(MarkupNode node)
...
+
+        private PopupViewer? GetPopupViewerParent()
+        {
+            var parent = this.Parent;
+            while (parent is not null && parent is not PopupViewer popup)
+            {
+                parent = parent.Parent;
+            }
+            return parent as PopupViewer;
+        }
```
dotMorten, line 463: Move to Popupviewer and make static. See https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/ac1bc3e6c5a6239f60e9f9dc42f986f3495ade3b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.Maui.cs#L161 for example

dotMorten (changes requested):

src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
```diff
@@ -56,6 +56,32 @@ private static void OnPopupManagerPropertyChanged(DependencyObject d, Dependency
...
+        private static T? GetParent<T>(DependencyObject? child) where T : DependencyObject
+        {
+            // Traverse up the visual tree to find the PopupViewer parent
+            while (child is not null)
+            {
+                if (child is FrameworkContentElement contentElement)
+                {
+                    // Use the logical parent for FrameworkContentElement instances
+                    child = contentElement.Parent ?? VisualTreeHelper.GetParent(contentElement);
+                }
```
dotMorten, line 71: this rewrite is quite confusing to me - there's now two `VisualTreeHelper.GetParent` calls in this method. I think it was clearer before how we had it. The fallback would never work because using the VisualTreeHelper.GetParent on a FrameworkContentElement was clearly throwing when we first tested it.
Perhaps all you really need to do is remove the fallback.

src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.cs
```diff
@@ -19,9 +19,13 @@
 using Esri.ArcGISRuntime.Toolkit.Internal;
 using Esri.ArcGISRuntime.UI;
 #if WPF
+using Esri.ArcGISRuntime.Toolkit.UI.Controls;
 using System.Windows.Documents;
 using System.Windows.Input;
 using System.Windows.Navigation;
+#elif MAUI
+using Esri.ArcGISRuntime.Toolkit.Maui;
+using Microsoft.Maui.ApplicationModel;
```
dotMorten, line 28: Undo

src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
```diff
@@ -56,6 +56,32 @@ private static void OnPopupManagerPropertyChanged(DependencyObject d, Dependency
...
+                else
```
dotMorten, line 72: Should this then check that `child` is a `Visual` ?

prathameshnarkhede: That would be clearer I think.

src/Toolkit/Toolkit/UI/Controls/PopupViewer/PopupViewer.Windows.cs
prathameshnarkhede:
```
internal static PopupViewer? GetPopupViewerParent(DependencyObject? child)
{
    if (child is FrameworkContentElement elm)
    {
        child = GetVisualParent(elm);
    }
    if (child is null)
        return null;
    var parent = VisualTreeHelper.GetParent(child);
    while (parent is not null and not PopupViewer)
    {
        parent = VisualTreeHelper.GetParent(parent);
    }
    return parent as PopupViewer;
}
internal static Visual? GetVisualParent(FrameworkContentElement child)
{
    var parent = child.Parent;
    while (parent is FrameworkContentElement elm)
    {
        parent = elm.Parent;
    }
    return parent as Visual;
}

```

This was the original implementation, Its just refactor and had two `VisualTreeHelper.GetParent` calls.

dotMorten: It's not the same. If `VisualTreeHelper.GetParent(contentElement)` is ever hit it'll throw an exception.

prathameshnarkhede: Got it. Reverting it back.

mstefarov (approved):

dotMorten (approved):

PR #586, opened 2024-07-25
BugFix: BasemapGallery fails to load default Basemaps in WPF

prathameshnarkhede:
### Description:
The BasemapGallery used to be fully empty when default Basemaps are loaded.
It also preserved loading custom Basemaps into the BasemapGallery.

#### BasemapGallery with Default Basemaps:

image

#### BasemapGallery with Custom Basemaps:

image


mstefarov (approved):

dotMorten (changes requested):
Breaking change. The AvailableBasemaps collection has always been initialized. It means existing code like this is now throwing where it was working before:
`new BasemapGallery().AvailableBasemaps.Add(...)`

dotMorten (approved):

PR #593, opened 2024-08-16
Switch cancelButton image to FontImageSource allowing for more flexible and scalable icon usage

prathameshnarkhede:
This also fixes problem of visibility with Dark/Light mode of the Cancel button.


src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
```diff
@@ -71,7 +71,7 @@ public SearchView()
 
         if (GetTemplateChild(nameof(PART_CancelButton)) is ImageButton cancelButton)
         {
-            cancelButton.Source = ImageSource.FromResource($"Esri.ArcGISRuntime.Toolkit.Maui.Assets.x{suffix}.png", Assembly.GetAssembly(typeof(SearchView)));
```
dotMorten, line 74: Do we need to delete the image? Is it used elsewhere?

dotMorten (approved):

src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
prathameshnarkhede: Yea. You're right. I just removed files from resources now.

mstefarov (approved):

PR #598, opened 2024-08-28
Breaking change: Changed from ListView to CollectionView in BookmarksView which resolves ItemTemplate issue in iOS

prathameshnarkhede:
The bug is related to Bookmarks updating template does not apply on all rows for iOS.

MAUI Bug Reported [dotnet/maui#24492](https://github.com/dotnet/maui/issues/24492)

EDIT:
Found that changing to `CollectionView` actually solves this issue and I was able to test it on Windows, iOS and Android platforms.
There is going to be a breaking change for `ItemTemplates` Cells used for `CollectionView`.

### DIFF
--- a/src/Samples/Toolkit.SampleApp.Maui/Samples/BookmarksViewSample.xaml
+++ b/src/Samples/Toolkit.SampleApp.Maui/Samples/BookmarksViewSample.xaml
@@ -10,12 +10,10 @@
              x:Class="Toolkit.SampleApp.Maui.Samples.BookmarksViewSample">
     <ContentPage.Resources>
         <DataTemplate x:Key="ItemTemplateOne">
-            <TextCell Text="{Binding Name}" TextColor="Red" />
+            <Label Text="{Binding Name}" TextColor="Red" />
         </DataTemplate>
         <DataTemplate x:Key="ItemTemplateTwo">
-            <ViewCell>
-                <Label BackgroundColor="Red" TextColor="White" Text="{Binding Name}" />
-            </ViewCell>
+            <Label BackgroundColor="Red" TextColor="White" Text="{Binding Name}" />
         </DataTemplate>
     </ContentPage.Resources>
     <ContentPage.Content>
--- a/src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
+++ b/src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
@@ -24,7 +24,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui;
 /// </summary>
 public class BookmarksView : TemplatedView
 {
-    private BookmarksViewDataSource _dataSource = new BookmarksViewDataSource();
+    private readonly BookmarksViewDataSource _dataSource = new();
     private const string _presentingViewName = "PresentingView";
 
     private static readonly DataTemplate DefaultDataTemplate;
@@ -33,18 +33,17 @@ public class BookmarksView : TemplatedView
     [DynamicDependency(nameof(Bookmark.Name), "Esri.ArcGISRuntime.Mapping.Bookmark", "Esri.ArcGISRuntime")]
     static BookmarksView()
     {
-        DefaultDataTemplate = new DataTemplate(() =>
-        {
-            var defaultLabel = new Label();
-            defaultLabel.SetBinding(Label.TextProperty, nameof(Bookmark.Name));
-            return defaultLabel;
-        });
+        string dataTemplate =
+            @"<DataTemplate xmlns=""http://schemas.microsoft.com/dotnet/2021/maui"">
+                <Label Text=""{Binding Name}"" />
+              </DataTemplate>";
+        DefaultDataTemplate = Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(new DataTemplate(), dataTemplate);
 
         string template =
             $@"<ControlTemplate xmlns=""http://schemas.microsoft.com/dotnet/2021/maui""
-                               xmlns:x=""http://schemas.microsoft.com/winfx/2009/xaml""
-                               xmlns:esriTK=""clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui"">
-                <CollectionView x:Name=""{_presentingViewName}"" SelectionMode=""Single"" />
+                                xmlns:x=""http://schemas.microsoft.com/winfx/2009/xaml""
+                                xmlns:esriTK=""clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui"">
+                <CollectionView x:Name=""{_presentingViewName}"" SelectionMode=""Single"" ItemTemplate=""{{TemplateBinding ItemTemplate}}"" />
              </ControlTemplate>";
         DefaultControlTemplate = Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(new ControlTemplate(), template);
     }
@@ -70,8 +69,8 @@ public class BookmarksView : TemplatedView
         {
             view.SelectionChanged -= Internal_bookmarkSelected;
             view.SelectionChanged += Internal_bookmarkSelected;
+            view.ItemsSource = _dataSource;
         }
-        UpdatePresentingView();
     }
 
     /// <summary>
@@ -123,7 +122,7 @@ public class BookmarksView : TemplatedView
     /// Identifies the <see cref="ItemTemplate" /> bindable property.
     /// </summary>
     public static readonly BindableProperty ItemTemplateProperty =
-        BindableProperty.Create(nameof(ItemTemplate), typeof(DataTemplate), typeof(BookmarksView), DefaultDataTemplate, BindingMode.OneWay, null, propertyChanged: ItemTemplateChanged);
+        BindableProperty.Create(nameof(ItemTemplate), typeof(DataTemplate), typeof(BookmarksView), DefaultDataTemplate, BindingMode.OneWay, null);
 
     /// <summary>
     /// Handles property changes for the <see cref="BookmarksOverride" /> bindable property.
@@ -141,15 +140,6 @@ public class BookmarksView : TemplatedView
         BookmarksView bookmarkView = (BookmarksView)sender;
 
         bookmarkView._dataSource.SetGeoView(newValue as GeoView);
-    }
-
-    /// <summary>
-    /// Handles property changes for the <see cref="ItemTemplate" /> bindable property.
-    /// </summary>
-    private static void ItemTemplateChanged(BindableObject sender, object? oldValue, object? newValue)
-    {
-        BookmarksView bookmarksView = (BookmarksView)sender;
-        bookmarksView.UpdatePresentingView();
     }
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


mstefarov (approved):

mstefarov (approved):

src/Samples/Toolkit.SampleApp.Maui/Samples/BookmarksViewSample.xaml
```diff
@@ -10,12 +10,10 @@
...
     <ContentPage.Resources>
         <DataTemplate x:Key="ItemTemplateOne">
-            <TextCell Text="{Binding Name}" TextColor="Red" />
+            <Label Text="{Binding Name}" TextColor="Red" />
         </DataTemplate>
         <DataTemplate x:Key="ItemTemplateTwo">
-            <ViewCell>
-                <Label BackgroundColor="Red" TextColor="White" Text="{Binding Name}" />
-            </ViewCell>
+            <Label BackgroundColor="Red" TextColor="White" Text="{Binding Name}" />
```
dotMorten, line 16: Will cells still work or is this a breaking behavior?

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -66,20 +69,7 @@ public BookmarksView()
...
-        }
-
-        _presentingView = GetTemplateChild("PresentingView") as ListView;
-
-        if (_presentingView != null)
-        {
-            _presentingView.ItemSelected += Internal_bookmarkSelected;
-            _presentingView.ItemTemplate = ItemTemplate;
-            _presentingView.ItemsSource = _dataSource;
-        }
```
dotMorten, line 82: Hooking up and unhooking from events on view elements needs to stay in OnApplyTemplate.
The change also means if a template gets changed, the old listener is still hanging around and

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -192,16 +170,36 @@ private void SelectAndNavigateToBookmark(Bookmark bookmark)
...
+
+    private void UpdateListView()
     {
-        if (e.SelectedItem is Bookmark bm)
+        var collection = (CollectionView)GetTemplateChild(_presentingViewName);
+        if (collection != null)
         {
-            SelectAndNavigateToBookmark(bm);
+            collection.SelectionChanged -= Internal_bookmarkSelected;
+            collection.SelectionChanged += Internal_bookmarkSelected;
```
dotMorten, line 191: This shouldn't be necessary - we should know when we need to subscribe and unsubscribe.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -192,16 +170,36 @@ private void SelectAndNavigateToBookmark(Bookmark bookmark)
...
+        }
+    }
+
+    private void UpdateListView()
     {
-        if (e.SelectedItem is Bookmark bm)
+        var collection = (CollectionView)GetTemplateChild(_presentingViewName);
+        if (collection != null)
         {
-            SelectAndNavigateToBookmark(bm);
```
dotMorten, line 199: What happened to this line? What triggers the navigation event now?

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -192,16 +170,36 @@ private void SelectAndNavigateToBookmark(Bookmark bookmark)
...
 
-        if (e.SelectedItem != null && sender is ListView lv)
+    private void OnBookmarksCollectionChanged(object? sender, NotifyCollectionChangedEventArgs e)
+    {
+        if (e.Action is NotifyCollectionChangedAction.Add)
         {
-            lv.SelectedItem = null;
+            UpdateListView();
         }
     }
```
dotMorten, line 204: This should be in `#if Android` since that's the only thing using it. However I think cleaning up the event handling will aviod this workaround

dotMorten (changes requested):

src/Samples/Toolkit.SampleApp.Maui/Samples/BookmarksViewSample.xaml
prathameshnarkhede: It think [CollectionView won't support Cell](https://learn.microsoft.com/en-us/dotnet/maui/user-interface/controls/collectionview/populate-data?view=net-maui-8.0#:~:text=a%20DataTemplateSelector.-,Important,an%20exception%20being%20thrown%20because%20CollectionView%20has%20no%20concept%20of%20cells.,-Context%20menus) if we set root of the Template to ViewCell, but we can wrap it inside another element for purpose when we apply template.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
prathameshnarkhede: There is a call to `UpdatePresentingView` in `ItemTemplateChanged` handler [here](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/598/files#diff-b515b77dae873f62a7192e60f5ac03fd034dd3bb71c44393ca4cef9adc673752R150).

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
prathameshnarkhede: Navigation event is triggered by [same code](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/598/files#diff-b515b77dae873f62a7192e60f5ac03fd034dd3bb71c44393ca4cef9adc673752R179). It's just Git changes are skewed a little.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
prathameshnarkhede: I have corrected this code change.

src/Samples/Toolkit.SampleApp.Maui/Samples/BookmarksViewSample.xaml
prathameshnarkhede: I tried looking at the options for cells if we can put it in any container. But looks like there is only TableView which takes Cells. But it takes whole lot of space in a row. So it might be breaking change for code using cells.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -67,19 +65,13 @@ protected override void OnApplyTemplate()
...
-        _presentingView = GetTemplateChild("PresentingView") as ListView;
-
-        if (_presentingView != null)
+        var collectionView = (CollectionView)GetTemplateChild(_presentingViewName);
+        if (collectionView is CollectionView view)
         {
-            _presentingView.ItemSelected += Internal_bookmarkSelected;
-            _presentingView.ItemTemplate = ItemTemplate;
-            _presentingView.ItemsSource = _dataSource;
+            view.SelectionChanged -= Internal_bookmarkSelected;
```
dotMorten, line 71: This should be moved up to the code you deleted in line 70-73

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -35,18 +35,17 @@ static BookmarksView()
...
-                                        <x:Arguments>
-                                            <ListViewCachingStrategy>RecycleElement</ListViewCachingStrategy>
-                                        </x:Arguments>
-                                    </ListView>
-                                </ControlTemplate>";
+        string template =
+            $@"<ControlTemplate xmlns=""http://schemas.microsoft.com/dotnet/2021/maui""
+                               xmlns:x=""http://schemas.microsoft.com/winfx/2009/xaml""
+                               xmlns:esriTK=""clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui"">
+                <CollectionView x:Name=""{_presentingViewName}"" SelectionMode=""Single"" />
```
dotMorten, line 47: You should add `ItemTemplate="{TemplateBinding ItemTemplate}"` here, instead of having to deal with it in the code.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -192,16 +172,25 @@ private void SelectAndNavigateToBookmark(Bookmark bookmark)
...
 
-        if (e.SelectedItem != null && sender is ListView lv)
+    private void UpdatePresentingView()
+    {
+        var collectionView = (CollectionView)GetTemplateChild(_presentingViewName);
+        if (collectionView is CollectionView view)
         {
-            lv.SelectedItem = null;
+            view.ItemTemplate = ItemTemplate;
+            view.ItemsSource = _dataSource;
```
dotMorten, line 193: The datasource never changes so I would move this to OnApplyTemplate.

dotMorten (changes requested):
With the suggested changes, you don't need the method `UpdatePresentingView()` at all.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -24,29 +24,28 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui;
 /// </summary>
 public class BookmarksView : TemplatedView
 {
-    private ListView? _presentingView;
-    private BookmarksViewDataSource _dataSource = new BookmarksViewDataSource();
+    private readonly BookmarksViewDataSource _dataSource = new();
+    private const string _presentingViewName = "PresentingView";
+    CollectionView? _presentingView;
```
dotMorten, line 29: needs `private` prefix

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -67,17 +65,15 @@ protected override void OnApplyTemplate()
...
         base.OnApplyTemplate();
 
-        if (_presentingView != null)
+        if (_presentingView is not null)
         {
-            _presentingView.ItemSelected -= Internal_bookmarkSelected;
+            _presentingView.SelectionChanged -= Internal_bookmarkSelected;
         }
 
-        _presentingView = GetTemplateChild("PresentingView") as ListView;
```
dotMorten, line 75: Keep this line but update it for CollectionView. That ensures an old _presentingView value would get overwritten with null if it isn't found.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -192,16 +167,15 @@ private void SelectAndNavigateToBookmark(Bookmark bookmark)
...
-
-        if (e.SelectedItem != null && sender is ListView lv)
+        if (sender is CollectionView cv)
         {
-            lv.SelectedItem = null;
+            if (cv.SelectedItem is Bookmark item)
+            {
+                SelectAndNavigateToBookmark(item);
+            }
+            cv.ClearValue(CollectionView.SelectedItemProperty);
```
dotMorten, line 178: I don't think you need to do this. Instead in the template set `SelectionMode='None'`. I believe you can still click it.

dotMorten (changes requested):
Almost there

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
prathameshnarkhede: By default, Selection mode of `CollectionView` is `None` which actually disables the `SelectionChanged` event.
So we can not get to hit this handler.

dotMorten (approved):

PR #601, opened 2024-09-03
FeatureForms: Adds a temporary combox value when no coded value matches

dotMorten:
This change allows the comboboxes to show values that are outside the coded value domains. A warning will still be shown, and once selecting a valid value in the dropdown, the temporary placeholder entry in the dropdown will be removed. This matches the behavior of ArcGIS Online webviewer


src/Toolkit/Toolkit/UI/Controls/FeatureForm/ComboboxFormInputView.cs
```diff
@@ -123,6 +132,10 @@ private class ComboBoxNullValue
             public string? Name { get; set; }
             public override string ToString() => Name!;
         }
+
+        private class ComboBoxPlaceHolderValue : ComboBoxNullValue
+        {
+		}
```
prathameshnarkhede, line 138: Spaces seems to be skewed here.

mstefarov (approved):

PR #603, opened 2024-09-18
Fixing SymbolDisplay appearing like dots on certain Android devices

prathameshnarkhede:


src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
```diff
@@ -161,8 +161,12 @@ private async Task UpdateSwatchAsync()
                 image.MaximumWidthRequest = imageData.Width / scale;
                 image.MaximumHeightRequest = imageData.Height / scale;
                 image.Source = await imageData.ToImageSourceAsync();
-                SourceUpdated?.Invoke(this, EventArgs.Empty);
+                SourceUpdated?.Invoke(this, EventArgs.Empty);
 #pragma warning restore ESRI1800
+#if ANDROID
+                this.MinimumWidthRequest = imageData.Width / scale;
+                this.MinimumHeightRequest = imageData.Height / scale;
+#endif
```
dotMorten, line 169: This would override any setting you might have on the control. You might want to limit the size of the symbol, and this would prevent that.
The issue is either in the MAUI layout classes (which can be worked around in the sample), or perhaps how this control measures itself and reports its ideal size, while allowing itself to size smaller if there's no room for it.

dotMorten (approved):

src/Toolkit/Toolkit.Maui/SymbolDisplay/SymbolDisplay.cs
prathameshnarkhede: Reverted this and removed Border.

mstefarov (approved):

PR #611, opened 2024-11-14
Font fixes

dotMorten:
Fixed font references so they work even if font isn't installed


jnery (approved):
LGTM (I was wondering how it affected UWP, makes sense now)

PR #615, opened 2024-12-11
Add built-in barcode scanning

dotMorten:
Adds built-in barcode scanning 
 - [x] Windows
 - [x] iOS
   - Swift barcode scanner: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/81552fcff71e638a2c306ead6406a8a0c93ef334/Sources/ArcGISToolkit/Common/CodeScanner.swift
 - [ ] Android 
    - Kotlin barcode scanner: https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/30b0339327181595db2c889eb33c7ce8f7914a8d/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/barcode/BarcodeScanner.kt
    - Android doc: https://developers.google.com/ml-kit/vision/barcode-scanning/android

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -45,8 +45,8 @@
     <NETMauiTarget Condition="'$(NETMauiTarget)'==''">net8.0</NETMauiTarget>
     <NetMauiWindowsTargetFramework Condition="'$(NetMauiWindowsTargetFramework)'==''">$(NETMauiTarget)-windows10.0.$(WindowsSDKTargetBuild).0</NetMauiWindowsTargetFramework>
     <NetAndroidTargetFramework Condition="'$(NetAndroidTargetFramework)'==''">$(NETMauiTarget)-android34.0</NetAndroidTargetFramework>
-    <NetCatalystTargetFramework Condition="'$(NetCatalystTargetFramework)'==''">$(NETMauiTarget)-maccatalyst17.0</NetCatalystTargetFramework>
-    <NetiOSTargetFramework Condition="'$(NetiOSTargetFramework)'==''">$(NETMauiTarget)-ios17.0</NetiOSTargetFramework>  
+    <NetCatalystTargetFramework Condition="'$(NetCatalystTargetFramework)'==''">$(NETMauiTarget)-maccatalyst18.0</NetCatalystTargetFramework>
+    <NetiOSTargetFramework Condition="'$(NetiOSTargetFramework)'==''">$(NETMauiTarget)-ios18.0</NetiOSTargetFramework>  
     <MauiVersion Condition="'$(MauiVersion)'==''">8.0.3</MauiVersion>
     <WinAppSDKVersion Condition="'$(WinAppSDKVersion)'==''">1.4.230822000</WinAppSDKVersion>
   </PropertyGroup>
--- a/src/Samples/Toolkit.SampleApp.Maui/Samples/FeatureFormViewSample.xaml.cs
+++ b/src/Samples/Toolkit.SampleApp.Maui/Samples/FeatureFormViewSample.xaml.cs
@@ -57,7 +57,7 @@ namespace Toolkit.SampleApp.Maui.Samples
 
         private async void BarcodeButtonClicked(object sender, Esri.ArcGISRuntime.Toolkit.Maui.BarcodeButtonClickedEventArgs e)
         {
-#if !__IOS__ // iOS has this built-in
+#if !__IOS__ && !WINDOWS // iOS and Windows has this built-in
             // If user clicks the barcode button in a barcode input element, use ZXing library (https://github.com/Redth/ZXing.Net.Maui)
             // to scan a barcode using the device camera
             e.Handled = true;
--- a/src/Samples/Toolkit.SampleApp.Maui/Samples/FeatureFormViewSample.xaml.cs
+++ b/src/Samples/Toolkit.SampleApp.Maui/Samples/FeatureFormViewSample.xaml.cs
@@ -57,8 +57,10 @@ namespace Toolkit.SampleApp.Maui.Samples
 
         private async void BarcodeButtonClicked(object sender, Esri.ArcGISRuntime.Toolkit.Maui.BarcodeButtonClickedEventArgs e)
         {
+#if !__IOS__ // iOS has this built-in
             // If user clicks the barcode button in a barcode input element, use ZXing library (https://github.com/Redth/ZXing.Net.Maui)
             // to scan a barcode using the device camera
+            e.Handled = true;
             ZXing.Net.Maui.Controls.CameraBarcodeReaderView view = new ZXing.Net.Maui.Controls.CameraBarcodeReaderView();
             TaskCompletionSource<string?> tcs = new TaskCompletionSource<string?>();
             view.BarcodesDetected += (s, e) =>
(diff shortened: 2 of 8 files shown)


dotMorten:
Android support will be a separate PR to keep the reviews easier

src/Directory.Build.props
```diff
@@ -45,8 +45,8 @@
     <NETMauiTarget Condition="'$(NETMauiTarget)'==''">net8.0</NETMauiTarget>
     <NetMauiWindowsTargetFramework Condition="'$(NetMauiWindowsTargetFramework)'==''">$(NETMauiTarget)-windows10.0.$(WindowsSDKTargetBuild).0</NetMauiWindowsTargetFramework>
     <NetAndroidTargetFramework Condition="'$(NetAndroidTargetFramework)'==''">$(NETMauiTarget)-android34.0</NetAndroidTargetFramework>
-    <NetCatalystTargetFramework Condition="'$(NetCatalystTargetFramework)'==''">$(NETMauiTarget)-maccatalyst17.0</NetCatalystTargetFramework>
-    <NetiOSTargetFramework Condition="'$(NetiOSTargetFramework)'==''">$(NETMauiTarget)-ios17.0</NetiOSTargetFramework>  
+    <NetCatalystTargetFramework Condition="'$(NetCatalystTargetFramework)'==''">$(NETMauiTarget)-maccatalyst18.0</NetCatalystTargetFramework>
+    <NetiOSTargetFramework Condition="'$(NetiOSTargetFramework)'==''">$(NETMauiTarget)-ios18.0</NetiOSTargetFramework>
```
mstefarov, line 49: Should this PR go into v.next branch instead?  It's raising our dependencies beyond what was required for 200.6.0

src/Toolkit/Toolkit.Maui/FeatureForm/ScannerViewController.iOS.cs
```diff
@@ -0,0 +1,347 @@
...
+            base.ViewDidLayoutSubviews();
+            UpdateReticleAndAutoFocus();
+        }
+
+        public override async void ViewDidLoad()
+        {
+            base.ViewDidLoad();
+            var success = await AVCaptureDevice.RequestAccessForMediaTypeAsync(AVAuthorizationMediaType.Video);
+            if (success != true)
+                return;
```
mstefarov, line 68: It looks like our error-detection code just bails out and leaves the user at a blank screen.  Would you want to improve on that, e.g. telling user to grant permission or printing an error message if videoInput creation fails?

dotMorten: The user will be presented with a dialog to access the camera. If they deny it, it basically just cancels out and doesn't do anything. It should be obvious why that button doesn't do anything when you already hit cancel, rather than bugging them with an additional prompt to close.

src/Directory.Build.props
dotMorten: yes we can change that

dotMorten: it's actually not a dependency change... if you compile for net8.0, it defaults to 18.0 now. 
image

src/Samples/Toolkit.SampleApp.Maui/Samples/FeatureFormViewSample.xaml.cs
```diff
@@ -57,8 +57,10 @@ private async void mapView_GeoViewTapped(object sender, Esri.ArcGISRuntime.Maui.
 
         private async void BarcodeButtonClicked(object sender, Esri.ArcGISRuntime.Toolkit.Maui.BarcodeButtonClickedEventArgs e)
         {
+#if !__IOS__ && !WINDOWS // iOS and Windows has this built-in
             // If user clicks the barcode button in a barcode input element, use ZXing library (https://github.com/Redth/ZXing.Net.Maui)
             // to scan a barcode using the device camera
+            e.Handled = true;
             ZXing.Net.Maui.Controls.CameraBarcodeReaderView view = new ZXing.Net.Maui.Controls.CameraBarcodeReaderView();
```
prathameshnarkhede, line 64: Are we using ZXing Library here as a fallback?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/BarcodeScanner.Windows.cs
```diff
@@ -0,0 +1,88 @@
...
+                return await TaskCompletionSource.Task;
+            }
+            catch(System.Exception ex)
+            {
+                TaskCompletionSource?.TrySetException(ex);
+                throw;
+            }
+            finally
+            {
+                await claimedScanner.DisableAsync();
```
prathameshnarkhede, line 61: Do we need a null check here in case exception is thrown before its assignment?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/TextFormInputView.cs
```diff
@@ -374,5 +374,34 @@ private void UpdateValidationState()
 #endif
             }
         }
+        
+
+        private
+#if __IOS__ || WINDOWS && NETCOREAPP || NETFX_CORE
+            async
+#endif
+            void LaunchBarcodeScanner(FieldFormElement element)
```
prathameshnarkhede, line 383: Should we add logic to check if Barcode Scanner is available before we start scanning?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/TextFormInputView.cs
```diff
@@ -374,5 +374,34 @@ private void UpdateValidationState()
...
+                // TODO
+#endif
+                if (!string.IsNullOrWhiteSpace(barcode))
+                {
+                    element.UpdateValue(barcode);
+                }
+            }
+            catch
+            {
+
```
prathameshnarkhede, line 403: What could be the exceptions thrown by the current block? Should we handle those exceptions and return a message for the user?

src/Samples/Toolkit.SampleApp.Maui/Samples/FeatureFormViewSample.xaml.cs
dotMorten: Yes. This is existing code, and disabling for windows and ios since it now has it built in. Once android changes go in, this'll be completely removed.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/TextFormInputView.cs
dotMorten: Cancellation is the most typical one. The other is denying access. Both of which shouldn't update the element. I think ignore is sufficient here. We do need the try/catch for the async void

src/Toolkit/Toolkit/UI/Controls/FeatureForm/TextFormInputView.cs
dotMorten: If it's not available, this is a no-op, like it is today.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/BarcodeScanner.Windows.cs
dotMorten: Fixed

mstefarov (approved):

PR #620, opened 2025-01-14
Rewrites several parts of the SearchView to make it AoT compatible in WinUI

dotMorten:
Main change is to move away from raising INPC in SearchView and instead rely on a TemplateSettings class to act as a view model. This is following the common WinUI/UWP pattern described here: https://learn.microsoft.com/en-us/windows/uwp/xaml-platform/template-settings-classes. Since so much code is shared with WPF, I made the same changes there, so that the code doesn't get too platform-convoluted. It also does clean up the control class nicely. Basically a bunch of readonly properties that raises INPC has been moved to the `SearchViewTemplateSettings` class.

Other things of note:
- Mark classes with `[WinRT.GeneratedBindableCustomProperty]` and make them partial so code generation will occur for types being bound in XAML.
- Use dependency properties when possible
- `IList<T>` is problematic for code generation since you can't generate binding code for it, so introduced some private collection classes that inherit from List<T> and adds the `GeneratedBindableCustomProperty` attribute. Replaced the setter with a Set method instead to ensure the correct type would be used.

Internal ref: 13091

### DIFF
--- a/src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
@@ -35,7 +35,10 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
 #endif
 #pragma warning disable IDE0079
 #pragma warning disable CA1001
-    public partial class SearchView : Control, INotifyPropertyChanged
+#if WINUI
+    [WinRT.GeneratedBindableCustomProperty]
+#endif
+    public partial class SearchView : Control
 #pragma warning restore CA1001
 #pragma warning restore IDE0079
     {
@@ -43,7 +46,6 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         private const int TypingDelayMilliseconds = 75;
         private GeoModel? _lastUsedGeomodel;
         private readonly GraphicsOverlay _resultOverlay;
-        private bool _isSourceSelectOpen;
         private CancellationTokenSource? _configurationCancellationToken;
 
         // Flag indicates whether control is waiting after user finished typing.
@@ -61,9 +63,14 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             DataContext = this;
             SearchViewModel = new SearchViewModel();
             _resultOverlay = new GraphicsOverlay { Id = "SearchView_Result_Overlay" };
-            ClearCommand = new DelegateCommand(HandleClearSearchCommand);
-            SearchCommand = new DelegateCommand(HandleSearchCommand);
-            RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
+#if WINDOWS_XAML
+            SetValue(TemplateSettingsProperty, new SearchViewTemplateSettings(this));
+#elif WPF
+            SetValue(TemplateSettingsPropertyKey, new SearchViewTemplateSettings(this));
+#endif
+            TemplateSettings.ClearCommand = new DelegateCommand(HandleClearSearchCommand);
+            TemplateSettings.SearchCommand = new DelegateCommand(HandleSearchCommand);
+            TemplateSettings.RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
             InitializeLocalizedStrings();
         }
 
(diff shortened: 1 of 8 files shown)


src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -776,28 +697,33 @@ private void UpdateGroupingForUWP()
             _groupListSelectionFlag = false;
         }
 
-        private List<SuggestionsGrouped>? _groupedSuggestions;
+        // private List<SuggestionsGrouped>? _groupedSuggestions;
```
mstefarov, line 700: Can this commented-out field be removed?

src/Toolkit/Toolkit/UI/Controls/SearchView/SearchViewModel.cs
```diff
@@ -238,18 +238,29 @@ public MapPoint? QueryCenter
         /// Gets or sets the list of available search sources, which can be updated dynamically.
         /// </summary>
         /// <remarks>See <see cref="ConfigureDefaultWorldGeocoder(CancellationToken)"/> for a convenient method to populate this collection automatically.</remarks>
-        public IList<ISearchSource> Sources { get; set; } = new ObservableCollection<ISearchSource>();
+        public IList<ISearchSource> Sources { get; set; } = new SourceSourceCollection();
```
mstefarov, line 241: Was this meant to be named "<i>Search</i>SourceCollection?"

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -820,6 +744,135 @@ internal SuggestionsGrouped(ISearchSource owningSource, IEnumerable<SearchSugges
...
+        private SearchView _owner;
+        internal SearchViewTemplateSettings(SearchView owner)
+        {
+            _owner = owner;
+        }
+
+        /// <summary>
+        /// Gets the visibility for the presentation of the <see cref="SearchView.NoResultMessage"/>.
+        /// </summary>
+        public Visibility ResultMessageVisibility
```
mstefarov, line 794: This probably should've been named *No*ResultMessageVisibility, but I would understand if you don't want to mess with existing names as part of this PR

mstefarov (approved):

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -820,6 +744,135 @@ internal SuggestionsGrouped(ISearchSource owningSource, IEnumerable<SearchSugges
             System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator() => _suggestions.GetEnumerator();
         }
 #endif
+
+        /// <summary>
+        /// SearchViewTemplateSettings provides a set of properties that are used when you define a new control template for a control that derives from <see cref="SearchView"/>.
```
prathameshnarkhede, line 749:
```suggestion
        /// <see cref="SearchViewTemplateSettings"/> provides a set of properties that are used when you define a new control template for a control that derives from <see cref="SearchView"/>.
```

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -820,6 +744,135 @@ internal SuggestionsGrouped(ISearchSource owningSource, IEnumerable<SearchSugges
...
+        public static readonly DependencyProperty TemplateSettingsProperty =
+            DependencyProperty.Register(nameof(TemplateSettings), typeof(SearchViewTemplateSettings), typeof(SearchView), new PropertyMetadata(null));
+#elif WPF
+        internal static readonly DependencyPropertyKey TemplateSettingsPropertyKey =
+            DependencyProperty.RegisterReadOnly(nameof(TemplateSettings), typeof(SearchViewTemplateSettings), typeof(SearchView), new FrameworkPropertyMetadata());
+#endif
+    }
+
+    /// <summary>
+    /// SearchViewTemplateSettings provides a set of properties that are used when you define a new control template for a control that derives from <see cref="SearchView"/>.
```
prathameshnarkhede, line 774:
```suggestion
    /// <see cref="SearchViewTemplateSettings"/> provides a set of properties that are used when you define a new control template for a control that derives from <see cref="SearchView"/>.
```

src/Toolkit/Toolkit/UI/Controls/SearchView/SearchViewModel.cs
```diff
@@ -238,18 +238,29 @@ public MapPoint? QueryCenter
...
+        private void SetResults(IEnumerable<SearchResult>? results)
+        {
+            SetPropertyChanged<IList<SearchResult>?>(results is null ? null : new SearchResultList(results), ref _results, nameof(Results));
+        }
 
         /// <summary>
         /// Gets the list of search suggestions. This value is set after calls to <see cref="UpdateSuggestions"/>.
         /// </summary>
-        public IList<SearchSuggestion>? Suggestions { get => _suggestions; private set => SetPropertyChanged(value, ref _suggestions); }
+        public IList<SearchSuggestion>? Suggestions { get => _suggestions; } //private set => SetPropertyChanged(value, ref _suggestions); }
```
prathameshnarkhede, line 257:
```suggestion
        public IList<SearchSuggestion>? Suggestions { get => _suggestions; }
```

prathameshnarkhede (commented):
Just some minor queries to fix.

PR #624, opened 2025-01-26
Adding barcode scanner to Android

dotMorten:
Port of kotlin barcode scanner:
https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/30b0339327181595db2c889eb33c7ce8f7914a8d/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/barcode/BarcodeScanner.kt

Follow-up to #615

### DIFF
--- a/src/Toolkit/Toolkit.Maui/Esri.ArcGISRuntime.Toolkit.Maui.csproj
+++ b/src/Toolkit/Toolkit.Maui/Esri.ArcGISRuntime.Toolkit.Maui.csproj
@@ -16,6 +16,7 @@
         <DefineConstants>$(DefineConstants);MAUI</DefineConstants>
         <Nullable>annotations</Nullable>
         <WindowsSdkPackageVersion Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinAppSDKVersion)','1.6.240923002'))">10.0.19041.38</WindowsSdkPackageVersion> <!-- Workaround for older .NET SDKs. Not needed with VS17.12+  -->
+        <NoWarn>$(NoWarn);NU1608</NoWarn>
     </PropertyGroup>
 
     <ItemGroup>
@@ -37,7 +38,6 @@
 
     <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net8.0'))">
       <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiVersion)" />
-      <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
       <PackageReference Include="Markdig" Version="0.31.0" />
     </ItemGroup>
     <ItemGroup Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'android'">
--- a/src/Toolkit/Toolkit.Maui/Esri.ArcGISRuntime.Toolkit.Maui.csproj
+++ b/src/Toolkit/Toolkit.Maui/Esri.ArcGISRuntime.Toolkit.Maui.csproj
@@ -15,6 +15,7 @@
         <TargetPlatformMinVersion Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">10.0.$(ArcGISRuntime_MinimumSupportedWindowsBuild).0</TargetPlatformMinVersion>
         <DefineConstants>$(DefineConstants);MAUI</DefineConstants>
         <Nullable>annotations</Nullable>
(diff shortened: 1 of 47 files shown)


src/Toolkit/Toolkit.Maui/Esri.ArcGISRuntime.Toolkit.Maui.csproj
```diff
@@ -17,6 +17,7 @@
         <Nullable>annotations</Nullable>
         <WindowsSdkPackageVersion Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(WinAppSDKVersion)','1.6.240923002'))">10.0.19041.38</WindowsSdkPackageVersion> <!-- Workaround for older .NET SDKs. Not needed with VS17.12+  -->
         <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
+        <NoWarn>$(NoWarn);NU1608</NoWarn>
```
dotMorten, line 20: Note: There's no way to reconcile all Android package warnings at this point that the AndroidX libraries introduce. Talking to the Android team, merely ignoring this is what they also do and should be ok.

src/Toolkit/Toolkit.Maui/FeatureForm/BarcodeImageAnalyzer.Android.cs
```diff
@@ -0,0 +1,175 @@
...
+                _onSuccess(_barcodeMap.Values.ToList());
+            }
+        }
+
+        private Matrix? GetTransformationMatrix(IImageProxy imageProxy)
+        {
+            var analysisToTarget = new Matrix();
+            if (_sensorToTargetMatrix == null)
+            {
+                imageProxy.Close();
```
mstefarov, line 128: Is it safe to Close on an IImageProxy twice? You do it here and also in ProcessBarcodes after the foreach loop.
Maybe you could just call GetTransformationMatrix once before the loop, and bail out if we can't get the transformation matrix.

src/Toolkit/Toolkit.Maui/FeatureForm/BarcodeImageAnalyzer.Android.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        void ImageAnalysis.IAnalyzer.Analyze(IImageProxy image)
+        {
+            var mediaImage = image.Image;
+            if (mediaImage != null)
+            {
+                _barcodeScanner.Process(mediaImage, image.ImageInfo.RotationDegrees)
+                    .AddOnSuccessListener(new OnSuccessListener(barcodes =>
+                    {
+                        _frames = (_frames + 1) % long.MaxValue;
```
mstefarov, line 65: It's good to think ahead to the 9223372036854775808th frame 😀 

The wrapping around might make `.Where(info => info.LastSeenFrame < _frames - 4)` (in ProcessBarcodes below) behave oddly, but I'm not sure how realistic is this scenario.

src/Toolkit/Toolkit.Maui/FeatureForm/BarcodeScanner.Android.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                autoScanDelayCancellationToken?.Dispose();
+                cameraController?.Dispose();
+                executor?.Dispose();
```
mstefarov, line 115: Isn't this the main executor? Are we supposed to be disposing that?

This can probably be tested by opening/closing barcode scanner a couple times during the same app session, just to make sure we aren't accidentally disposing a long-lived shared object here.

src/Toolkit/Toolkit.Maui/FeatureForm/BarcodeImageAnalyzer.Android.cs
dotMorten: This is coming from the direct port of kotlin: https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/blob/30b0339327181595db2c889eb33c7ce8f7914a8d/toolkit/featureforms/src/main/java/com/arcgismaps/toolkit/featureforms/internal/components/barcode/BarcodeImageAnalyzer.kt#L82

src/Toolkit/Toolkit.Maui/FeatureForm/BarcodeScanner.Android.cs
dotMorten: GetMainExecutor that creates the executor returns a new Java Object wrapper each time which must be disposed once we're done with it (it doesn't destroy the underlying java executor). Multiple scans are no issue.

src/Toolkit/Toolkit.Maui/FeatureForm/BarcodeImageAnalyzer.Android.cs
dotMorten: There's no issue: 
image

However it's extremely important that it does get closed, or the app will stop processing frames

mstefarov (approved):

Issue #625, opened 2025-02-12
basemap tool cause unhandled exception in MAUI if no connection


federicoMoroni-git:
I am using a the change basemap tool in my MAUI application. in the XAML I have

`
xmlns:arcgistoolkit="clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui;assembly=Esri.ArcGISRuntime.Toolkit.Maui"

<arcgistoolkit:BasemapGallery 
GeoModel="{Binding Map}"
MaximumHeightRequest="{OnPlatform Default=300, iOS=300, Android=300}"
MaximumWidthRequest="{OnPlatform Default=300, iOS=280, Android=280}" />
`

If I start the app with no internet connection this component cause an unhandled exception from this block
`
#if DEBUG && !DISABLE_XAML_GENERATED_BREAK_ON_UNHANDLED_EXCEPTION
            UnhandledException += (sender, e) =>
            {
                if (global::System.Diagnostics.Debugger.IsAttached) global::System.Diagnostics.Debugger.Break();
            };
#endif
`

where the sender is {appName.WinUI.App} and e is {Microsoft.UI.Xaml.UnhandledExceptionEventArgs}, No such host is known. (www.arcgis.com:443)

this exception causes the app to crash. I don`t see how I can handle this in the backend.

PR #627, opened 2025-02-25
Introduced error handling in the `BasemapGallery_Loaded` method to avoid crash

prathameshnarkhede:
## Related Issue #625 

- Introduced error handling in the `BasemapGallery_Loaded` method for both `BasemapGallery.cs` and `BasemapGallery.Windows.cs`.

- These changes prevent crashes and provide useful debug information during the basemap loading process.

### DIFF
--- a/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
+++ b/src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
@@ -46,12 +46,19 @@ public partial class BasemapGallery
 
     private async void BasemapGallery_Loaded(object? sender, EventArgs e)
     {
-        // Unsubscribe from the Loaded event to ensure this only runs once.
-        Loaded -= BasemapGallery_Loaded;
-
-        if (AvailableBasemaps is null)
+        try
         {
-            await _controller.LoadFromDefaultPortal();
+            // Unsubscribe from the Loaded event to ensure this only runs once.
+            Loaded -= BasemapGallery_Loaded;
+
+            if (AvailableBasemaps is null)
+            {
+                await _controller.LoadFromDefaultPortal();
+            }
+        }
+        catch (Exception ex)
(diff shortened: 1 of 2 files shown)


src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
```diff
@@ -44,16 +44,23 @@ public BasemapGallery()
...
+            Loaded -= BasemapGallery_Loaded;
+
+            if (AvailableBasemaps is null)
+            {
+                await _controller.LoadFromDefaultPortal();
+            }
+        }
+        catch (Exception ex)
+        {
+            System.Diagnostics.Debug.WriteLine($"Error loading basemaps: {ex.Message}");
```
dotMorten, line 61: Debug.WriteLine doesn't work in release build of the API which is all we ship. You will need Trace.WriteLine for that.
However we should already be writing out load errors to the debug console.

src/Toolkit/Toolkit.Maui/BasemapGallery/BasemapGallery.cs
```diff
@@ -44,16 +44,23 @@ public BasemapGallery()
...
-        {
-            await _controller.LoadFromDefaultPortal();
-        }
-    }
+    private async void BasemapGallery_Loaded(object? sender, EventArgs e)
+    {
+        try
+        {
+            // Unsubscribe from the Loaded event to ensure this only runs once.
+            Loaded -= BasemapGallery_Loaded;
```
dotMorten, line 52: does this need to be in the try?

prathameshnarkhede: Not necessarily. Removing it out.

mstefarov (approved):

dotMorten (approved):

PR #642, opened 2025-05-08
Only show validation errors if a field got focus

dotMorten:
Validation errors should only start showing when a control gets focus the first time.
`*InputView` controls notify the `FieldFormElementView` parent instance that it has received focus. This will make the `FieldFormElementView` display the validation error. In addition `TextFormInputView` shows the red outline, and this has the same form of first-time-focus tracking to show its error border.
Similarly the validation errors should go away on apply or discard and reset the "have had focus" flag.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Windows.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Windows.cs
@@ -20,6 +20,20 @@ namespace Esri.ArcGISRuntime.Toolkit.Primitives
 #endif
         {
             base.OnApplyTemplate();
+            if (_datePicker != null)
+            {
+                _datePicker.GotFocus -= Picker_GotFocus;
+                _datePicker.SelectedDateChanged -= DatePicker_SelectedDateChanged;
+            }
+            if (_timePicker != null)
+            {
+                _timePicker.GotFocus -= Picker_GotFocus;
+#if WINDOWS_XAML
+                _timePicker.SelectedTimeChanged -= TimePicker_SelectedTimeChanged;
+#else
+                _timePicker.TimeChanged -= TimePicker_TimeChanged;
+#endif
+            }
             _datePicker = GetTemplateChild("DatePicker") as DatePicker;
             _timePicker = GetTemplateChild("TimePicker") as TimePicker;
             if (_datePicker != null)
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Windows.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Windows.cs
@@ -24,10 +24,12 @@ namespace Esri.ArcGISRuntime.Toolkit.Primitives
             _timePicker = GetTemplateChild("TimePicker") as TimePicker;
             if (_datePicker != null)
             {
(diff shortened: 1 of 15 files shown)


src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.Windows.cs
```diff
@@ -54,15 +54,36 @@ public bool IsValid
...
-            return parent as UI.Controls.FeatureFormView;
+            return parent as T;
+        }
+
+        internal static IEnumerable<T> GetDescendentsOfType<T>(FrameworkElement root)
+        {
+            if (root is null)
+                yield break;
+
+            for(int i = 0; i<VisualTreeHelper.GetChildrenCount(root);i++)
```
mstefarov, line 75: Nitpick: this file could use an auto-format (ctrl+k, ctrl+d)

mstefarov (approved):

src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Windows.cs
```diff
@@ -24,10 +24,12 @@ protected override void OnApplyTemplate()
             _timePicker = GetTemplateChild("TimePicker") as TimePicker;
             if (_datePicker != null)
             {
+                _datePicker.GotFocus += Picker_GotFocus;
                 _datePicker.SelectedDateChanged += DatePicker_SelectedDateChanged;
             }
             if (_timePicker != null)
             {
+                _timePicker.GotFocus += Picker_GotFocus;
```
prathameshnarkhede, line 32: Should we also detach these handlers before `GetTemplateChild` and attaching here like we did in `ComboboxFormInputView`?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.Maui.cs
```diff
@@ -146,6 +146,26 @@ internal static Style GetStyle(string resourceKey, Style defaultStyle)
...
+            {
+                if (child == root) continue;
+                if (child is Element frameworkElement)
+                {
+                    if (frameworkElement is T targetElement)
+                        yield return targetElement;
+
+                    foreach (var descendant in GetDescendentsOfType<T>(frameworkElement))
+                        yield return descendant;
+                }
```
prathameshnarkhede, line 166: I noticed that we're using the `GetVisualTreeDescendants()` method. Since this method already returns all descendants in the visual tree, do we still need to implement recursion here? It seems like `GetVisualTreeDescendants()` should handle this for us.

prathameshnarkhede: Just a thought :
if we don’t actually need recursion here, we could putin a bit of LINQ and write the piece like this:

```C#
internal static IEnumerable<T> GetDescendentsOfType<T>(Element root)
{
    return root == null
        ? Enumerable.Empty<T>()
        : root.GetVisualTreeDescendants()
              .Where(child => child != root)
              .OfType<T>();
}
```

dotMorten: I'm not quite following. We do need recursion

dotMorten: Good catch on the difference in behavior from WPF/WinUI here

PR #645, opened 2025-05-16
Fix: MAUI Floor Filter Crash on Site Selection in .NET 9

prathameshnarkhede:
Updated navigation methods to use `PushAsync` and `PopAsync` instead of `PushModalAsync` and `PopModalAsync`, shifting from modal to standard navigation.

This is temporary fix that prevents a MAUI bug reported
dotnet/maui#29512

### DIFF
--- a/src/Toolkit/Toolkit.Maui/FloorFilter/FloorFilter.cs
+++ b/src/Toolkit/Toolkit.Maui/FloorFilter/FloorFilter.cs
@@ -170,8 +170,8 @@ public partial class FloorFilter : TemplatedView
             finally
             {
                 _navigationStackCounter--;
+            }
         }
-    }
     }
 
     internal async void GoBack()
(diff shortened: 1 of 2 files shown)


dotMorten:
Unfortunately we need Modal here. We don't want to completely navigate away from the underlying active mapview page.

src/Toolkit/Toolkit.Maui/FloorFilter/FloorFilter.cs
```diff
@@ -146,7 +146,11 @@ internal async void NavigateForward(ContentPage page)
     {
         try
         {
+#if WINDOWS && ANDROID
```
dotMorten, line 149: Please add link to maui issue and explain here why we do this different. Same below.

dotMorten (approved):

PR #650, opened 2025-05-29
New FileDownloadTask helper

dotMorten:
API:
image

### DIFF
--- a/src/Samples/Toolkit.SampleApp.Maui/Samples/FileDownloadSample.xaml
+++ b/src/Samples/Toolkit.SampleApp.Maui/Samples/FileDownloadSample.xaml
@@ -1,0 +1,44 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:esriUI="clr-namespace:Esri.ArcGISRuntime.Maui;assembly=Esri.ArcGISRuntime.Maui"
+             xmlns:esriTK="clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui;assembly=Esri.ArcGISRuntime.Toolkit.Maui"
+             xmlns:vm="clr-namespace:Esri.ArcGISRuntime.Toolkit.Samples"
+             x:Class="Toolkit.SampleApp.Maui.Samples.FileDownloadSample" Title="File Download">
+    <Grid>
+
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto" />
+            <RowDefinition />
+        </Grid.RowDefinitions>
+
+        <Label Margin="8" LineBreakMode="WordWrap">Downloads can be resumed, even in the event of app crash. Try starting a download, and restart the application mid-download. You'll be able to resume the download after.</Label>
+
+        <CollectionView x:Name="DownloadItems" Margin="6,0" Grid.Row="1" >
+            <CollectionView.ItemTemplate>
+                <DataTemplate x:DataType="vm:DownloadItemVM">
+                    <Grid Margin="2">
+                        <Grid.ColumnDefinitions>
+                            <ColumnDefinition Width="*" />
+                            <ColumnDefinition Width="Auto" />
+                            <ColumnDefinition Width="Auto" />
+                        </Grid.ColumnDefinitions>
+                        <ProgressBar Progress="{Binding Progress, Mode=OneWay}" HorizontalOptions="Fill" VerticalOptions="End" Opacity=".25" Grid.ColumnSpan="3" />
+                        <Label Text="{Binding Item.Name, Mode=OneWay}" VerticalOptions="Center" HorizontalOptions="Start" Margin="4" FontSize="16" LineBreakMode="TailTruncation" />
+                        <HorizontalStackLayout Grid.Column="1" HorizontalOptions="End" >
+                            <Label Text="{Binding DownloadSpeed, Mode=OneWay}" VerticalOptions="Center" FontAttributes="Bold" />
+                            <Label Text="{Binding Progress, Mode=OneWay, StringFormat='{0:P1}'}" VerticalOptions="Center" FontAttributes="Bold" Margin="4,0" />
+                            <Label Text="{Binding Status, Mode=OneWay}" VerticalOptions="Center" />
+                        </HorizontalStackLayout>
+                        <HorizontalStackLayout Margin="4" Grid.Column="2">
+                            <Button Command="{Binding StartDownloadCommand, Mode=OneWay}" Text="Play"/>
+                            <Button Command="{Binding PauseDownloadCommand, Mode=OneWay}" Text="Pause"/>
+                            <Button Command="{Binding CancelDownloadCommand, Mode=OneWay}" Text="Cancel"/>
+                            <Button Command="{Binding DeleteDownloadCommand, Mode=OneWay}" Text="Delete"/>
+                        </HorizontalStackLayout>
+                    </Grid>
+                </DataTemplate>
+            </CollectionView.ItemTemplate>
+        </CollectionView>
+    </Grid>
+</ContentPage>
--- a/src/Samples/Toolkit.SampleApp.Maui/Toolkit.SampleApp.Maui.csproj
+++ b/src/Samples/Toolkit.SampleApp.Maui/Toolkit.SampleApp.Maui.csproj
@@ -9,6 +9,7 @@
     <SingleProject>true</SingleProject>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <DefineConstants>$(DefineConstants);MAUI</DefineConstants>
 
     <WarningsAsErrors>$(WarningsAsErrors);IL2036;IL2037</WarningsAsErrors>
 
@@ -53,6 +54,10 @@
     <Compile Remove="Samples\TimeSliderSample.xaml.cs" />
     <MauiXaml Remove="Samples\LayerLegendSample.xaml" />
     <MauiXaml Remove="Samples\TimeSliderSample.xaml" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="..\Toolkit.SampleApp.WPF\Samples\Downloading\DownloadItemVM.cs" Link="Samples\DownloadItemVM.cs" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Samples/Toolkit.SampleApp.UWP/Samples/Downloading/FileDownloadSample.xaml
+++ b/src/Samples/Toolkit.SampleApp.UWP/Samples/Downloading/FileDownloadSample.xaml
@@ -1,0 +1,61 @@
+﻿<Page
+    x:Class="Esri.ArcGISRuntime.Toolkit.SampleApp.Samples.Downloading.FileDownloadSample"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:sys="using:System"
+    xmlns:local="using:Esri.ArcGISRuntime.Toolkit.SampleApp.Downloading.Samples"
+    xmlns:local2="using:Esri.ArcGISRuntime.Toolkit.Samples"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <Grid>
+        <Grid.Resources>
+            <Style TargetType="Button">
+                <Setter Property="Width" Value="50" />
+                <Setter Property="Margin" Value="4" />
+                <Setter Property="Background" Value="CornflowerBlue" />
+                <Setter Property="Foreground" Value="White" />
+            </Style>
+            <x:Double x:Key="ProgressBarMinHeight">55</x:Double>
+            <x:Double x:Key="ProgressBarTrackHeight">0</x:Double>
+        </Grid.Resources>
+
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto" />
+            <RowDefinition />
+        </Grid.RowDefinitions>
+
+        <TextBlock TextWrapping="Wrap" Margin="8">Downloads can be resumed, even in the event of app crash. Try starting a download, and restart the application mid-download. You'll be able to resume the download after.</TextBlock>
+
+        <ItemsControl x:Name="DownloadItems" Margin="6,0" Grid.Row="1" ItemsSource="{x:Bind Items}">
+            <ItemsControl.ItemTemplate>
+                <DataTemplate x:DataType="local2:DownloadItemVM">
+                    <Grid Margin="2" Height="{StaticResource ProgressBarMinHeight}">
+                        <Grid.ColumnDefinitions>
+                            <ColumnDefinition Width="*" />
+                            <ColumnDefinition Width="Auto" />
+                            <ColumnDefinition Width="Auto" />
+                        </Grid.ColumnDefinitions>
+                        <ProgressBar Minimum="0" Maximum="1" Value="{x:Bind Progress, Mode=OneWay}" HorizontalAlignment="Stretch" VerticalAlignment="Stretch" Opacity=".25" Grid.ColumnSpan="3" MinHeight="{StaticResource ProgressBarMinHeight}" />
+                        <TextBlock Text="{x:Bind Item.Name, Mode=OneWay}" VerticalAlignment="Center" HorizontalAlignment="Left" Margin="4" FontSize="16" TextTrimming="CharacterEllipsis" />
+                        <StackPanel Orientation="Horizontal" Grid.Column="1" HorizontalAlignment="Right" >
+                            <TextBlock Text="{x:Bind DownloadSpeed, Mode=OneWay}" VerticalAlignment="Center" FontWeight="Bold" />
+                            <TextBlock Text="{x:Bind sys:String.Format('{0:P1}', Progress), Mode=OneWay}" VerticalAlignment="Center" FontWeight="Bold" Margin="4,0" />
+                            <TextBlock Text="{x:Bind Status, Mode=OneWay}" VerticalAlignment="Center" />
+                        </StackPanel>
+                        <StackPanel Orientation="Horizontal" Margin="4" Grid.Column="2">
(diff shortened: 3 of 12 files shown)


src/Samples/Toolkit.SampleApp.UWP/Samples/Downloading/FileDownloadSample.xaml
```diff
@@ -0,0 +1,61 @@
...
+                        <StackPanel Orientation="Horizontal" Grid.Column="1" HorizontalAlignment="Right" >
+                            <TextBlock Text="{x:Bind DownloadSpeed, Mode=OneWay}" VerticalAlignment="Center" FontWeight="Bold" />
+                            <TextBlock Text="{x:Bind sys:String.Format('{0:P1}', Progress), Mode=OneWay}" VerticalAlignment="Center" FontWeight="Bold" Margin="4,0" />
+                            <TextBlock Text="{x:Bind Status, Mode=OneWay}" VerticalAlignment="Center" />
+                        </StackPanel>
+                        <StackPanel Orientation="Horizontal" Margin="4" Grid.Column="2">
+                            <Button Command="{x:Bind StartDownloadCommand, Mode=OneWay}"><SymbolIcon Symbol="Play" /></Button>
+                            <Button Command="{x:Bind PauseDownloadCommand, Mode=OneWay}" ><SymbolIcon Symbol="Pause" /></Button>
+                            <Button Command="{x:Bind CancelDownloadCommand, Mode=OneWay}" ><SymbolIcon Symbol="Cancel" /></Button>
+                            <Button Command="{x:Bind DeleteDownloadCommand, Mode=OneWay}" ><SymbolIcon Symbol="Delete" /></Button>
```
prathameshnarkhede, line 52:
```suggestion
                            <Button Command="{x:Bind PauseDownloadCommand, Mode=OneWay}"><SymbolIcon Symbol="Pause" /></Button>
                            <Button Command="{x:Bind CancelDownloadCommand, Mode=OneWay}"><SymbolIcon Symbol="Cancel" /></Button>
                            <Button Command="{x:Bind DeleteDownloadCommand, Mode=OneWay}"><SymbolIcon Symbol="Delete" /></Button>
```

src/Samples/Toolkit.SampleApp.Maui/Samples/FileDownloadSample.xaml
```diff
@@ -0,0 +1,44 @@
...
+                <DataTemplate x:DataType="vm:DownloadItemVM">
+                    <Grid Margin="2">
+                        <Grid.ColumnDefinitions>
+                            <ColumnDefinition Width="*" />
+                            <ColumnDefinition Width="Auto" />
+                            <ColumnDefinition Width="Auto" />
+                        </Grid.ColumnDefinitions>
+                        <ProgressBar Progress="{Binding Progress, Mode=OneWay}" HorizontalOptions="Fill" VerticalOptions="End" Opacity=".25" Grid.ColumnSpan="3" />
+                        <Label Text="{Binding Item.Name, Mode=OneWay}" VerticalOptions="Center" HorizontalOptions="Start" Margin="4" FontSize="16" LineBreakMode="TailTruncation" />
+                        <HorizontalStackLayout Grid.Column="1" HorizontalOptions="End" >
```
prathameshnarkhede, line 28:
```suggestion
                        <HorizontalStackLayout Grid.Column="1" HorizontalOptions="End">
```

src/Samples/Toolkit.SampleApp.Maui/Samples/FileDownloadSample.xaml
```diff
@@ -0,0 +1,44 @@
...
+    <Grid>
+
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto" />
+            <RowDefinition />
+        </Grid.RowDefinitions>
+
+        <Label Margin="8" LineBreakMode="WordWrap">Downloads can be resumed, even in the event of app crash. Try starting a download, and restart the application mid-download. You'll be able to resume the download after.</Label>
+
+        <CollectionView x:Name="DownloadItems" Margin="6,0" Grid.Row="1" >
```
prathameshnarkhede, line 17:
```suggestion
        <CollectionView x:Name="DownloadItems" Margin="6,0" Grid.Row="1">
```

src/Samples/Toolkit.SampleApp.Maui/Samples/FileDownloadSample.xaml
```diff
@@ -0,0 +1,44 @@
...
+                        <HorizontalStackLayout Grid.Column="1" HorizontalOptions="End" >
+                            <Label Text="{Binding DownloadSpeed, Mode=OneWay}" VerticalOptions="Center" FontAttributes="Bold" />
+                            <Label Text="{Binding Progress, Mode=OneWay, StringFormat='{0:P1}'}" VerticalOptions="Center" FontAttributes="Bold" Margin="4,0" />
+                            <Label Text="{Binding Status, Mode=OneWay}" VerticalOptions="Center" />
+                        </HorizontalStackLayout>
+                        <HorizontalStackLayout Margin="4" Grid.Column="2">
+                            <Button Command="{Binding StartDownloadCommand, Mode=OneWay}" Text="Play"/>
+                            <Button Command="{Binding PauseDownloadCommand, Mode=OneWay}" Text="Pause"/>
+                            <Button Command="{Binding CancelDownloadCommand, Mode=OneWay}" Text="Cancel"/>
+                            <Button Command="{Binding DeleteDownloadCommand, Mode=OneWay}" Text="Delete"/>
```
prathameshnarkhede, line 37:
```suggestion
                            <Button Command="{Binding StartDownloadCommand, Mode=OneWay}" Text="Play" />
                            <Button Command="{Binding PauseDownloadCommand, Mode=OneWay}" Text="Pause" />
                            <Button Command="{Binding CancelDownloadCommand, Mode=OneWay}" Text="Cancel" />
                            <Button Command="{Binding DeleteDownloadCommand, Mode=OneWay}" Text="Delete" />
```

src/Samples/Toolkit.SampleApp.Maui/Toolkit.SampleApp.Maui.csproj
```diff
@@ -9,6 +9,7 @@
     <SingleProject>true</SingleProject>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <DefineConstants>$(DefineConstants);MAUI</DefineConstants>
```
prathameshnarkhede, line 12: Do we need this? Or we missed it using in `DownloadItemVM.cs`?

prathameshnarkhede (approved):
Just some minor suggestions. Otherwise, Looks good!

Also, Recently I read about `SemaphoreSlim` and seems this could be a good candidate for its usage.

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,804 @@
...
+    {
+        RequestUri = requestUri;
+        this.client = client;
+        this.Filename = filename;
+#if NETFX_CORE || __IOS__ || __ANDROID__
+        this.TempFile = Path.GetTempFileName();
+#else
+        this.TempFile = Path.Combine(ArcGISRuntimeEnvironment.TempPath, Path.GetRandomFileName());
+#endif
+        this.TempFile = Path.GetTempFileName();
```
mstefarov: I'm guessing you did not mean to assign TempFile twice here

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,804 @@
...
+    /// <summary>
+    /// Downloads the data content from a Portal Item.
+    /// </summary>
+    /// <param name="portalItem">Portal Item who's data should be downloaded</param>
+    /// <param name="destinationPath">File path to save the download to.</param>
+    /// <returns>The download task.</returns>
+    /// <exception cref="IOException">The destination file already exists</exception>.
+    public static Task<FileDownloadTask> BeginDownloadAsync(this Portal.PortalItem portalItem, string destinationPath)
+    {
+        string requestUri = $"{portalItem.Portal.Uri.OriginalString.TrimEnd('/') + "/sharing/rest/"}/content/items/{portalItem.ItemId}/data";
```
mstefarov: Can be simplified:
```suggestion
        string requestUri = $"{portalItem.Portal.Uri.OriginalString.TrimEnd('/')}/sharing/rest/content/items/{portalItem.ItemId}/data";
```

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,804 @@
...
+            {
+                Status = FileDownloadStatus.Cancelled;
+            }
+            else
+            {
+                cancellationSource?.Cancel();
+            }
+            if (task is not null)
+            {
+                return task.ContinueWith(t => { File.Delete(TempFile); BytesDownloaded = 0; Status = FileDownloadStatus.Cancelled; ProcessQueue(); });
```
mstefarov: This continuation won't actually run if task faulted or was canceled, right?  I would test if this works as intended.

And since BeginTransfer always calls ProcessQueue at the end, do you need to call it again here?

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,804 @@
...
+    /// </summary>
+    /// <returns></returns>
+    public async Task DownloadAsync()
+    {
+        if (Status == FileDownloadStatus.Paused || Status == FileDownloadStatus.Error)
+            await ResumeAsync().ConfigureAwait(false);
+        else if (transferTask == null)
+        {
+            throw new InvalidOperationException();
+        }
```
mstefarov: It seems this will throw if FileDownloadTask is queued -- is that right?  Awaiting a queued-up file until downloads seems like a reasonable scenario.

Also, what about tasks in "Cancelled" status -- should they automatically call ResumeAsync (the way "Error" tasks do) or fail with an exception?

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,804 @@
...
+        cancellationSource = new CancellationTokenSource();
+        IsResumable = content.Headers.AcceptRanges?.Contains("bytes") == true;
+        if (IsResumable)
+        {
+            ETag = content.Headers.ETag?.Tag;
+            Date = content.Headers.Date;
+        }
+        var token = cancellationSource.Token;
+        Status = FileDownloadStatus.Downloading;
+        byte[] buffer = new byte[BufferSize];
```
mstefarov, line 707: This would be a great place to use the shared array pool

src/Samples/Toolkit.SampleApp.WPF/Samples/Downloading/DownloadItemVM.cs
```diff
@@ -0,0 +1,199 @@
...
+            {
+                if (Item is null)
+                    return "Loading item data...";
+                if (DownloadTask is null)
+                    return System.IO.File.Exists(Filename) ? "Completed" : "Not started";
+                return DownloadTask.Status.ToString();
+            }
+        }
+
+        private string Filename => Item is null ? null : System.IO.Path.Combine(_folder, Item?.Name!); // Where we will save the file
```
mstefarov: I wouldn't trust item Name to be a valid filename.  I would add something that [strips invalid characters out](https://stackoverflow.com/a/12800424).

Actually, I'm not sure if I could trust the Name to be unique either.  Would be it safer to just use an item ID for filename or possibly some combination of ID + sanitized Name?

dotMorten: This will be the filename of the item though, and in addition there's a limited set of items which all create valid filenames.

src/Toolkit/Toolkit/FileDownloadTask.cs
dotMorten: That feels overkill considering the small buffer size, and the likely very limited number of download tasks you'd have running.

src/Toolkit/Toolkit/FileDownloadTask.cs
dotMorten: I'm not following. Tasks will always run the continuewith, regardless of success or not, unless you use the optional overload to only do it on success (which isn't the case here)
Good call on processqueue though.

src/Toolkit/Toolkit/FileDownloadTask.cs
dotMorten: Cancelled tasks can't be resumed. Made it a bit clearer in latest commit.

src/Samples/Toolkit.SampleApp.Maui/Toolkit.SampleApp.Maui.csproj
dotMorten: Yes it was missing.

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,795 @@
...
+        {
+            request.Headers.Range = new System.Net.Http.Headers.RangeHeaderValue(offset, null);
+            if (ETag != null)
+                request.Headers.IfRange = new System.Net.Http.Headers.RangeConditionHeaderValue(ETag);
+            else if (Date != null)
+                request.Headers.IfRange = new System.Net.Http.Headers.RangeConditionHeaderValue(Date.Value);
+        }
+        try
+        {
+            var response = await client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead).ConfigureAwait(false);
```
mstefarov: Could the CancellationTokenSource be created earlier (in BeginDownload instead of BeginTransfer), so that it can be passed here too? This initial connection can potentially be quite slow.

src/Samples/Toolkit.SampleApp.WPF/Samples/Downloading/DownloadItemVM.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        public double Progress
+        {
+            get
+            {
+                if (Item is null) return 0;
+                if (System.IO.File.Exists(Filename)) return 1; // Already downloaded
+                if (DownloadTask is null) return 0;
+                if (!DownloadTask.TotalLength.HasValue) return 0; // Unknown size
+                return DownloadTask.BytesDownloaded / (double)DownloadTask.TotalLength.Value;
```
mstefarov, line 168: AGOL allows empty files, but if TotalLength is 0 then this would be division-by-zero / NaN, and the binding will throw.  Try it yourself: https://www.arcgis.com/home/item.html?id=59678000f2084c3c890e7b8983116f49

```suggestion
                if (DownloadTask.TotalLength == 0) return 1;
                return DownloadTask.BytesDownloaded / (double)DownloadTask.TotalLength.Value;
```

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,795 @@
...
+        return task.ContinueWith(t => { Status = FileDownloadStatus.Paused; });
+    }
+
+    private async Task BeginDownload(long offset)
+    {
+        if (offset == TotalLength)
+        {
+            transferTask = Task.CompletedTask;
+            Status = FileDownloadStatus.Completed;
+            return;
```
mstefarov: If I am understanding this correctly:
If we got to this code, we are resuming after File.Move failed.  We should probably check that the destination file exists and is the right size before calling the task "Completed".  You could even retry moving.

Something like...

```cs
        if (offset == TotalLength)
        {
            // Looks like the download to temp file finished, but moving possibly failed
            bool canComplete = false;
            if (File.Exists(Filename) && new FileInfo(Filename).Length == TotalLength)
            {
                canComplete = true;
            }
            else if (File.Exists(TempFile) && new FileInfo(TempFile).Length == TotalLength)
            {
                try
                {
                    File.Move(TempFile, Filename);
                    canComplete = true;
                }
                catch (Exception ex)
                {
                    Exception = ex;
                    Status = FileDownloadStatus.Error;
                    Error?.Invoke(this, ex);
                    throw;
                }
            }
            if (canComplete)
            {
                transferTask = Task.CompletedTask;
                Status = FileDownloadStatus.Completed;
                ProcessQueue();
                return;
            }
            // If we reach here, either the file was not fully downloaded or temp folder was cleared. Restart!
            offset = 0;
        }
```

Also, we should raise the "Completed" event for consistency, and call `ProcessQueue()` before returning.

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -0,0 +1,795 @@
...
+        }
+        try
+        {
+            var response = await client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead).ConfigureAwait(false);
+            content = response.EnsureSuccessStatusCode();
+            transferTask = BeginTransfer(content);
+        }
+        catch (System.Exception)
+        {
+            Status = FileDownloadStatus.Error;
```
mstefarov: If you're setting status to Error, should this also set the Exception property and raise Error event? Otherwise any errors that happen here may not be reflected in the UI.

src/Samples/Toolkit.SampleApp.WPF/Samples/Downloading/DownloadItemVM.cs
dotMorten: I wasn't able to repro an issue with this. However if totallength is 0, it's more likely because the server doesn't report a file-length. In those cases it would be better for the VM to report bytes downloaded, instead of a percentage and set the progress bar to indeterminate. Since nothing in the sample data actually hits this issue, and this is just sample code, I didn't think adding even more to the complexity to the VM was helpful. I'll add a comment instead add the proposed guard just in case.

src/Toolkit/Toolkit/FileDownloadTask.cs
dotMorten: Thanks. All the extra checks really aren't necessary, since only ResumeAsync will call with an offset, and if the temp file doesn't exist, the length will be zero.

src/Toolkit/Toolkit/FileDownloadTask.cs
dotMorten: Instead of moving it, I added the option to specify a cancellationtoken for the initiation of the download that just applies to that bit. This makes it a little cleaner between a cancellation token in ResumeAsync/BeginAsync for just that start-task, and calling CancelAsync() later (the latter which cancels the download stream after it has begun)

src/Toolkit/Toolkit/FileDownloadTask.cs
```diff
@@ -605,10 +610,28 @@ public Task PauseAsync()
...
+            catch (Exception ex)
+            {
+                Exception = ex;
+                Status = FileDownloadStatus.Error;
+                Error?.Invoke(this, ex);
+                throw;
+            }
             transferTask = Task.CompletedTask;
             Status = FileDownloadStatus.Completed;
             return;
```
mstefarov: This should raise the Completed event, to match the other place where we reach "Completed" status.

PR #655, opened 2025-06-06
Bugfix: Fixing bug with EnableDefaultWorldGeocode value not being honored in MAUI

prathameshnarkhede:
- Add Loaded event handler to SearchView for initialization.
- Removed calls ConfigureForCurrentConfiguration as it is being called before EnableDefaultWorldGeocode property is assigned.
- Updated the way default value is set for EnableDefaultWorldGeocode in WPF, UWP and WinUI.

### DIFF
--- a/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
+++ b/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
@@ -84,17 +84,10 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
         SearchCommand = new DelegateCommand(HandleSearchCommand);
         RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
         Loaded += SearchView_Loaded;
-        Unloaded += SearchView_Unloaded;
     }
-
-    private void SearchView_Unloaded(object? sender, EventArgs e) => _loadedHandled = false;
 
     private void SearchView_Loaded(object? sender, EventArgs e)
     {
-        if (_loadedHandled)
-            return;
-        _loadedHandled = true;
-
         if (GeoView != null)
         {
             HandleViewpointChanged();
--- a/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
+++ b/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
@@ -81,6 +81,19 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
         ClearCommand = new DelegateCommand(HandleClearSearchCommand);
         SearchCommand = new DelegateCommand(HandleSearchCommand);
         RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
+        Loaded += SearchView_Loaded;
+    }
+
+    private void SearchView_Loaded(object? sender, EventArgs e)
+    {
+        // Unsubscribe from the Loaded event to ensure this only runs once.
+        Loaded -= SearchView_Loaded;
+
+        if (GeoView != null)
+        {
+            HandleViewpointChanged();
+        }
+        _ = ConfigureForCurrentConfiguration();
     }
 
     private void InitializeLocalizedStrings()
@@ -450,8 +463,6 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
                 newGeoView.ViewpointChanged += sendingView.GeoView_ViewpointChanged;
                 newGeoView.GraphicsOverlays?.Add(sendingView._resultOverlay);
             }
-
-            _ = sendingView.ConfigureForCurrentConfiguration();
         }
     }
 
@@ -531,7 +542,6 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
     {
         if (e.PropertyName == nameof(Mapping.Map) || e.PropertyName == nameof(Scene))
         {
-            _ = ConfigureForCurrentConfiguration();
             return;
         }
 
@@ -545,8 +555,6 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
             {
                 _lastUsedGeomodel = scene;
             }
-
-            _ = ConfigureForCurrentConfiguration();
         }
     }
 
--- a/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
+++ b/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
@@ -48,6 +48,8 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
 
     private bool _sourceSelectToggled;
 
+    private bool _loadedHandled;
+
     /// <summary>
     /// Initializes a new instance of the <see cref="SearchView"/> class.
     /// </summary>
@@ -82,10 +84,17 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
         SearchCommand = new DelegateCommand(HandleSearchCommand);
         RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
         Loaded += SearchView_Loaded;
+        Unloaded += SearchView_Unloaded;
     }
+
+    private void SearchView_Unloaded(object? sender, EventArgs e) => _loadedHandled = false;
 
     private void SearchView_Loaded(object? sender, EventArgs e)
     {
+        if (_loadedHandled)
+            return;
+        _loadedHandled = true;
+
         if (GeoView != null)
         {
             HandleViewpointChanged();
--- a/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
+++ b/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
@@ -86,9 +86,6 @@ public partial class SearchView : TemplatedView, INotifyPropertyChanged
 
     private void SearchView_Loaded(object? sender, EventArgs e)
     {
-        // Unsubscribe from the Loaded event to ensure this only runs once.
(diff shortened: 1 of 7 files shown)


src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
```diff
@@ -81,6 +81,19 @@ public SearchView()
...
+    }
+
+    private void SearchView_Loaded(object? sender, EventArgs e)
+    {
+        // Unsubscribe from the Loaded event to ensure this only runs once.
+        Loaded -= SearchView_Loaded;
+
+        if (GeoView != null)
+        {
+            HandleViewpointChanged();
```
dotMorten, line 93: Why is this change needed? If you don't want the HandleViewpointChanged code to run before load, then there should be a guard for this in the HandleViewpointChanged method as well.

src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
```diff
@@ -81,6 +81,19 @@ public SearchView()
         ClearCommand = new DelegateCommand(HandleClearSearchCommand);
         SearchCommand = new DelegateCommand(HandleSearchCommand);
         RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
+        Loaded += SearchView_Loaded;
+    }
+
+    private void SearchView_Loaded(object? sender, EventArgs e)
+    {
+        // Unsubscribe from the Loaded event to ensure this only runs once.
+        Loaded -= SearchView_Loaded;
```
dotMorten: Could there be a situation where it unloads, sources change and then loads again?

dotMorten (commented):
Is this issue only present in .NET MAUI? Will WPF/WinUI/UWP not have the same issue?

src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
prathameshnarkhede: Right, This is to make sure to load the Viewpoint is set when SearchView is loaded. Adding guard in `HandleViewpointChanged` as wel.

src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
prathameshnarkhede: Good point. Maybe with hacky way but to keep the guard, there won't be harm in leaving the event attached even after SearchView is loaded.

prathameshnarkhede:
> Is this issue only present in .NET MAUI? Will WPF/WinUI/UWP not have the same issue?

I tested the behavior with WinUI, WPF and and UWP and seems that it is working as expected. It is only bug with MAUI.

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -615,7 +615,7 @@ public string? RepeatSearchButtonText
         /// Identifies the <see cref="EnableDefaultWorldGeocoder"/> dependency property.
         /// </summary>
         public static readonly DependencyProperty EnableDefaultWorldGeocoderProperty =
-            DependencyProperty.Register(nameof(EnableDefaultWorldGeocoder), typeof(bool), typeof(SearchView), new PropertyMetadata(false, OnEnableDefualtWorldGeocoderPropertyChanged));
```
dotMorten, line 618: This is a breaking behavior. Please undo

prathameshnarkhede: Since default value was set by [`<Setter Property="EnableDefaultWorldGeocoder" Value="True" />`](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/655/files#diff-406072ed33bfc537f9f3c747a46f8c8928996a4eb73b06a9df6548ba67ddb5f2L572) and it's removed in this PR.
Given that, is it still considered a breaking change?

mstefarov (approved):
I was able to set EnableDefaultWorldGeocoder to "false" in XAML or in code, and the control honored it by not sending any arcgis.com requests 👍

src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
```diff
@@ -81,6 +81,16 @@ public SearchView()
         ClearCommand = new DelegateCommand(HandleClearSearchCommand);
         SearchCommand = new DelegateCommand(HandleSearchCommand);
         RepeatSearchHereCommand = new DelegateCommand(HandleRepeatSearchHereCommand);
+        Loaded += SearchView_Loaded;
+    }
+
+    private void SearchView_Loaded(object? sender, EventArgs e)
```
dotMorten, line 89: This event can fire many times. Do we want to run this each time? Is there a guard in the code to ensure we don't do it if it's already configured?

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
dotMorten: Technically is yet. Probably not a big deal, but why is this change needed?

prathameshnarkhede: The change does not affect anything as long as the Toolkit package is used as-is, but it may have an impact if the default behavior is overridden by merged resource dictionaries and dynamic resource references in the app. Since this aligns with how MAUI sets the default value for the property, I thought it made sense to make this change.

src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
prathameshnarkhede: I checked it is only running once when SearchView is initialized and loaded in to view. But there is still [EnableDefaultWorldGeocoder](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/a1bc4f8149c13c6aea65e86c94a09a6168aace58/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs?plain=1#L328C1-L331C10) check in ConfigureForCurrentConfiguration() method. Also, when [adding Default source](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/a1bc4f8149c13c6aea65e86c94a09a6168aace58/src/Toolkit/Toolkit/UI/Controls/SearchView/SearchViewModel.cs?plain=1#L547C1-L551C10) to Sources, it clears out all previous sources.

prathameshnarkhede: But for extra safety, I am adding a flag to loaded and unloaded events.

src/Toolkit/Toolkit/UI/Controls/SearchView/LocatorSearchSource.cs
```diff
@@ -146,7 +146,19 @@ public LocatorSearchSource(LocatorTask locator)
...
+        }
+
+        private void RefreshDisplayName()
+        {
+            if (DisplayName != Locator?.LocatorInfo?.Name && !string.IsNullOrWhiteSpace(Locator?.LocatorInfo?.Name))
+            {
+                DisplayName = Locator?.LocatorInfo?.Name ?? "Locator";
+            }
+
+            GeocodeParameters.ResultAttributeNames.Add("*");
```
dotMorten: Shouldn't this need to be called again once the load task completes and then raise property changed for the displayname property? Also "Locator" should be localized, but it's probably better to leave it empty, and then update it once the task completes.
I would think that hitting the getter for DisplayName should also initiate the load and return empty (and then raise INPC once it completes).

src/Toolkit/Toolkit/UI/Controls/SearchView/LocatorSearchSource.cs
```diff
@@ -169,15 +181,8 @@ private async Task EnsureLoaded()
                 pinSymbol.OffsetY = 16.5;
                 DefaultSymbol = pinSymbol;
             }
-
-            if (DisplayName != Locator?.LocatorInfo?.Name && !string.IsNullOrWhiteSpace(Locator?.LocatorInfo?.Name))
-            {
-                DisplayName = Locator?.LocatorInfo?.Name ?? "Locator";
             }
```
dotMorten: It makes sense that this stays, but needs to raise INPC for DisplayName

src/Toolkit/Toolkit/UI/Controls/SearchView/LocatorSearchSource.cs
```diff
@@ -146,12 +171,36 @@ public LocatorSearchSource(LocatorTask locator)
...
+
+            if (Locator?.LocatorInfo is LocatorInfo info)
+            {
+                // Locators from online services have descriptions but not names.
+                if (!string.IsNullOrWhiteSpace(info.Name) && info.Name != Locator.Uri?.ToString())
+                    _displayName = info.Name;
+                else if (!string.IsNullOrWhiteSpace(info.Description))
+                    _displayName = info.Description;
+                else
+                    _displayName = string.Empty;
```
dotMorten: Do we want it to fall back to empty? Wasn't the old code falling back to "Locator" ?

dotMorten (approved):
Looks good, except the default fallback to string.empty seem wrong?

src/Toolkit/Toolkit/UI/Controls/SearchView/LocatorSearchSource.cs
prathameshnarkhede: Ah. Maybe I misread [this](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/655#discussion_r2153224440) comment.

PR #657, opened 2025-06-09
Accessibility changes for ScaleLine control

Samhitha-Burugupalli:
- Set the value of IsTabSop and Focusable to False by default
- Added Automation name, help text for the scaleline control
- Updated the "custom" control type to "Scale"
- Exposed the Value of Scale Line for the automation tools  -- 1 to (roundedValue of MapScale)
- Added programmatic control type of ScaleLine as Text. So that the narrator knows to read the value of the control
- Set AutomationProperties.LiveSetting to Polite on the control, so that screen readers (like Narrator) will announce the scale value changes automatically in a polite, non-interruptive manner. This is useful when the control is configured with IsTabStop = true or Focusable = true and if the user has logic that announces the scale whenever it changes — such as when the MapScale property is updated programmatically.

### DIFF
--- a/src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
@@ -160,50 +160,6 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         {
             Visibility = isVisible ? Visibility.Visible : Visibility.Collapsed;
         }
-#if WPF || WINDOWS_XAML
-        /// <inheritdoc />
-        public partial class ScaleLineAutomationPeer : FrameworkElementAutomationPeer, IValueProvider
-        {
-            private readonly ScaleLine _owner;
-
-            internal ScaleLineAutomationPeer(ScaleLine owner) : base(owner)
-            {
-                _owner = owner;
-            }
-            /// <inheritdoc />
-            protected override string GetLocalizedControlTypeCore() => Properties.Resources.GetString("ScaleLineAutomationTypeName")!;
-            /// <inheritdoc />
-            protected override AutomationControlType GetAutomationControlTypeCore() => AutomationControlType.Text;
-
-            /// <inheritdoc />
-            protected override AutomationLiveSetting GetLiveSettingCore() => AutomationLiveSetting.Polite;
-
-            /// <inheritdoc />
-        #if WPF
-            public override object GetPattern(PatternInterface patternInterface)
-        #elif WINDOWS_XAML
-            protected override object GetPatternCore(PatternInterface patternInterface)
-        #endif           
-            {
-                if (patternInterface == PatternInterface.Value)
-                {
-                    return this;
-                }
-            #if WPF
-                return base.GetPattern(patternInterface);
-            #elif WINDOWS_XAML
-                return base.GetPatternCore(patternInterface);
-            #endif
-            }
-            /// <inheritdoc />
-            public string Value => String.Format(Properties.Resources.GetString("ScaleLineAutomationValue")!, GetRoundedValue(_owner.MapScale));
-            
-            /// <inheritdoc />
-            public bool IsReadOnly => true;
-            /// <inheritdoc />
-            public void SetValue(string value) => throw new System.NotSupportedException("ScaleLine value is read-only.");
-        }
-#endif
     }
 }
 #endif
--- a/src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
@@ -178,9 +178,12 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             /// <inheritdoc />
             protected override AutomationLiveSetting GetLiveSettingCore() => AutomationLiveSetting.Polite;
 
-        #if WPF
             /// <inheritdoc />
+        #if WPF
             public override object GetPattern(PatternInterface patternInterface)
+        #elif WINDOWS_XAML
+            protected override object GetPatternCore(PatternInterface patternInterface)
+        #endif           
             {
                 if (patternInterface == PatternInterface.Value)
                 {
@@ -188,17 +191,6 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                 }
                 return base.GetPattern(patternInterface);
             }
-        #else
-            /// <inheritdoc />
-            protected override object GetPatternCore(PatternInterface patternInterface)
-            {
-                if (patternInterface == PatternInterface.Value)
-                {
-                    return this;
-                }
-                return base.GetPatternCore(patternInterface);
-            }
-        #endif
             /// <inheritdoc />
             public string Value
             {
--- a/src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
@@ -16,6 +16,11 @@
 
 #if WPF || WINDOWS_XAML
 using Esri.ArcGISRuntime.UI.Controls;
+#if WPF
+using System.Windows.Automation;
+using System.Windows.Automation.Peers;
+using System.Windows.Automation.Provider;
+#endif
 #if NETFX_CORE
 using Windows.UI.Xaml.Shapes;
 #elif WINUI
@@ -55,7 +60,12 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             _metricScaleLine = GetTemplateChild("MetricScaleLine") as Rectangle;
             Refresh();
         }
-
+#if WPF
+        protected override AutomationPeer OnCreateAutomationPeer()
+        {
+            return new ScaleLineAutomationPeer(this);
+        }
+#endif
         /// <summary>
         /// Gets or sets the platform-specific implementation of the <see cref="MapScale"/> property.
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
```diff
@@ -55,7 +60,12 @@ public override void OnApplyTemplate()
             _metricScaleLine = GetTemplateChild("MetricScaleLine") as Rectangle;
             Refresh();
         }
-
+#if WPF
```
dotMorten: Would this be changed to `WPF || WINDOWS_XAML` since the same thing applies to WinUI/UWP?

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
```diff
@@ -141,6 +151,50 @@ private void SetVisibility(bool isVisible)
         {
             Visibility = isVisible ? Visibility.Visible : Visibility.Collapsed;
         }
+#if WPF
+        private class ScaleLineAutomationPeer : FrameworkElementAutomationPeer, IValueProvider
```
dotMorten: Should this be private? Generally automation peers are exposed for extensibility and the scaleline is already extensible.

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
```diff
@@ -141,6 +151,50 @@ private void SetVisibility(bool isVisible)
...
+                    return this;
+                }
+                return base.GetPattern(patternInterface);
+            }
+
+            public string Value
+            {
+                get
+                {
+                    return $"1 to {GetRoundedValue(_owner.MapScale).ToString()}";
```
dotMorten: This should be localized. You'd use string.format for this. Ie
return `String.Format(localizedstring, GetRoundedValue(_owner.MapScale))` where `localizedstring` would be `"1 to {0}"`.

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
```diff
@@ -55,7 +60,12 @@ public override void OnApplyTemplate()
             _metricScaleLine = GetTemplateChild("MetricScaleLine") as Rectangle;
             Refresh();
         }
-
+#if WPF
+        protected override AutomationPeer OnCreateAutomationPeer()
+        {
+            return new ScaleLineAutomationPeer(this);
```
dotMorten, line 75: I think you're missing a call in ScaleLine to announce the value changes on the automation peer when the scale changes. Something like
`this?._automationPeer.RaisePropertyChangedEvent(ValuePatternIdentifiers.ValueProperty, oldValue, newValue)`
This would then also mean you'd need to cache the created automation peer here for later event notification. ie something like `return _automationPeer ?? _automationPeer = new ScaleLineAutomationPeer(this);`

dotMorten (commented):
I think you're missing a call in ScaleLine to announce the value changes. Something like
`this?._automationPeer.RaisePropertyChangedEvent(ValuePatternIdentifiers.ValueProperty, oldValue, newValue)`

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
Samhitha-Burugupalli: Updated the value to be localized

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
Samhitha-Burugupalli: Thanks for pointing that out! I updated the class to public accordingly.

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
Samhitha-Burugupalli: Extended the accessibility to WINUI and UWP. GetPattern is being implemented differently in UWP/WINUI and WPF. So added that logic as well.

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
Samhitha-Burugupalli: I thought, not all users might need the announcement whenever Scale changes. So, I did not consider this change at the control level. Verified that this can be done at the application level, based on MapScaleProperty changes.

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
```diff
@@ -141,6 +160,65 @@ private void SetVisibility(bool isVisible)
         {
             Visibility = isVisible ? Visibility.Visible : Visibility.Collapsed;
         }
+#if WPF || WINDOWS_XAML
+        /// <inheritdoc />
+        public partial class ScaleLineAutomationPeer : FrameworkElementAutomationPeer, IValueProvider
```
dotMorten: We generally avoid public classes inside other classes. Can you move this outside the ScaleLine class?

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
```diff
@@ -141,6 +160,65 @@ private void SetVisibility(bool isVisible)
...
+            {
+                _owner = owner;
+            }
+            /// <inheritdoc />
+            protected override string GetLocalizedControlTypeCore() => Properties.Resources.GetString("ScaleLineAutomationTypeName")!;
+            /// <inheritdoc />
+            protected override AutomationControlType GetAutomationControlTypeCore()
+            {
+                return AutomationControlType.Text;
+            }
```
dotMorten: Nitpick: Before and after this you use the short-hand `=>` syntax, but you don't here. For consistency consider also doing that here. 

```suggestion
            protected override AutomationControlType GetAutomationControlTypeCore() => AutomationControlType.Text;
```

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.Windows.cs
```diff
@@ -141,6 +160,65 @@ private void SetVisibility(bool isVisible)
...
+            /// <inheritdoc />
+            protected override object GetPatternCore(PatternInterface patternInterface)
+            {
+                if (patternInterface == PatternInterface.Value)
+                {
+                    return this;
+                }
+                return base.GetPatternCore(patternInterface);
+            }
+        #endif
```
dotMorten: You would want to try and reuse implementation if you can. Since the contents are the same and only the signature, you can change it to the below. Also it's better to use `#elif` than `#else` to be explicit about the other platforms the code addresses, in case we ever add more platforms in the future.

```suggestion
            /// <inheritdoc />
        #if WPF
            public override object GetPattern(PatternInterface patternInterface)
        #elif WINDOWS_XAML
            protected override object GetPatternCore(PatternInterface patternInterface)
        #endif           
            {
                if (patternInterface == PatternInterface.Value)
                {
                    return this;
                }
                return base.GetPattern(patternInterface);
            }
```

src/Toolkit/Toolkit.UI.Controls/ScaleLine/ScaleLine.cs
```diff
@@ -171,6 +171,26 @@ private static double GetRoundedValue(double value)
...
+                return value - (value % 1000000);
+            }
+            else if (value >= 100000)
+            {
+                return value - (value % 100000);
+            }
+            else if (value >= 10000)
+            {
+                return value - (value % 10000);
+            }
```
dotMorten: What was the reason for this change? I believe this was just to do a little rounding on the scaleline numbers to be 1000s for large numbers, but leave it at that. This change would be a behavioral change, and feels a little out of scope for a11y work, unless I'm missing something here.

Samhitha-Burugupalli: I used the existing  GetRoundedValue method to roundoff the MapScale Property (which is like 11151234.5) 
If I do not add the additional rounding logic, the Value which would be exposed to the user in the accessibility tools will look something like "1 to 11151000" and read out by the narrator as "one to Eleven million One Hundred Fifty-One Thousand" which I felt can be rounded off further. Anyways, as I will be moving the ScaleLineAutomationPeer class to a new file now, I can remove the logic which I added in ScaleLine.cs and add a new GetRoundedValue method with this additional rounding or I can leave it as it is if rounding like "1 to 11151000" this works.

Samhitha-Burugupalli: I wanted to correct an earlier point I made and also clarify the intent of the rounding logic.

Previously, I mentioned that screen reader would read a value like 11151234.5 as "eleven million one hundred fifty-one thousand", but that was incorrect. In reality, it's read out digit-by-digit as "one to one one one five one zero zero zero" which is not narrator friendly. 

To address this, rounding to 1 million is essential - it ensures that value will be rounded to 11000000 and read as "Eleven Million", which is much clearer and more usable by narrator users

I agree that additional rounding to 10 thousand, 100 thousand, 10 million and 100 million is not mandatory for this context and I can remove that part while keeping 1 million thresholds in place.

mstefarov: @dotMorten The original GetRoundedMethod was only used on rather small numbers ("km" and "mi" labels, so at most tens of thousands), so there was never a need for larger thresholds.  But the automation peers is dealing with much larger numbers because it's reading out map scale, where the denominator could be in hundreds of millions.

The current code (adding a 1M threshold for the peer) looks good.  Alternatively &mdash; perhaps I'm overcomplicating it &mdash; we could round to a couple significant figures to balance precision and readability, [for example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7NAExAGoAfAAQCYBGAWAChGAFKASzwwAoAlCfQgBUIAZTYBzPGwBmbMAENOAMQk5YAZy41tAVhpUAzCgB0OgJRmA3C3ade/PENESpshctUauADhQBOAxoDKmM/AHYLawZWDm4+AWExSRk5RQwVcTUYTT8qP28jSJtY+wTnZLc0jKzNHUDDIujbOIcnJNdUj0yvbyRCq2K7eMdElxT3dM9srQaBppLhtrHKrpquAAZGxgoaJAACQn5gABsYPcXRis7J7unDnBOzgDc0Pdi99QlZcXU9gF49lQzIwAN6MPYQt7SPZce6PYwASXUADl5MiuE8zHsyGQ9k89gAePY0YEMSHkvYUMLEqLgyEAenpe1kjjx8mOOBgAHJfngYPINBg9gAHCAAdxgUD2EGhNHWr3UAAsZBgYIQ9hgIHsANYwGDCvbyZkwMUHCRsDC/YAwaTQM4YRVnQgwMBsAC27I++oF8k1UGMdIhcNOew9yQwOGd/z2AFlfYrjEpjhBoFw4w7jAAZCDiOUYxrk4NndQKEMA9MJuMILSvCvGZjirTy0PycORs5wGGfcTfX6dkkFyGBvaMvZ4aAe45sABeMFeUFarwdMDwe0Fdo+pZgw6psfjxkWGJHm/Zc72zZjbEIotiiw44mMAEExfIAJ5KBdugBakogWIAVCepxRAAvow2y7JSKB7DEdhFmOOButaUCkmCZKQjsfhcAAJAARCCeCIchIDIusIG4XMIFAA==):

```cs
// Rounds a positive double to an integer‐valued double with desired precision.
static double RoundToSignificantFigures(double v, int sigfigs = 2)
{
    if (double.IsNaN(v) || v < 1)
        return 1;

    // find value's nearest power of 10, shifted to keep a few digits before the decimal separator.
    double magnitude = Math.Floor(Math.Log10(v));
    double scale = Math.Max(1, Math.Pow(10, magnitude - (sigfigs - 1)));
    
    // normalize, round, then restore scale
    return Math.Round(v / scale, 0, MidpointRounding.AwayFromZero) * scale;
}
```

But I am honestly fine with the current approach too.

mstefarov (approved):

PR #659, opened 2025-06-10
Utility associations form elements

dotMorten:
https://github.com/user-attachments/assets/66cef671-f803-4585-98a3-2548c60ce08f


src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
```diff
@@ -450,6 +418,155 @@ internal void OnHyperlinkClicked(Uri uri)
...
+            {
+                _elementPropertyChangedListener = new WeakEventListener<FeatureFormView, INotifyPropertyChanged, object?, PropertyChangedEventArgs>(this, inpcNew)
+                {
+                    OnEventAction = static (instance, source, eventArgs) => instance.FeatureForm_PropertyChanged(source, eventArgs),
+                    OnDetachAction = static (instance, source, weakEventListener) => source.PropertyChanged -= weakEventListener.OnEvent,
+                };
+                inpcNew.PropertyChanged += _elementPropertyChangedListener.OnEvent;
+            }
+            UpdateIsValidProperty();
+        }
```
prathameshnarkhede, line 569: Can the conditions be simplified here?
Maybe with just single `if (newForm is INotifyPropertyChanged inpcNew)` and `if (oldForm is not null)`

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationsFormElementView.cs
```diff
@@ -0,0 +1,147 @@
...
+                    OnEventAction = static (instance, source, eventArgs) => instance.UtilityAssociationsFormElement_PropertyChanged(source, eventArgs),
+                    OnDetachAction = static (instance, source, weakEventListener) => source.PropertyChanged -= weakEventListener.OnEvent,
+                };
+                inpcNew.PropertyChanged += _elementPropertyChangedListener.OnEvent;
+                RefreshAssociations();
+                UpdateVisibility();
+            }
+        }
+
+        private async void RefreshAssociations()
```
prathameshnarkhede, line 117: Can this become `async Task` and just called as `_ = RefreshAssociations();`?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
dotMorten: no that would keep subscribing to the old feature.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationsFormElementView.cs
dotMorten: That just moves the problem up to an unobserved exception higher up the stack. It's handled safely here with the try/catch.

prathameshnarkhede (approved):

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
```diff
@@ -533,4 +533,36 @@
...
+  <data name="FeatureFormPendingEditsTitle" xml:space="preserve">
+    <value>Unsaved edits</value>
+    <comment>The title of the dialog when user is warned about unsaved edits</comment>
+  </data>
+  <data name="FeatureFormUtilityElementFractionAlongEdge" xml:space="preserve">
+    <value>Fraction along edge: {0}</value>
+    <comment>The percentage fraction along an edge of a utility element association</comment>
+  </data>
+  <data name="FeatureFormUtilityElementIsContentVisible" xml:space="preserve">
+    <value>Is content visible: {0}</value>
```
mstefarov: Are we supposed to localize the boolean value too? In .NET the strings "true" and "false" are [hardcoded](https://learn.microsoft.com/en-us/dotnet/api/system.boolean.tostring?view=netstandard-2.0#system-boolean-tostring(system-iformatprovider)) with no localization.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
```diff
@@ -450,6 +418,155 @@ internal void OnHyperlinkClicked(Uri uri)
...
+                try
+                {
+                    var result = await dialog.ShowAsync();
+                    if (result == ContentDialogResult.Primary)
+                    {
+                        await FinishEditingAsync();
+                       }
+                    else if (result == ContentDialogResult.Secondary)
+                    {
+                        _ = DiscardEditsAsync();
```
mstefarov: Why no await DiscardEditsAsync? Are you okay with it invoking FetchAttachmentsAsync, EvaluateExpressions, and ResetValidationStates _after_ you proceed with this method?

dotMorten: Good call. We don't really need the async method that does all that. We can just call the synchronous `CurrentFeatureForm?.DiscardEdits();` since we're leaving this page anyway

PR #660, opened 2025-06-13
BugFix: Fixing BasemapGallery Sceneview appears blank sometimes.

prathameshnarkhede:
Introduced a semaphore for thread-safe access to UpdateBasemaps.
This avoids multiple invocations of UpdateBasemaps() and NotifySpatialReferenceChanged() and locks resource when in use by another call.

Did testing by opening SceneView multipel times and did not observed the bug.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
@@ -201,27 +201,29 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
 
         public async Task UpdateBasemaps()
         {
+            // Cancel any pending load before starting a new one
+            _loadCancellationTokenSource?.Cancel();
+
             await _updateBasemapsSemaphore.WaitAsync();
             try
             {
-            IsLoading = true;
-            // Cancel any pending load before starting a new one
-            _loadCancellationTokenSource?.Cancel();
-            _loadCancellationTokenSource = new CancellationTokenSource();
-            try
-            {
-                _availableBasemaps = await PopulateBasemaps(_loadCancellationTokenSource.Token);
-                HandleAvailableBasemapsChanged();
+                IsLoading = true;
+                _loadCancellationTokenSource = new CancellationTokenSource();
+                try
+                {
+                    _availableBasemaps = await PopulateBasemaps(_loadCancellationTokenSource.Token);
+                    HandleAvailableBasemapsChanged();
+                }
+                catch (OperationCanceledException) { }
+                catch (Exception ex)
+                {
+                    System.Diagnostics.Trace.WriteLine(ex.Message);
+                }
+                finally
+                {
+                    IsLoading = false;
+                }
             }
-            catch (Exception ex)
-            {
-                System.Diagnostics.Trace.WriteLine(ex.Message);
-            }
-            finally
-            {
-                IsLoading = false;
-            }
-        }
             finally
             {
                 _updateBasemapsSemaphore.Release();
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
@@ -301,13 +301,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
 
         private Task<ObservableCollection<BasemapGalleryItem>> LoadBasemapGalleryItems(ArcGISPortal portal, CancellationToken cancellationToken = default)
         {
-            if (_loadBasemapGalleryItemsTask is null || _loadBasemapGalleryItemsTask.IsCompleted)
-            {
-                _loadBasemapGalleryItemsTask = LoadBasemapGalleryItemsInternal(portal, cancellationToken);
-                return _loadBasemapGalleryItemsTask;
-            }
-
-            return _loadBasemapGalleryItemsTask;
+            return LoadBasemapGalleryItemsInternal(portal, cancellationToken);
         }
 
         private async Task<ObservableCollection<BasemapGalleryItem>> LoadBasemapGalleryItemsInternal(ArcGISPortal portal, CancellationToken cancellationToken = default)
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
@@ -299,12 +299,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
             return await LoadBasemapGalleryItems(Portal, cancellationToken);
         }
 
-        private Task<ObservableCollection<BasemapGalleryItem>> LoadBasemapGalleryItems(ArcGISPortal portal, CancellationToken cancellationToken = default)
-        {
-            return LoadBasemapGalleryItemsInternal(portal, cancellationToken);
-        }
-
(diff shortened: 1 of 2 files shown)


src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
```diff
@@ -197,8 +197,13 @@ private async Task HandlePortalChanged()
             await UpdateBasemaps();
         }
 
+        private readonly SemaphoreSlim _updateBasemapsSemaphore = new(1, 1);
+
         public async Task UpdateBasemaps()
         {
+            await _updateBasemapsSemaphore.WaitAsync();
```
dotMorten: Wouldn't you want to cancel the current populate operation by calling `_loadCancellationTokenSource.Cancel()` here, so you can get the semaphore released sooner?
(if so, also make a special case for OperationCanceledException below that doesn't write to the Trace)

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
```diff
@@ -197,24 +197,36 @@ private async Task HandlePortalChanged()
...
+        private readonly SemaphoreSlim _updateBasemapsSemaphore = new(1, 1);
+
         public async Task UpdateBasemaps()
         {
-            IsLoading = true;
             // Cancel any pending load before starting a new one
             _loadCancellationTokenSource?.Cancel();
-            _loadCancellationTokenSource = new CancellationTokenSource();
+
+            await _updateBasemapsSemaphore.WaitAsync();
```
mstefarov: If there are multiple rapid calls to UpdateBasemaps, they will all get queued up here.  That means PopulateBasemaps will eventually run the same number of times that UpdateBasemaps was called.

This might be okay if UpdateBasemaps doesn't get called in bursts &mdash; and it's still an improvement over current behavior.  But if you don't want updates to queue up, you might need...

<details><summary>something more complicated, e.g.</summary>

```cs
private Task? _updateTask;
private bool _pending;
private CancellationTokenSource? _cts;

// Assume this is always called on the UI thread -- I think it is!
public Task UpdateBasemaps()
{
    // If nobody's running, kick off the update task
    if (_updateTask is null || _updateTask.IsCompleted)
    {
        _pending = true;
        _updateTask = EnsureBasemapsAreUpdatedAsync();
    }
    else
    {
        // Otherwise, make sure PopulateBasemaps will run again
        _pending = true;
        _cts?.Cancel();
    }

    // Every caller can await the same Task
    return _updateTask;
}

// No extra thread-safety needed here because task stays on the UI thread
private async Task EnsureBasemapsAreUpdatedAsync()
{
    do
    {
        // consume the pending flag
        _pending = false;

        // replace the CTS
        _cts?.Dispose();
        _cts = new CancellationTokenSource();
        var token = _cts.Token;

        try
        {
            IsLoading = true;
            // no ConfigureAwait(false) here! We do want to continue on UI thread.
            _availableBasemaps = await PopulateBasemaps(token);
            HandleAvailableBasemapsChanged();
        }
        catch (OperationCanceledException)
        {
            // will loop again if _pending
        }
        catch (Exception ex)
        {
            System.Diagnostics.Debug.WriteLine(ex);
        }
        finally
        {
            IsLoading = false;
        }

        // loop again if someone called UpdateBasemaps while we were running
    } while (_pending);

    // once no more pending, exit to complete _updateTask
}
```
</details>

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
```diff
@@ -197,24 +197,36 @@ private async Task HandlePortalChanged()
...
-                HandleAvailableBasemapsChanged();
-            }
-            catch (Exception ex)
-            {
-                System.Diagnostics.Trace.WriteLine(ex.Message);
+                IsLoading = true;
+                _loadCancellationTokenSource = new CancellationTokenSource();
+                try
+                {
+                    _availableBasemaps = await PopulateBasemaps(_loadCancellationTokenSource.Token);
```
mstefarov: Since we are now serializing calls to PopulateBasemaps, does this make the task-gating code in LoadBasemapGalleryItems(...) redundant?  Perhaps that can be simplified.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
dotMorten: @mstefarov My impression is that the cancel call right before will stop this load immediately, and start a fresh run

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
```diff
@@ -197,24 +197,39 @@ private async Task HandlePortalChanged()
...
+            // Try to enter the semaphore without waiting
+            if (!await _updateBasemapsSemaphore.WaitAsync(0))
             {
-                _availableBasemaps = await PopulateBasemaps(_loadCancellationTokenSource.Token);
-                HandleAvailableBasemapsChanged();
+                // Another update is already running; exit immediately
+                return;
             }
-            catch (Exception ex)
+            try
```
mstefarov: I'm still not 100% sure if this avoids a race condition.  Consider:
1. Something causes UpdateBasemaps to be called, which kicks off PopulateBasemaps
2. PopulateBasemaps inits the Portal property and calls LoadBasemapGalleryItems(oldPortal) in the background
3. Meanwhile, something else updates the Portal property and calls UpdateBasemaps. It will not call PopulateBasemaps again.
4. LoadBasemapGalleryItems will finish loading basemaps for the old portal

If this is a plausible scenario, we will end up with basemapGalleryItems that no longer match the active Portal.  I am not sure exactly what triggers UpdateBasemaps to be called multiple times -- is this realistic?

prathameshnarkhede: Okay. I think I get it now. I have added new changed and realized we can achieve it without SemaphoreSlim using pending boolean locks to make sure to collate subsequent UpdateBasemaps calls and make it a single call.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
```diff
@@ -197,24 +197,44 @@ private async Task HandlePortalChanged()
...
+                        HandleAvailableBasemapsChanged();
+                    }
+                    catch (OperationCanceledException) { }
+                    catch (Exception ex)
+                    {
+                        System.Diagnostics.Trace.WriteLine(ex.Message);
+                    }
+                    finally
+                    {
+                        IsLoading = false;
```
dotMorten: Idea: If you set _loadCancellationTokenSource to null here, whether you have a cancellation source or not defines whether _isUpdatingBasemaps is true or false, so you won't need that variable.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryController.cs
```diff
@@ -197,24 +197,44 @@ private async Task HandlePortalChanged()
             await UpdateBasemaps();
         }
 
+        private bool _pendingUpdateBasemaps;
+        private bool _isUpdatingBasemaps;
```
dotMorten: Isn't this essentially the same as "IsLoading" ?

prathameshnarkhede: I have removed `_isUpdatingBasemaps` and unified the implementation with just `IsLoading` boolean property.

mstefarov (approved):

PR #665, opened 2025-06-23
Improve feature form load time on maui

dotMorten:
Slightly improves load time of feature forms by avoiding use of a templated view.

Also huge performance improvement with successive loads of a featureformdefinition where the set of FormElements doesn't change.

### DIFF
--- a/src/Toolkit/Toolkit.Maui/ContentControl.cs
+++ b/src/Toolkit/Toolkit.Maui/ContentControl.cs
@@ -42,7 +42,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui
         }
 
         /// <summary>
-        /// Gets or sets the template selector for the header.
+        /// Gets or sets the template for the header.
         /// </summary>
         public DataTemplate? ContentTemplate
         {
--- a/src/Toolkit/Toolkit.Maui/ContentControl.cs
+++ b/src/Toolkit/Toolkit.Maui/ContentControl.cs
@@ -5,6 +5,8 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui
 {
     internal class ContentControl : ContentPresenter
     {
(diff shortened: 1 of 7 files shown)


src/Toolkit/Toolkit.Maui/ContentControl.cs
```diff
@@ -3,79 +3,69 @@
...
-        {
-            base.OnApplyTemplate();
-            UpdateData();
         }
 
         private void UpdateData()
         {
-            if (GetTemplateChild("ContentPresenter") is ContentPresenter presenter)
+            if (ContentData is null)
+                Content = null;
```
mstefarov, line 17: Does this need to reset `currentTemplate` as well?

e.g. consider the scenario where ContentData goes to null and then comes back to the same value &mdash; won't we end up with null Content because "currentTemplate" has not changed?

dotMorten: no we want to be able to reuse the current view, and currentTemplate tells us which template was used to create that view. This is one of the really big improvements to performance.

mstefarov (approved):

src/Toolkit/Toolkit.Maui/ContentControl.cs
```diff
@@ -3,79 +3,69 @@
...
+                else  if (Content is not null && Content.BindingContext != ContentData)
+                    Content.BindingContext = ContentData;
             }
         }
 
         /// <summary>
         /// Gets or sets the template selector for the header.
         /// </summary>
-        public DataTemplateSelector? ContentTemplateSelector
+        public DataTemplate? ContentTemplate
```
prathameshnarkhede: Nitpick: Should we update the comment on this ContentTemplate as well?
`Gets or sets the DataTemplate used to display the content`

prathameshnarkhede (approved):

Issue #662, opened 2025-06-16
Question Regarding:  Only show validation errors if a field got focus #642


fcjoe-gis:
During testing with the current main branch we notice that we do not see required fields when the FeatureForm loaded.  I then notice in the history [Only show validation errors if a field got focus#642](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/642/commits).  After rolling back to before this pull request I see the desired behavior has returned.
  
I am quite curious to understand the feedback you received that led to this change.  I found the previous behavior a much better user experience and in communicating this change to users they seem to agree.  Especially with a larger form the user wants to know which fields are required.  Prior to this PR this was very clear.  Could you please reconsider this change.  Or at least give a flag the that could be set so they could have it show when the form loads  

Thanks
-Joe

dotMorten:
Thank you for the feedback! The thinking here is that when editing an existing feature, it can still be considered "invalid" according to the rules but still save-able in the sense that it doesn't break with the table schema. The problem is that if someone chooses to edit a feature with values that doesn't pass the validation requirements, it is confusing to get validation errors for fields you didn't even touch. You might only be interested in changing one value, but suddenly you can't apply that change without also dealing with other failures.

Does this help explain the reasoning? I'd love to hear the workflows you got and better understand it so we can look into adding more control here.

ramtechjoe:
Specific to required fields.  A driver to using the FeatureForm is that I can define my form on the web and use that configuration on the form.  One aspect is making fields required that may allow null values.  This is critical.

I feel like this is removing an entire driver for the FeatureFormView.   A key aspect was informing users of all the required fields on the form based on the configuration.  Having the form not show required fields, now requires the user tap through every field on the form to determine if it is required.  The user does not know they filled in the form completely until they Submit and get errors

The user knows they have filled in the form when all the required fields have been addresses (maybe not the best example because almost every field is required in this case)

![Image](https://github.com/user-attachments/assets/db35917a-6cfa-4cd3-9b8a-b753230a9033)

I am not clear on the use case of someone wanting to have a form configured in the web map, but that then allows a user to save just based on the table definition ignoring the configuration.

mstefarov:
[...]

PR #666, opened 2025-06-25
FeatureForm: Expose bindable apply/discard commands

dotMorten:
- New `FinishEditingCommand` and `DiscardEditsCommand` + Auto-scroll to first error on FinishEditing, and allow developer to perform similar scrolling when using custom validation workflows. This greatly reduces the amount of code the user have to write, while getting a good default editing experience.
- Adds ability to always show errors, or only show on finishediting/focus. Also fixes: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/issues/662
- Better handle apply errors when applying edits from a sub-featureform.

### DIFF
--- a/src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
+++ b/src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
@@ -545,7 +545,11 @@
     <value>Discard</value>
     <comment>The text on the Discard button of the unsaved-edits warning dialog</comment>
   </data>
-  <data name="FeatureFormPendingEditsMessage" xml:space="preserve">
+    <data name="FeatureFormApplyEditsErrorTitle" xml:space="preserve">
+    <value>Error</value>
+    <comment>The text on the dialog shown if saving edits to a feature form fails</comment>
+  </data>
+    <data name="FeatureFormPendingEditsMessage" xml:space="preserve">
     <value>You need to apply or discard edits of the current feature before moving to another feature. Do you want to apply the changes?</value>
     <comment>The content of the dialog when user is warned about unsaved edits</comment>
   </data>
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
@@ -48,7 +48,45 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
 #else
             DefaultStyleKey = typeof(FeatureFormView);
 #endif
+            FinishEditingCommand = new Command(() => _ = FinishEditingAsync(), () => IsValid && CurrentFeatureForm?.HasEdits == true);
+            DiscardEditsCommand = new Command(() => _ = DiscardEditsAsync(), () => CurrentFeatureForm?.HasEdits == true);
         }
+
+        private class Command : System.Windows.Input.ICommand
+        {
+            private Action _execute;
+            private Func<bool> _canExecute;
+
+            public Command(Action execute, Func<bool> canExecute)
+            {
+                _execute = execute;
+                _canExecute = canExecute;
+            }
+            internal void RaiseCanExecuteChanged() => CanExecuteChanged?.Invoke(this, EventArgs.Empty);
+
+            public event EventHandler? CanExecuteChanged;
+
+            public bool CanExecute(object? parameter) => _canExecute();
+
+            public void Execute(object? parameter) => _execute();
+        }
+
+        /// <summary>
+        /// Command for calling <see cref="FinishEditingAsync"/> and applying edits to the currently active Feature Form if all fields are valid.
+        /// </summary>
+        /// <seealso cref="FinishEditingAsync"/>
+        /// <seealso cref="IsValid"/>
+        /// <seealso cref="FeatureForm.HasEdits"/>
+        /// <seealso cref="DiscardEditsCommand"/>
+        public System.Windows.Input.ICommand FinishEditingCommand { get; }
+
+        /// <summary>
+        /// Discards any pending edits to the currently active Feature Form, if any edits have been made
+        /// </summary>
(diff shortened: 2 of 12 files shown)


src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
```diff
@@ -525,6 +734,32 @@ private async void SubView_OnNavigating(object? sender, NavigationSubView.Naviga
...
+        private void ShowError(string title, string content)
+#else
+        private async Task ShowErrorAsync(string title, string content)
+#endif
+        {
+#if MAUI
+            var page = GetParent<Page>(this);
+            if(page is not null) {
+                await page.DisplayAlert(title, content, Properties.Resources.GetString("FeatureFormPendingEditsCancel")!);
+            }
```
mstefarov: Nitpick: format this

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
```diff
@@ -545,7 +545,11 @@
     <value>Discard</value>
     <comment>The text on the Discard button of the unsaved-edits warning dialog</comment>
   </data>
-  <data name="FeatureFormPendingEditsMessage" xml:space="preserve">
+    <data name="FeatureFormApplyEditsErrorTitle" xml:space="preserve">
+    <value>Error</value>
```
mstefarov, line 549: Is this title too generic? Maybe something more informative like "Error applying edits" or "Could not apply edits" would be more helpful?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
```diff
@@ -48,7 +48,60 @@ public FeatureFormView()
 #else
             DefaultStyleKey = typeof(FeatureFormView);
 #endif
+
+            DiscardEditsCommand = new Command(() => _ = DiscardEditsAsync(), () => CurrentFeatureForm?.HasEdits == true);
```
mstefarov, line 52: Running DiscardEditsAsync as fire-and-forget task does not seem safe.  It does not guarantee against throwing.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
```diff
@@ -525,6 +734,32 @@ private async void SubView_OnNavigating(object? sender, NavigationSubView.Naviga
...
+            var dialog = new ContentDialog
+            {
+                Title = title,
+                Content = content,
+                CloseButtonText = Properties.Resources.GetString("FeatureFormPendingEditsCancel")!
+            };
+            dialog.XamlRoot = this.XamlRoot;
+            var result = await dialog.ShowAsync();
+#elif WPF
+            MessageBox.Show(title, content);
```
mstefarov: The two parameters should be swapped -- WPF MessageBox takes [content first, title second](https://learn.microsoft.com/en-us/dotnet/api/system.windows.messagebox.show?view=windowsdesktop-9.0#system-windows-messagebox-show(system-string-system-string)).
```suggestion
            MessageBox.Show(content, title);
```

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
mstefarov: And one more thing: did you mean to remove all the "Are you sure?" dialogs when discarding in the samples?

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
dotMorten: The details is in the message though. This is just the title bar.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
dotMorten: > did you mean to remove all the "Are you sure?" dialogs when discarding in the samples?

Yes. Wanted the sample to focus on the simple scenario and make sure the code is tested for the command. You can still do the manual checks if you want.

dotMorten: @mstefarov discarded tasks don't throw: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards#a-standalone-discard
The discard operation itself is synchronous. It is the evaluate/fetch-attachments part that technically could throw, but that is already handled the same way on initial load of the form as well.

mstefarov: Ah, right!  For a moment I was thinking about unobserved exceptions [as they were in .NET 2.0-4.0](https://devblogs.microsoft.com/dotnet/tasks-and-unhandled-exceptions/) instead of [how they are in .NET 4.5+](https://devblogs.microsoft.com/dotnet/task-exception-handling-in-net-4-5/).  Old habit, sorry!

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FeatureFormView.cs
```diff
@@ -48,7 +48,60 @@ public FeatureFormView()
...
+#if WPF
+                    ShowError(Properties.Resources.GetString("FeatureFormApplyEditsErrorTitle")!, ex.Message);
+#else
+                    await ShowErrorAsync(Properties.Resources.GetString("FeatureFormApplyEditsErrorTitle")!, ex.Message);
+#endif
+                }
+            }, () => CurrentFeatureForm?.HasEdits == true);
+        }
+
+        private class Command : System.Windows.Input.ICommand
```
prathameshnarkhede, line 70: Nitpick: Can this command class be renamed to something like RelayCommand or DelegateCommand or FeatureFormCommand? It would avoid confusion with Command property of the controls.

PR #670, opened 2025-07-11
Use IdentifyGeometryEditorAsync in GeoViewController for MVVM workflows

prathameshnarkhede:
Added `MapViewController` class that inherits from `GeoViewController` to accomodate MapView's `IdentifyGeometryEditorAsync` method.

### DIFF
--- a/src/Toolkit/Toolkit/MapViewController.cs
+++ b/src/Toolkit/Toolkit/MapViewController.cs
@@ -17,7 +17,8 @@
 #nullable enable
 using Esri.ArcGISRuntime.Data;
 using Esri.ArcGISRuntime.Geometry;
-#if WINUI 
+using Esri.ArcGISRuntime.UI;
+#if WINUI
 using Point = Windows.Foundation.Point;
 #elif WINDOWS_UWP
 using Point = Windows.Foundation.Point;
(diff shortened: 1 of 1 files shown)


duffh (approved):
👍

src/Toolkit/Toolkit/MapViewController.cs
```diff
@@ -0,0 +1,51 @@
...
+        #region Identify
+
+        /// <inheritdoc cref="MapView.IdentifyGeometryEditorAsync(Point, double)" />
+        public virtual async Task<IdentifyGeometryEditorResult?> IdentifyGeometryEditorAsync(Point screenPoint, double tolerance) =>
+            ConnectedMapView is null
+                ? null
+                : await ConnectedMapView.IdentifyGeometryEditorAsync(screenPoint, tolerance).ConfigureAwait(false);
+
+        #endregion
+    }
```
dotMorten, line 215: This needs all the MapView specific methods and not just the editor-specific one, otherwise introducing this class doesn't make sense.

prathameshnarkhede: Ah. You're right. I’ve now added more MapView specific methods and protected methods to the class to make it more comprehensive and aligned with its intended purpose.

src/Toolkit/Toolkit/MapViewController.cs
```diff
@@ -0,0 +1,217 @@
...
+        protected new MapView? ConnectedView => base.ConnectedView as MapView;
+
+        /// <summary>
+        /// Attaches a Mapview to the controller, or detaches if <c>null</c>.
+        /// </summary>
+        /// <param name="mapView">The <see cref="MapView"/> to attach, or <c>null</c> is detaching.</param>
+        public void Attach(MapView mapView)
+        {
+            base.Attach(mapView);
+        }
```
mstefarov: I would remove this (and associated GeoViewController.cs changes).

The base class already has an Attach(GeoView) that will gladly take a MapView, there is no benefit from an extra overload.

src/Toolkit/Toolkit/MapViewController.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        /// <inheritdoc />
+        protected sealed override void OnGeoViewUnloaded(GeoView geoView)
+        {
+            if (geoView is MapView mapView)
+                OnMapViewUnloaded(mapView);
+            else
+                base.OnGeoViewUnloaded(geoView);
+        }
```
mstefarov: I wouldn't call base implementations at all here -- they are empty, and the geoView is guaranteed to be a MapView or null anyway.

mstefarov (approved):

PR #673, opened 2025-07-18
BugFix: Fix border (remove HasShadow) in UtilityNetworkTraceTool UI

gdestigter:
Removes `HasShadow` from `Border` (recently updated from `Frame`) in `UtilityNetworkTraceTool` UI.

`Border` does not contain a `HasShadow` property, and the app throws an exception when creating the page. In release mode, this results in a silent error when adding the page to the navigation stack. This PR removes the `HasShadow` property, allowing the page to be created and added to the navigation stack.


jnery (approved):
Can you also fix v.next :) Thank you, Greg!!

prathameshnarkhede (approved):

PR #676, opened 2025-07-23
NavigationSubView: Ensure content is preserved when setting content to null

dotMorten:
Ensure content is preserved when setting content to null and restoring the content later with the same datatemplate.


gdestigter (approved):
Looks good to me :+1:

mstefarov (approved):

PR #680, opened 2025-07-30
Removes UWP support

dotMorten:


src/Samples/Toolkit.SampleApp.WinUI/Toolkit.SampleApp.WinUI.csproj
mstefarov, line 1: You can also delete the unused `CopyUWPSamples` target in this file

PR #682, opened 2025-08-13
popup: Support clickable image links on MAUI, WinUI, and UWP

mstefarov:
Example map: https://www.arcgis.com/apps/mapviewer/index.html?webmap=b460e579c5914b719188afd1013e0054

The HTML renderers for MAUI and WinUI/UWP did not correctly handle image links (i.e. `<a href="..."></a>`.  Not only were they not clickable, they did not display at all.

Before fix:

2025-08-13_031056 Toolkit SampleApp WinUI

After fix:

2025-08-13_030906 Toolkit SampleApp WinUI

<hr>
While closely reading the code I also spotted a couple more bugs -- like rowSpan/colSpan mixup, and some incorrect font-style inheritance code.  I am tacking these on to this PR as passenger fixes.

### DIFF
--- a/src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
+++ b/src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
@@ -82,6 +82,8 @@ internal static class HtmlToView
         }
     }
 
+    // CreateBlock builds a view for a single *blocky* node. It is called for nodes that are
+    // blocks themselves (List/Table/Block/Divider/Image) and also for nodes that contain block descendants.
     private static View CreateBlock(MarkupNode node, EventHandler<Uri>? urlClickHandler)
     {
         // Create a view for a single block node.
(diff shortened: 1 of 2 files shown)


src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
```diff
@@ -161,6 +163,47 @@ private static View CreateBlock(MarkupNode node, EventHandler<Uri>? urlClickHand
                     imageElement.Source = imageSource;
                 return imageElement;
 
+            case MarkupType.Link:
+                // If the link wraps block content (like <img>), render it as a tappable ContentView.
+                if (Uri.TryCreate(node.Content, UriKind.Absolute, out var linkUri))
+                {
+                    View content;
+                    if (node.Children.Count == 1)
```
prathameshnarkhede, line 171: ~Would it be helpful to have null check for `node.Children`?~
I realized Children is always initialized with empty list.

prathameshnarkhede (approved):

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
mstefarov: I tried to make my life easy with 0-length collections instead of nulls :)

dotMorten (approved):

PR #685, opened 2025-08-21
Search view dropdown fix

fcjoe-gis:
Associated to #686

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -225,3 +225,6 @@ MigrationBackup/
 # Repo-specific additions
 src/Toolkit/Toolkit.UWP/LocalizedStrings/*/Resources.resw
 Output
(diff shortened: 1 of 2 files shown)


dotMorten (changes requested):
Thank you for submitting a fix! Do you mind clearing out the unrelated changes (.gitignore and whitespace changes).

ramtechjoe:
@dotMorten, it seems like Rider is doing some weird automatic formatting with the white space which I cannot get it to not do.  I tried a second time and all those changes still show up.

Perhaps you could just have one of your developers make the code changes and submit from their environments so those formatting changes don't occur

ramtechjoe:
See #691 I used notepad++

PR #687, opened 2025-08-22
Added logic so that the error message and text box outline updates are displayed based on ErrorsVisibility

ramtechjoe:
The FieldFormElementView and TextFormInputView call code to update based on if an error is found (missing Required data).  When this is initially the controls have not fully loaded and so the code to make these updates is not executed properly.

Code was added to tie in the Loaded event handler and when OnLoaded gets call the methods to update the view 
```
// FeatureFormInputView
protected override void OnApplyTemplate()
{
	base.OnApplyTemplate();
	/* Removed from bevity */
	ConfigureTextBox();
	
	//JH: Same issue as in FieldFormElement, the first time UpdateValidationState is called Element is null
	Loaded += OnLoaded;
	UpdateValidationState();
}
```

```
// FieldForElementView
protected override void OnApplyTemplate()
{
	base.OnApplyTemplate();
	if (GetTemplateChild(FieldInputName) is ContentControl content)
	{
		content.ContentTemplate = new FieldTemplateSelector(this);
	}
	
	// JH: If UpdateErrorMessages is called before the FieldFormElementView is Loaded it fails
	Loaded += OnLoaded;
	UpdateErrorMessages();
}

private void OnLoaded(object? sender, EventArgs e)
{
	UpdateErrorMessages();
	Loaded -= OnLoaded;
}
```

In UpdateValidationState where the TextInput border is set to red when the ErrorsVisibility="Visible" property is set

```
	private void UpdateValidationState()
	{
		var err = Element?.ValidationErrors;

		//JH: Add a check to see if the ErrorsVisibility is set to true, not only if _hadFocus is true
		bool shouldShowError = (FeatureFormView.GetFeatureFormViewParent(this)?.ShouldShowError() ?? false) || _hadFocus;
		
		if (err != null && err.Any() && Element?.IsEditable == true && shouldShowError)
		{
#if MAUI
  //replaced with above...    
     // if (err != null && err.Any() && Element?.IsEditable == true && _hadFocus)
```

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/FieldFormElementView.Maui.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/FieldFormElementView.Maui.cs
@@ -131,7 +131,16 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui.Primitives
             {
                 content.ContentTemplate = new FieldTemplateSelector(this);
             }
+            
+            // JH: If UpdateErrorMessages is called before the FieldFormElementView is Loaded it fails
+            Loaded += OnLoaded;
             UpdateErrorMessages();
+        }
+
+        private void OnLoaded(object? sender, EventArgs e)
+        {
+            UpdateErrorMessages();
+            Loaded -= OnLoaded;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


ramtechjoe:
Also note that this comment description seems incorrect: FeatureFromView L:156

```
        /// <summary>
        /// Gets or sets the associated PopupManager which contains popup and sketch editor.
        /// </summary>
        public ValidationErrorVisibility ErrorsVisibility
        {
            get { return (ValidationErrorVisibility)GetValue(ErrorsVisibilityProperty); }
            set { SetValue(ErrorsVisibilityProperty, value); }
        }
```

Issue #686, opened 2025-08-21
SearchView Dropdown Behavior


ramtechjoe:
In the SearchView if the dropdown menu is engaged showing the available searches

<img width="491" height="172" alt="Image" src="https://github.com/user-attachments/assets/8d8910d4-9b44-4204-92e5-c1a2f4feb447" />

If you then start to type in the search box the available searches does not collapse and the search result suggestions cannot be seen

<img width="454" height="193" alt="Image" src="https://github.com/user-attachments/assets/1b11635c-b567-46ed-a883-16f1e53eec97" />

I know this seems like a weirdly specific case, but actually a number of users brought it up

ramtechjoe:
Pull request:

[Search view dropdown fix](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/685)

PR #691, opened 2025-08-22
Add code to collapse dialog if search source dropdown visible and user start to type in search

fcjoe-gis:
#686


prathameshnarkhede (approved):
Appreciate your efforts on this!

ramtechjoe:
@prathameshnarkhede 

I was seeing some weird behavior with whitespace when I added the lines using Rider on Mac, which is my main development environment.  So I just tried to just add the code in a text editor so it would not show a ton of 'ghost' (the same lines were showing as removed and added) changes to the file.  Not sure if making the fix in Visual Studio would be better

fcjoe-gis:
Is there a way to merge this fix, if not via this PR by one of the developers.  I don't know what the issue with the white space but would be an easy fix for your team?

PR #692, opened 2025-08-22
Remove SketchEditor dependency from MeasureToolbar

prathameshnarkhede:
Replaced obsolete SketchEditor with the new GeometryEditor  in MeasureToolbar.

### DIFF
--- a/src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
@@ -16,13 +16,14 @@
 
 #if WPF || WINDOWS_XAML
 using System.Collections.ObjectModel;
+using System.ComponentModel;
 using Esri.ArcGISRuntime.Data;
 using Esri.ArcGISRuntime.Geometry;
 using Esri.ArcGISRuntime.UI;
+using Esri.ArcGISRuntime.UI.Editing;
 #if WPF
 using System.Windows.Controls.Primitives;
 #endif
-#pragma warning disable CS0618 // SketchEditor is Obsolete
 namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
 {
     /// <summary>
@@ -75,7 +76,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         private ButtonBase? _clearButton;
 
         // Used for replacing measure editors
-        private SketchEditor? _originalSketchEditor;
+        private GeometryEditor? _originalGeometryEditor;
 
         // Used for highlighting feature for measurement
         private readonly GraphicsOverlay _measureFeatureResultOverlay = new GraphicsOverlay();
@@ -112,10 +113,10 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                     Geometry.AreaUnits.SquareMillimeters,
                     Geometry.AreaUnits.SquareYards,
                 };
-            LineSketchEditor = new SketchEditor();
-            AreaSketchEditor = new SketchEditor();
-            SelectionLineSymbol = LineSketchEditor.Style?.LineSymbol;
-            SelectionFillSymbol = AreaSketchEditor.Style?.FillSymbol;
+            LineGeometryEditor = new GeometryEditor();
+            AreaGeometryEditor = new GeometryEditor();
+            SelectionLineSymbol = LineGeometryEditor.Tool.Style?.LineSymbol;
+            SelectionFillSymbol = AreaGeometryEditor.Tool.Style?.FillSymbol;
         }
 
         /// <inheritdoc/>
@@ -206,7 +207,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         /// - Only one of the measure toggle buttons is enabled
         /// - Only one of the units selector is visible
         /// - Updates instruction text
-        /// - Assigns the appropriate SketchEditor
+        /// - Assigns the appropriate GeometryEditor
         /// - Updates command to execute on clear.
         /// </summary>
         private void PrepareMeasureMode()
@@ -215,19 +216,19 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             var isMeasuringArea = _mode == MeasureToolbarMode.Area;
             var isMeasuringFeature = _mode == MeasureToolbarMode.Feature;
 
-            var sketchEditor = isMeasuringLength ?
-                LineSketchEditor :
-                (isMeasuringArea ? AreaSketchEditor : _originalSketchEditor);
+            var geometryEditor = isMeasuringLength ?
+                LineGeometryEditor :
+                (isMeasuringArea ? AreaGeometryEditor : _originalGeometryEditor);
             if (MapView != null)
             {
-                if (MapView.SketchEditor != sketchEditor)
+                if (MapView.GeometryEditor != geometryEditor)
                 {
-                    MapView.SketchEditor = sketchEditor;
+                    MapView.GeometryEditor = geometryEditor;
                 }
 
-                if (MapView.SketchEditor != null)
+                if (MapView.GeometryEditor != null)
                 {
-                    MapView.SketchEditor.IsVisible = isMeasuringLength || isMeasuringArea;
+                    MapView.GeometryEditor.IsVisible = isMeasuringLength || isMeasuringArea;
                 }
 
                 if (isMeasuringFeature)
@@ -271,7 +272,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             {
                 if (isMeasuringLength || isMeasuringArea)
                 {
-                    _clearButton.IsEnabled = MapView?.SketchEditor?.Geometry != null;
+                    _clearButton.IsEnabled = MapView?.GeometryEditor?.Geometry != null;
                 }
                 else
                 {
@@ -313,13 +314,13 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                     {
                         case MeasureToolbarMode.Line:
                             {
-                                geometry = LineSketchEditor?.Geometry;
+                                geometry = LineGeometryEditor?.Geometry;
                                 break;
                             }
 
                         case MeasureToolbarMode.Area:
                             {
-                                geometry = AreaSketchEditor?.Geometry;
+                                geometry = AreaGeometryEditor?.Geometry;
                                 break;
                             }
 
@@ -354,11 +355,11 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         }
 
         /// <summary>
-        /// Toggles between measure modes and starts SketchEditor when not already started for length and area.
+        /// Toggles between measure modes and starts <see cref="GeometryEditor"/> when not already started for length and area.
         /// </summary>
         /// <param name="sender">Toggle button that raised click event.</param>
         /// <param name="e">Contains information or event data associated with routed event.</param>
-        private async void OnToggleMeasureMode(object? sender, RoutedEventArgs e)
+        private void OnToggleMeasureMode(object? sender, RoutedEventArgs e)
         {
             var toggleButton = sender as ToggleButton;
             Mode = toggleButton != null && toggleButton.IsChecked.HasValue && toggleButton.IsChecked.Value ?
@@ -366,13 +367,13 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
                toggleButton == _measureAreaButton ? MeasureToolbarMode.Area :
                toggleButton == _measureFeatureButton ? MeasureToolbarMode.Feature : MeasureToolbarMode.None) :
                MeasureToolbarMode.None;
-            if (MapView?.SketchEditor != null && MapView.SketchEditor.Geometry == null && (Mode == MeasureToolbarMode.Line || Mode == MeasureToolbarMode.Area))
+            if (MapView?.GeometryEditor != null && MapView.GeometryEditor.Geometry == null && (Mode == MeasureToolbarMode.Line || Mode == MeasureToolbarMode.Area))
             {
                 try
                 {
-                    var creationMode = Mode == MeasureToolbarMode.Line ? SketchCreationMode.Polyline : SketchCreationMode.Polygon;
-                    var geometry = await MapView.SketchEditor.StartAsync(creationMode);
-                    DisplayResult(geometry);
+                    var creationMode = Mode == MeasureToolbarMode.Line ? GeometryType.Polyline : GeometryType.Polygon;
+                    MapView.GeometryEditor.Start(creationMode);
+                    DisplayResult(MapView.GeometryEditor.Geometry);
                 }
                 catch (TaskCanceledException)
                 {
@@ -385,18 +386,26 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         }
 
         /// <summary>
-        /// Displays the measurement result for the given new geometry.
+        /// Handles the <see cref="INotifyPropertyChanged.PropertyChanged"/> event for the <see cref="GeometryEditor"/>
+        /// instance.
         /// </summary>
-        /// <param name="sender">SketchEditor that raised GeometryChanged event.</param>
-        /// <param name="e">Data for the GeometryChanged event.</param>
-        private void OnGeometryChanged(object? sender, GeometryChangedEventArgs e)
+        /// <remarks>This method updates the state of the clear button based on the <see
+        /// cref="GeometryEditor.Geometry"/> property and displays the updated geometry result. The <paramref
+        /// name="sender"/> must be a <see cref="GeometryEditor"/>  instance for the method to function
+        /// correctly.</remarks>
+        /// <param name="sender">The source of the event, expected to be a <see cref="GeometryEditor"/> instance.</param>
+        /// <param name="e">The event data containing the name of the property that changed.</param>
+        private void OnGeometryEditorPropertyChanged(object? sender, PropertyChangedEventArgs e)
         {
-            if (_clearButton != null)
+            if (e.PropertyName is nameof(GeometryEditor.Geometry))
             {
-                _clearButton.IsEnabled = e.NewGeometry != null;
+                var editor = sender as GeometryEditor;
+                if (_clearButton != null)
+                {
+                    _clearButton.IsEnabled = editor?.Geometry != null;
+                }
+                DisplayResult(editor?.Geometry);
             }
-
-            DisplayResult(e.NewGeometry);
         }
 
         /// <summary>
@@ -531,7 +540,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         {
             if (Mode == MeasureToolbarMode.Line || Mode == MeasureToolbarMode.Area)
             {
-                MapView?.SketchEditor?.ClearGeometry();
+                MapView?.GeometryEditor?.ClearGeometry();
             }
             else if (Mode == MeasureToolbarMode.Feature)
             {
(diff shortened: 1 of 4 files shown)


src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -385,18 +386,26 @@ private async void OnToggleMeasureMode(object? sender, RoutedEventArgs e)
...
+        /// <remarks>This method updates the state of the clear button based on the <see
+        /// cref="GeometryEditor.Geometry"/> property and displays the updated geometry result. The <paramref
+        /// name="sender"/> must be a <see cref="GeometryEditor"/>  instance for the method to function
+        /// correctly.</remarks>
+        /// <param name="sender">The source of the event, expected to be a <see cref="GeometryEditor"/> instance.</param>
+        /// <param name="e">The event data containing the name of the property that changed.</param>
+        private void OnGeometryEditorPropertyChanged(object? sender, PropertyChangedEventArgs e)
         {
-            if (_clearButton != null)
+            if (e.PropertyName is nameof(GeometryEditor.Geometry))
```
dotMorten, line 401: using `is` instead of `==` here is a little odd.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -354,25 +355,25 @@ private void DisplayResult(Geometry.Geometry? geometry = null)
...
                MeasureToolbarMode.None;
-            if (MapView?.SketchEditor != null && MapView.SketchEditor.Geometry == null && (Mode == MeasureToolbarMode.Line || Mode == MeasureToolbarMode.Area))
+            if (MapView?.GeometryEditor != null && MapView.GeometryEditor.Geometry == null && (Mode == MeasureToolbarMode.Line || Mode == MeasureToolbarMode.Area))
             {
                 try
                 {
-                    var creationMode = Mode == MeasureToolbarMode.Line ? SketchCreationMode.Polyline : SketchCreationMode.Polygon;
-                    var geometry = await MapView.SketchEditor.StartAsync(creationMode);
-                    DisplayResult(geometry);
+                    var creationMode = Mode == MeasureToolbarMode.Line ? GeometryType.Polyline : GeometryType.Polygon;
```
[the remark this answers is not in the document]
prathameshnarkhede: Yes. Verified that they are right replacements.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -354,25 +355,25 @@ private void DisplayResult(Geometry.Geometry? geometry = null)
...
+                    MapView.GeometryEditor.Start(creationMode);
```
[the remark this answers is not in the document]
prathameshnarkhede: I've made sure that removing this awaitable method is handling the event in correct synchronization context.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -112,10 +113,14 @@ public MeasureToolbar()
...
                     Geometry.AreaUnits.SquareYards,
                 };
-            LineSketchEditor = new SketchEditor();
-            AreaSketchEditor = new SketchEditor();
-            SelectionLineSymbol = LineSketchEditor.Style?.LineSymbol;
-            SelectionFillSymbol = AreaSketchEditor.Style?.FillSymbol;
+            LineGeometryEditor = new GeometryEditor();
+            AreaGeometryEditor = new GeometryEditor();
+            SelectionLineSymbol = LineGeometryEditor.Tool?.Style?.LineSymbol;
+            SelectionFillSymbol = AreaGeometryEditor.Tool?.Style?.FillSymbol;
```
dotMorten: What happens if the tool or style is null? Then the symbol wouldn't get set. If you know they are always set, resolve the warning with `!` and comment what the default is or explicitly instantiate it. 
Furthermore, the geometry editor is public here, so the developer could be modifying the editors and remove/replace/reconfigure the tool.
Does it make more sense to not expose the various geometry editors instances, just have one editor internally, and instead expose the various tool configurations?

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
prathameshnarkhede: I double checked and played around a bit with `is` operator along with `nameof` and seems to be working same as `==` operator.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -112,10 +116,17 @@ public MeasureToolbar()
...
                 };
-            LineSketchEditor = new SketchEditor();
-            AreaSketchEditor = new SketchEditor();
-            SelectionLineSymbol = LineSketchEditor.Style?.LineSymbol;
-            SelectionFillSymbol = AreaSketchEditor.Style?.FillSymbol;
+            if (_geometryEditor.Tool is VertexTool tool)
+            {
+                AreaVertexTool = tool;
+                LineVertexTool = tool;
+            }
```
dotMorten: What is the purpose of this condition? Shouldn't we be instantiating two vertex tools instead? What happens if it isn't?

Also AreaVertexTool and LineVertexTool here is sharing the same instance. I doubt that was intended.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -112,10 +116,17 @@ public MeasureToolbar()
...
+            SelectionLineSymbol = _geometryEditor.Tool!.Style!.LineSymbol; // Default tool and style are always set by GeometryEditor
+            SelectionFillSymbol = _geometryEditor.Tool!.Style!.FillSymbol; // Default tool and style are always set by GeometryEditor
```
dotMorten: Considering the style has a symbol for each, do we even need two tools?

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -112,10 +116,17 @@ public MeasureToolbar()
...
+            if (SelectionFillSymbol is Symbology.SimpleFillSymbol fillSymbol)
+            {
+                fillSymbol.Color = System.Drawing.Color.FromArgb(90, 60, 60, 60);
+            }
```
dotMorten: Instead of the if-condition, just make it a new instance of FillSymbol with the settings you want, so you know what it is.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -215,19 +226,23 @@ private void PrepareMeasureMode()
...
             if (MapView != null)
             {
-                if (MapView.SketchEditor != sketchEditor)
+                if (MapView.GeometryEditor is GeometryEditor geometryEditor)
                 {
-                    MapView.SketchEditor = sketchEditor;
-                }
+                    if (geometryEditor.Tool != tool)
+                    {
+                        _originalTool = MapView.GeometryEditor.Tool as VertexTool;
```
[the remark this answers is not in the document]
dotMorten: Don't cast the original tool. Just store the base-class instance. This is used for restoring the geometryeditor back to original condition once measure completes.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -271,7 +286,7 @@ private void PrepareMeasureMode()
             {
                 if (isMeasuringLength || isMeasuringArea)
                 {
-                    _clearButton.IsEnabled = MapView?.SketchEditor?.Geometry != null;
+                    _clearButton.IsEnabled = MapView?.GeometryEditor?.Geometry != null;
```
dotMorten: use the private geometryeditor instance directly here instead of going to MapView.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -354,25 +364,27 @@ private void DisplayResult(Geometry.Geometry? geometry = null)
...
         /// <summary>
-        /// Toggles between measure modes and starts SketchEditor when not already started for length and area.
+        /// Toggles between measure modes and starts <see cref="GeometryEditor"/> when not already started for length and area.
         /// </summary>
         /// <param name="sender">Toggle button that raised click event.</param>
         /// <param name="e">Contains information or event data associated with routed event.</param>
-        private async void OnToggleMeasureMode(object? sender, RoutedEventArgs e)
+        private void OnToggleMeasureMode(object? sender, RoutedEventArgs e)
         {
+            _geometryEditor.PropertyChanged += OnGeometryEditorPropertyChanged;
```
dotMorten: Just hook this up in the creation of _geometryEditor in the constructor once and for all, so you don't need to unsubscribe.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -354,25 +364,27 @@ private void DisplayResult(Geometry.Geometry? geometry = null)
...
+
             var toggleButton = sender as ToggleButton;
             Mode = toggleButton != null && toggleButton.IsChecked.HasValue && toggleButton.IsChecked.Value ?
                (toggleButton == _measureLengthButton ? MeasureToolbarMode.Line :
                toggleButton == _measureAreaButton ? MeasureToolbarMode.Area :
                toggleButton == _measureFeatureButton ? MeasureToolbarMode.Feature : MeasureToolbarMode.None) :
                MeasureToolbarMode.None;
-            if (MapView?.SketchEditor != null && MapView.SketchEditor.Geometry == null && (Mode == MeasureToolbarMode.Line || Mode == MeasureToolbarMode.Area))
+            if (MapView?.GeometryEditor != null && (Mode == MeasureToolbarMode.Line || Mode == MeasureToolbarMode.Area))
```
dotMorten: Here and everywhere else: Don't access `MapView?.GeometryEditor`, but access `_geometryEditor`. 
The only time you access MapView.GeometryEditor directly is when assigning _geometryEditor and restoring the original editor back.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -559,91 +579,71 @@ public MapView? MapView
...
             {
                 throw new ArgumentException($"{nameof(MapView)} cannot be null or empty.");
             }
 
-            var oldMapView = e.OldValue as MapView;
-            if (oldMapView != null)
+            if (e.OldValue is MapView oldMapView)
             {
-                oldMapView.SketchEditor = toolbar._originalSketchEditor;
+                oldMapView.GeometryEditor = toolbar._geometryEditor;
```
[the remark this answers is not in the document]
dotMorten: I agree with the AI here to some extent. The entire logic of assigning mapview.geometryeditor up front is just wrong. We only want to assign the geometry editor when a measure starts, and once done, restore it back to whatever it was before (we don't want to permanently overwrite users other geometry editors here).

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -559,91 +579,71 @@ public MapView? MapView
...
                 oldMapView.GeoViewTapped -= toolbar.OnMapViewTapped;
                 toolbar.RemoveMeasureFeatureResultOverlay(oldMapView);
             }
 
             newMapView.GeoViewTapped += toolbar.OnMapViewTapped;
-            toolbar._originalSketchEditor = newMapView.SketchEditor;
-            toolbar.DisplayResult(newMapView.SketchEditor?.Geometry);
+            toolbar._geometryEditor = newMapView.GeometryEditor!;
+            toolbar.DisplayResult(newMapView.GeometryEditor?.Geometry);
```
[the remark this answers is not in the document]
dotMorten: Yes this seems fishy

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -559,91 +579,71 @@ public MapView? MapView
...
         }
 
         /// <summary>
-        /// Gets or sets the sketch editor used for measuring distances.
+        /// Gets or sets the <see cref="VertexTool"/> used for measuring distances.
         /// </summary>
-        public SketchEditor? LineSketchEditor
+        public VertexTool? LineVertexTool
```
dotMorten: Should these just be read-only? I don't see much need for replacing them. They are still mutable.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -559,91 +579,71 @@ public MapView? MapView
...
-
-            var oldSketchEditor = e.OldValue as SketchEditor;
-            if (oldSketchEditor != null)
-            {
-                oldSketchEditor.GeometryChanged -= toolbar.OnGeometryChanged;
-            }
-
-            newSketchEditor.GeometryChanged += toolbar.OnGeometryChanged;
-            toolbar.DisplayResult(newSketchEditor.Geometry);
+            toolbar.DisplayResult();
```
dotMorten: Changing the tool wouldn't change the result

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -354,25 +388,31 @@ private void DisplayResult(Geometry.Geometry? geometry = null)
...
         /// </summary>
         /// <param name="sender">Toggle button that raised click event.</param>
         /// <param name="e">Contains information or event data associated with routed event.</param>
-        private async void OnToggleMeasureMode(object? sender, RoutedEventArgs e)
+        private void OnToggleMeasureMode(object? sender, RoutedEventArgs e)
         {
+            if (_geometryEditor is not null)
+            {
+                _geometryEditor.PropertyChanged -= OnGeometryEditorPropertyChanged;
+                _geometryEditor.PropertyChanged += OnGeometryEditorPropertyChanged;
```
dotMorten: Don't ever do -+ thing. Subscribe and unsubscribe in the correct place instead (ie when the editor gets assigned and unassigned)

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -423,11 +471,11 @@ private async void OnMapViewTapped(object? sender, GeoViewInputEventArgs e)
             Symbology.Symbol? symbol = null;
             if (geometry is Polyline)
             {
-                symbol = SelectionLineSymbol;
+                symbol = _geometryEditor?.Tool?.Style?.LineSymbol;
             }
             else if (geometry is Polygon || geometry is Envelope)
             {
-                symbol = SelectionFillSymbol;
+                symbol = _geometryEditor?.Tool?.Style?.FillSymbol;
```
dotMorten: Are we actually using this symbol for anything now?

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -559,144 +607,35 @@ public MapView? MapView
...
-        }
-
-        /// <summary>
-        /// Gets or sets the sketch editor used for measuring distances.
-        /// </summary>
-        public SketchEditor? LineSketchEditor
-        {
-            get { return GetValue(LineSketchEditorProperty) as SketchEditor; }
-            set { SetValue(LineSketchEditorProperty, value); }
+            toolbar._geometryEditor = newMapView.GeometryEditor!;
```
dotMorten: Don't do this until start of editing. Clear it out again when the editor stops or mapview is null.
Having null-forgiveness operator here feels wrong (isn't _geometryEditor already nullable?).

src/Toolkit/Toolkit.WPF/UI/Controls/MeasureToolbar/MeasureToolbar.Theme.xaml
```diff
@@ -127,8 +127,6 @@
                                     Padding="8,4,8,4"
                                     ToolTip="Undo"
                                     Grid.Column="5"
-                                    DataContext="{TemplateBinding MapView}"
-                                    Command="{Binding Path=SketchEditor.UndoCommand}"
```
dotMorten, line 131: Problem with this is you now lost the ability to enable/disable the button for when there is something to undo or not.
I'd instead suggest you create command properties on the toolbar you can bind here. Whether they can undo comes down to whether a measure is in process (meaning _geometryEditor is assigned if you make the change to only have it assigned while measure is in process) + geometry editor can undo (the first bit is important).

dotMorten (changes requested):
Getting closer, but still needs a little work. You need to remember the geometry editor is owned by the developer can could change on your, or be controlled externally. So user could call `MapView.GeometryEditor.Stop()`, or replace the editor at a later time after the toolbar has been assigned the mapview, so you shouldn't grab a reference until measure starts.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -215,19 +230,27 @@ private void PrepareMeasureMode()
...
+                    }
+                    else
+                    {
+                        // Restore the original tool when not measuring
+                        if (_originalTool != null)
+                        {
+                            geometryEditor.Tool = _originalTool;
+                            _originalTool = null;
+                        }
+                        geometryEditor.IsVisible = false;
```
dotMorten: Perhaps instead we should return it to the current state

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -215,19 +230,27 @@ private void PrepareMeasureMode()
...
-
-                if (MapView.SketchEditor != null)
-                {
-                    MapView.SketchEditor.IsVisible = isMeasuringLength || isMeasuringArea;
+                    if (isMeasuringLength || isMeasuringArea)
+                    {
+                        // Save the original tool before switching
+                        _originalTool ??= geometryEditor.Tool;
+                        geometryEditor.Tool = MeasureTool ?? _originalTool;
+                        geometryEditor.IsVisible = true;
```
dotMorten: If the editor is already in use, perhaps it should cancel?

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -559,144 +590,39 @@ public MapView? MapView
...
-            var newSketchEditor = e.NewValue as SketchEditor;
-            if (newSketchEditor == null)
-            {
-                throw new ArgumentException($"{nameof(AreaSketchEditor)} cannot be null or empty.");
-            }
-
-            var oldSketchEditor = e.OldValue as SketchEditor;
-            if (oldSketchEditor != null)
+            toolbar._geometryEditor = newMapView.GeometryEditor;
+            if (toolbar._geometryEditor != null)
```
dotMorten: Do these lines on measure start instead, and unhook on measure completion / cancellation (should also clear up hooks if I call `mapView.GeometryEditor.Cancel()` externally).

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
prathameshnarkhede: Removing Visibility toggle, instead relying on Start/Stop operations.

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
prathameshnarkhede: Moving this bit nearer to Toggle Measure Mode so that it is initialized when Toggle mode is set to Line or Area measurements

src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -351,52 +363,86 @@ private void DisplayResult(Geometry.Geometry? geometry = null)
...
-        private void OnGeometryChanged(object? sender, GeometryChangedEventArgs e)
+        private void EndMeasureSession()
         {
-            if (_clearButton != null)
+            MapView?.GeometryEditor?.Stop();
+            Mode = MeasureToolbarMode.None;
+            if (_geometryEditor is GeometryEditor geometryEditor)
             {
-                _clearButton.IsEnabled = e.NewGeometry != null;
+                if (_originalTool is not null && _isGeometryEditorHooked)
```
mstefarov: You would need to unhook the OnGeometryEditorPropertyChanged event even if the original tool _was_ null, right?

prathameshnarkhede: Ah. Right. Thanks for catching that.

PR #694, opened 2025-09-09
Popup: Added Audio and Video support (WPF and WinUI)

prathameshnarkhede:
Currently, Only added support for WinUI and WPF platforms.
Tested with URL sources for Audio and Video.

https://github.com/user-attachments/assets/4684c1a1-6de1-4fe2-b14f-965803711a77

### DIFF
--- a/src/Toolkit/Toolkit.WPF/Internal/MediaInlineControl.cs
+++ b/src/Toolkit/Toolkit.WPF/Internal/MediaInlineControl.cs
@@ -1,6 +1,4 @@
-﻿using System.Windows;
-using System.Windows.Controls;
-using System.Windows.Media;
+﻿using System.Windows.Threading;
 
 namespace Esri.ArcGISRuntime.Toolkit.Internal
 {
@@ -14,13 +12,20 @@ namespace Esri.ArcGISRuntime.Toolkit.Internal
     }
 
     /// <summary>
-    /// Private control for displaying inline media with play/pause overlay.
+    /// Control for displaying inline media with playback controls.
     /// </summary>
     internal class MediaInlineControl : UserControl
     {
+        private const string PlayButtonGlyph = "\u25B6";
+        private const string PauseButtonGlyph = "\u23F8";
         private readonly MediaElement _mediaElement;
-        private readonly Grid _overlay;
+        private readonly Button _playPauseButton;
+        private readonly Slider _progressSlider;
+        private readonly TextBlock _timestampText;
+        private readonly DispatcherTimer _timer;
         private bool _isPlaying;
+        private bool _isSliderDragging;
+        private TimeSpan _duration;
 
         public MediaInlineControl(Uri source, MediaType mediaType)
         {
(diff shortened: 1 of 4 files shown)


src/Toolkit/Toolkit.WPF/Internal/MediaInlineControl.cs
```diff
@@ -0,0 +1,193 @@
...
+                _progressSlider.Value = 0;
+                UpdateTimestamp();
+            };
+            _mediaElement.MediaFailed += (s, e) =>
+            {
+                _isPlaying = false;
+                _playPauseButton.Content = PlayButtonGlyph;
+                _timer.Stop();
+                UpdateTimestamp();
+            };
```
dotMorten: Do these events always trigger on the UI thread? If not do we need a call to the dispatcher first before touching UI components?

dotMorten (approved):

src/Toolkit/Toolkit.WinUI/Internal/HtmlToView.WinUI.cs
```diff
@@ -169,6 +169,33 @@ private static FrameworkElement CreateBlock(MarkupNode node, EventHandler<Uri>?
...
+                        var attr = HtmlUtility.ParseAttributes(child.Token?.Attributes);
+                        attr.TryGetValue("type", out mediaType);
+                        break;
+                    }
+                }
+                if (!string.IsNullOrEmpty(mediaSrc))
+                {
+                    var mediaPlayerElement = new MediaPlayerElement
+                    {
+                        Source = Windows.Media.Core.MediaSource.CreateFromUri(new Uri(mediaSrc, UriKind.RelativeOrAbsolute)),
```
mstefarov: I would use Uri.TryCreate instead of the constructor, so that we can fail more gracefully for bad URLs

src/Toolkit/Toolkit.WinUI/Internal/HtmlToView.WinUI.cs
```diff
@@ -169,6 +169,33 @@ private static FrameworkElement CreateBlock(MarkupNode node, EventHandler<Uri>?
...
+                        AreTransportControlsEnabled = true,
+                        Stretch = Stretch.Uniform,
+                    };
+                    return mediaPlayerElement;
+                }
+                return new TextBlock { Text = "Media not available" };
```
mstefarov, line 208: HTML audio and video tags are usually authored with _fallback content_.  It is arbitrary HTML inside the audio/video tag that should be displayed when media cannot be played/loaded.  We should display it here if possible.

For example: if the node has any non-Source children, render them (might be able to just do `goto case MarkupType.Block`).  And only use a TextBlock as a last resort when there's neither media to play, nor fallback content to show.

src/Toolkit/Toolkit.WPF/Internal/MediaInlineControl.cs
```diff
@@ -0,0 +1,193 @@
...
+            _mediaElement.MediaEnded += (s, e) =>
+            {
+                _mediaElement.Stop();
+                _isPlaying = false;
+                _playPauseButton.Content = PlayButtonGlyph;
+                _timer.Stop();
+                _progressSlider.Value = 0;
+                UpdateTimestamp();
+            };
+            _mediaElement.MediaFailed += (s, e) =>
```
mstefarov, line 170: It might be helpful to log the original media error (`e.ErrorException`) using Trace.WriteLine, like we do for e.g. layer load errors.

src/Toolkit/Toolkit.WPF/Internal/MediaInlineControl.cs
prathameshnarkhede: I just checked it's Dispatcher's access and seems to running on UI thread.

mstefarov (approved):

PR #697, opened 2025-09-15
Popup: Added Audio and Video support (MAUI iOS and Windows)

prathameshnarkhede:


src/Toolkit/Toolkit.Maui/Internal/MauiMediaElementHandler.Windows.cs
```diff
@@ -0,0 +1,34 @@
...
+                platformView.Source = MediaSource.CreateFromUri(source);
+                platformView.AreTransportControlsEnabled = true;
+                platformView.Stretch = Microsoft.UI.Xaml.Media.Stretch.Uniform;
+            }
+        }
+
+        protected override void ConnectHandler(MediaPlayerElement platformView)
+        {
+            base.ConnectHandler(platformView);
+        }
```
dotMorten: Is this needed?

src/Toolkit/Toolkit.Maui/Internal/MauiMediaElementHandler.iOS.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var url = NSUrl.FromString(source.ToString());
+            _player.ReplaceCurrentItemWithPlayerItem(new AVPlayerItem(url));
+        }
+    }
+
+    protected override void ConnectHandler(UIView platformView)
+    {
+        base.ConnectHandler(platformView);
+    }
```
dotMorten: Is this needed?

src/Toolkit/Toolkit.Maui/Internal/MauiMediaElement.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Esri.ArcGISRuntime.Toolkit.Maui.Internal;
+
+/// <summary>
+/// A view that displays audio or video media from a specified source.
+/// </summary>
+internal partial class MauiMediaElement : View
```
dotMorten, line 6: Should this class be ios/windows only?

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
```diff
@@ -163,6 +163,40 @@ private static View CreateBlock(MarkupNode node, EventHandler<Uri>? urlClickHand
                     imageElement.Source = imageSource;
                 return imageElement;
 
+            case MarkupType.Audio:
+            case MarkupType.Video:
```
dotMorten, line 167: What happens with this code on Android? Shouldn't that still show the placeholder instead?

src/Toolkit/Toolkit.Maui/Internal/MauiMediaElementHandler.Windows.cs
prathameshnarkhede: Ah. Not needed. Removing it.

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
prathameshnarkhede: I've redirected the markup to Block showing not supported for Android.

src/Toolkit/Toolkit.Maui/Internal/MauiMediaElement.cs
prathameshnarkhede: This class is currently platform-independent and doesn't contain any platform-specific logic. All the actual implementations are handled by the respective handlers. So, it should compile cleanly without requiring any conditional compilation directives.

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
```diff
@@ -163,6 +163,46 @@ private static View CreateBlock(MarkupNode node, EventHandler<Uri>? urlClickHand
...
+                // Find the first valid <source> child, or use the node's Content
+                string? mediaSrc = node.Content;
+                string? mediaType = null;
+                foreach (var child in node.Children)
+                {
+                    if (child.Type is MarkupType.Source && !string.IsNullOrEmpty(child.Content))
+                    {
+                        mediaSrc = child.Content;
+                        var attr = HtmlUtility.ParseAttributes(child.Token?.Attributes);
+                        attr.TryGetValue("type", out mediaType);
```
mstefarov: Is `mediaType` meant to be used somewhere here?  I see it is parsed but not used.

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
```diff
@@ -163,6 +163,46 @@ private static View CreateBlock(MarkupNode node, EventHandler<Uri>? urlClickHand
...
                 return imageElement;
 
+            case MarkupType.Audio:
+            case MarkupType.Video:
+#if ANDROID // TODO: This needs to be revisited when we have a support for Android
+                if (node.Children.Any(c => c.Type != MarkupType.Source))
+                {
+                    goto case MarkupType.Block;
+                }
+#endif
```
mstefarov: And what happens on Android if there is no fallback content?  Should we even create MauiMediaElement on Android, or skip straight to "Media not available"?

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
prathameshnarkhede: It's not needed. I used it in my previous logic. But forgot to remove. Removing now.

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
prathameshnarkhede: It already skips creating `MauiMediaElement` and to `Media not available` with `goto case` statement.

src/Toolkit/Toolkit.Maui/Internal/HtmlToView.Maui.cs
```diff
@@ -163,6 +163,43 @@ private static View CreateBlock(MarkupNode node, EventHandler<Uri>? urlClickHand
                     imageElement.Source = imageSource;
                 return imageElement;
 
+            case MarkupType.Audio:
+            case MarkupType.Video:
+#if ANDROID // TODO: This needs to be revisited when we have a support for Android
+                if (node.Children.Any(c => c.Type != MarkupType.Source))
+                {
+                    goto case MarkupType.Block;
+                }
```
mstefarov: I would add an 'else' fallback for Android -- just return the same "Media not available" label, instead of trying to create a MauiMediaElement that's not actually implemented on Android.  Or just remove the if statement and always goto Block.  An audio/video tag with no sources is malformed anyway.

Other than that, looks great 👍

PR #699, opened 2025-09-17
Bugfix: Adding converter to handle nullable bindings to boolean

prathameshnarkhede:
Introduce a Converter class to convert nullable values to Boolean.
This is required as part of a bug that is in MAUI Windows new static bindings.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/AttachmentsFormElementView.Maui.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/AttachmentsFormElementView.Maui.cs
@@ -53,7 +53,7 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui.Primitives
         private static object BuildDefaultTemplate()
         {
             var root = new VerticalStackLayout();
-            root.SetBinding(VerticalStackLayout.IsVisibleProperty, static (AttachmentsFormElementView view) => view.Element?.IsVisible, source: RelativeBindingSource.TemplatedParent);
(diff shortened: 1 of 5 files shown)


src/Toolkit/Toolkit/UI/Controls/FeatureForm/AttachmentsFormElementView.Maui.cs
```diff
@@ -53,7 +53,7 @@ static AttachmentsFormElementView()
         private static object BuildDefaultTemplate()
         {
             var root = new VerticalStackLayout();
-            root.SetBinding(VerticalStackLayout.IsVisibleProperty, static (AttachmentsFormElementView view) => view.Element?.IsVisible, source: RelativeBindingSource.TemplatedParent);
+            root.SetBinding(VerticalStackLayout.IsVisibleProperty, static (AttachmentsFormElementView view) => view.Element?.IsVisible, source: RelativeBindingSource.TemplatedParent, converter: BoolOrNullToBoolConverter.Instance);
```
mstefarov, line 56: We could probably avoid adding a new converter and just use `?? false` in these lambdas.

For example, `view.Element?.IsVisible ?? false` would resolve to a `bool` instead of `Nullable<bool>`.

What do you think?

prathameshnarkhede: Static binding currently doesn’t support logical operations or function resolution, which is a bit odd. It only works when properties are explicitly specified in the static binding expressions. Think of it as working somewhat like the `nameof()` operator, but replacing the property call, essentially making it property-safe and reducing the chance of typos. However, errors might not be apparent until compile time.

dotMorten: The reasons for this I believe is because of a code generator analyzing the expressions and turning them into some sort of binding-expression which limits the complexity of the expression

PR #706, opened 2025-12-15
Add a Job manager

dotMorten:
A port of the Swift toolkit JobManager. See: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/117a0a35662f7a2403b22d7048e4f0c53a97c07c/Sources/ArcGISToolkit/Components/JobManager/JobManager.swift

### DIFF
--- a/src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
+++ b/src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
@@ -27,7 +27,6 @@ namespace  Esri.ArcGISRuntime.Toolkit
         /// If you enable background status checks then you must also make sure to have enabled
         /// the "Background fetch" background mode in your application settings.
         /// </summary>
-        /// <exception cref="InvalidOperationException">Thrown if a schedule is set and the task identifier is not added to <c>info.plist</c>.</exception>
         /// <remarks>
         /// <note>
         /// You must also add the <see cref="StatusChecksTaskIdentifier"/> to the "Permitted


src/Toolkit/Toolkit/JobManager/JobManager.cs
```diff
@@ -0,0 +1,203 @@
...
+    /// </para>
+    /// </remarks>
+#pragma warning disable CA1060 // Move pinvokes to native methods class
+    public sealed partial class JobManager
+#pragma warning restore CA1060 // Move pinvokes to native methods class
+    {
+        /// <summary>
+        /// Gets the shared job manager.
+        /// </summary>
+        public static JobManager Shared { get; } = new JobManager(null);
```
prathameshnarkhede, line 76: Could this lead to a thread-safety issue? It might be a good idea to use Lazy loading for initialization to ensure safe and efficient singleton creation.

src/Toolkit/Toolkit/JobManager/JobManager.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        /// <summary>
+        /// Saves the current job state
+        /// </summary>
+        public void SaveState()
+        {
+            var array = Jobs.Select(j => j.ToJson()).ToArray();
+            lock (StateLock)
+            {
+                SaveState(string.Join("\n", array));
```
prathameshnarkhede: It looks like this implementation assumes there won't be any newline characters in actual `SaveState` data. If any serialized job contains a newline in a field like an error message or log. Splitting on `\n` character could break the deserialization.

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        private void RegisterBackgroundTask()
+        {
+            if (!IsAppPackaged)
+            {
+                return;
+            }
+            try
+            {
+                // TODO: Check app manifest for background task declaration and provide better instructions how to register the background task
```
prathameshnarkhede: Should this requirement be documented in the Readme about silent failures if not configured correctly?

prathameshnarkhede: Also, I was curious if the success and failure behavior are consistent across the platforms?

src/Samples/Toolkit.SampleApp.Maui/Toolkit.SampleApp.Maui.csproj
```diff
@@ -68,6 +68,8 @@
   <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net8.0'))">
     <PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiVersion)" />
     <PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
+    <PackageReference Include="Esri.Calcite.Maui" Version="1.0.0-rc.1" />
+  </ItemGroup>
```
prathameshnarkhede, line 72: This closing tag seems duplicated.

src/Toolkit/Toolkit/JobManager/JobManager.cs
```diff
@@ -0,0 +1,203 @@
...
+            }
+            SaveState();
+        }
+
+        private void Job_StatusChanged(object? sender, JobStatus e) =>  SaveState();
+
+        /// <summary>
+        /// The jobs being managed by the job manager.
+        /// </summary>
+        public IList<IJob> Jobs { get; } = new System.Collections.ObjectModel.ObservableCollection<IJob>();
```
prathameshnarkhede: This will reinitialize the Jobs collection every time we call `get` on Jobs and seems like a bug. We can make use of private backing field which stays for the lifetime of the JobManager.

src/Toolkit/Toolkit/JobManager/JobManager.cs
```diff
@@ -0,0 +1,203 @@
...
+        /// </summary>
+        /// <returns></returns>
+        public async Task PerformStatusChecks()
+        {
+            var tasks = Jobs.Select(async job =>
+            {
+                try { await job.CheckStatusAsync(); } catch { }
+            });
+            await Task.WhenAll(tasks);
+        }
```
prathameshnarkhede, line 181: This process could be made faster using `SimaphoreSlim` or `Parallel.ForEachAsync` which supports concurrency.

src/Samples/Toolkit.SampleApp.Maui/Samples/JobManagerSample.xaml.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            if (syncModel == SyncModel.Layer)
+            {
+                foreach (var info in task.ServiceInfo!.LayerInfos)
+                {
+                    parameters.LayerOptions.Add(new GenerateLayerOption(info.Id));
+                }
+            }
+
+            var path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), new Guid().ToString() + ".geodatabase");
```
mstefarov: `new Guid()` is the default constructor -- it's always going to be `00000000-0000-0000-0000-000000000000`.  Did you mean to do [`Guid.NewGuid()`](https://learn.microsoft.com/en-us/dotnet/api/system.guid.newguid?view=net-10.0)?


```suggestion
            var path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), Guid.NewGuid().ToString() + ".geodatabase");
```

src/Samples/Toolkit.SampleApp.Maui/Samples/JobView.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+        set { SetValue(JobProperty, value); }
+    }
+
+    public static readonly BindableProperty JobProperty =
+        BindableProperty.Create("Job", typeof(IJob), typeof(JobView), null, propertyChanged: (s, oldJob, newJob) => ((JobView)s).OnJobPropertyChanged(oldJob as IJob, newJob as IJob));
+
+    private void OnJobPropertyChanged(IJob? oldJob, IJob? newJob)
+    {
+        if (oldJob is not null)
+            oldJob.ProgressChanged -= Job_ProgressChanged;
```
mstefarov, line 45: Should this unsubscribe from `oldJob.StatusChanged` too -- like we do for `ProgressChanged`?

src/Toolkit/Toolkit/JobManager/JobManager.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        private object StateLock = new object();
+
+        /// <summary>
+        /// Saves the current job state
+        /// </summary>
+        public void SaveState()
+        {
+            var array = Jobs.Select(j => j.ToJson()).ToArray();
```
mstefarov: Events like ProgressChanged/StatusChanged will be coming from background threads and calling SaveState. Do we need some locks around this to avoid overlapping saves? Or e.g. what if user adds a Job (modifying Jobs collection) on the UI thread while SaveState is enumerating in the background?  That's a recipe for "collection was modified" exceptions.

src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// More information can be found <a href="https://developer.apple.com/documentation/backgroundtasks/refreshing_and_maintaining_your_app_using_background_tasks">here</a>.
+        /// </para>
+        /// </remarks>
+        public BackgroundStatusCheckSchedule PreferredBackgroundStatusCheckSchedule
+        {
+            get { return _preferredBackgroundStatusCheckSchedule; }
+            set
+            {
+                if (value != BackgroundStatusCheckSchedule.Disabled && ValidateInfoPlist())
+                    _preferredBackgroundStatusCheckSchedule = value;
```
mstefarov: If you ever enable this property, it would be impossible to disable due to this `if` check. Is that intentional?

src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
```diff
@@ -0,0 +1,213 @@
...
+            {
+                var value = array.GetItem<Foundation.NSString>(i);
+                if (value != null && value.ToString() == DefaultsKey)
+                {
+                    return true;
+                }
+            }
+            System.Diagnostics.Trace.WriteLine($"'BGTaskSchedulerPermittedIdentifiers' must contain '{DefaultsKey}'.", "ArcGIS Toolkit");
+            return false;
+            throw new InvalidOperationException($"'BGTaskSchedulerPermittedIdentifiers' must contain '{DefaultsKey}'.");
```
mstefarov: Were you undecided about returning or throwing? The xmldoc on PreferredBackgroundStatusCheckSchedule says that we throw, but the setter actually just fails. The `throw` is unreachable.

src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
```diff
@@ -0,0 +1,213 @@
...
+            if (PreferredBackgroundStatusCheckSchedule == BackgroundStatusCheckSchedule.Disabled)
+                return;
+
+            _isBackgroundStatusChecksScheduled = true;
+            var request = new BGAppRefreshTaskRequest(StatusChecksTaskIdentifier)
+            {
+                EarliestBeginDate = NSDate.Now.AddSeconds(PreferredBackgroundStatusCheckSchedule.Interval)
+            };
+            NSError error;
+            BGTaskScheduler.Shared.Submit(request, out error);
```
mstefarov: Should we perhaps check the `error` before setting `_isBackgroundStatusChecksScheduled` to true? It might have failed, [e.g.](https://developer.apple.com/documentation/backgroundtasks/bgtaskscheduler/submit(_:))

> There can be a total of 1 refresh task and 10 processing tasks scheduled at any time. Trying to schedule more tasks returns [BGTaskScheduler.Error.Code.tooManyPendingTaskRequests](https://developer.apple.com/documentation/backgroundtasks/bgtaskscheduler/error/code/toomanypendingtaskrequests).

src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
```diff
@@ -0,0 +1,213 @@
...
+            throw new InvalidOperationException($"'BGTaskSchedulerPermittedIdentifiers' must contain '{DefaultsKey}'.");
+        }
+    }
+
+    public class BackgroundStatusCheckSchedule
+    {
+        public static readonly BackgroundStatusCheckSchedule Disabled = new BackgroundStatusCheckSchedule();
+        public static BackgroundStatusCheckSchedule RegularInterval(double interval) => new BackgroundStatusCheckSchedule { Interval = interval };
+        public double Interval { get; private set; }
+    }
```
mstefarov: This class could use a bit of doc.  Also, it seems that you meant to make this factory-constructible -- should we make the default constructor private?

And while you're at it, the units of `interval` are not self-explanatory. How about using TimeSpan instead?

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,170 @@
...
+        internal sealed partial class BackgroundTask : IBackgroundTask
+        {
+            private BackgroundTaskDeferral? _deferral = null;
+            private volatile bool _cancelRequested = false;
+            private ThreadPoolTimer? _periodicTimer = null;
+            private JobManager? manager;
+            private IBackgroundTaskInstance? _taskInstance;
+
+            [MTAThread]
+            public void Run(IBackgroundTaskInstance taskInstance)
```
mstefarov: Is the `_taskInstance` field meant to be initialized here? Nothing assigns to that field right now.

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                
+                var taskId = taskInstance.Task.Name;
+                manager = JobManager.Create(taskId);
```
mstefarov: How does the BackgroundTask's JobManager communicate with the original JobManager that started it?  Are they meant to pass state around via the shared state file?  That seems pretty fragile -- both managers will be trying to read/write that file with no coordination.

As far as I can tell we never name our BackgroundTaskBuilder, so wouldn't `taskId` always be empty here?

src/Toolkit/Toolkit/JobManager/JobManager.cs
mstefarov: +1 on this. Newline-delimited JSON seems really fragile.

src/Toolkit/Toolkit/JobManager/JobManager.cs
mstefarov: Probably not -- CheckStatusAsync yields almost immediately, after signaling a background thread in core.  So we queue up all the checks and await them in parallel using `Task.WhenAll`.  This is about as efficient as it can get.

src/Toolkit/Toolkit/JobManager/JobManager.cs
dotMorten: As static initialization in C# is guaranteed to be thread-safe, public static JobManager Shared { get; } = new JobManager(null); is safe from race conditions during initialization. The runtime ensures the static constructor/initializers run once, even under concurrent access.

src/Toolkit/Toolkit/JobManager/JobManager.cs
dotMorten: No that's now how this works. If I had written
`public IList<IJob> Jobs => new System.Collections.ObjectModel.ObservableCollection<IJob>();` that would have been the case.

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,378 @@
...
+            }
+            catch (Exception ex)
+            {
+                System.Diagnostics.Trace.WriteLine("Failed to register background task for JobManager: " + ex.Message);
+            }
+        }
+
+        private void SaveState(string? json)
+        {
+            using var file = File.Open(GetStateFilename(), FileMode.Create, FileAccess.Write, FileShare.Write);
```
mstefarov: This SaveState implementation is still crash‑unsafe and concurrency‑unsafe.
- This truncates the file immediately and then writes the new content. LoadStateInternal has a chance to observe a half-written file.
- If the process/thread dies mid-write, we end up with an empty or partially-written file.
- With two possible writers (foreground or background), using FileShare.Write makes corruption _more_ likely. Both file-opens will succeed but writes will be interleaved.

You can fix some of these issues by writing to a temp file and then doing an atomic replacement (File.Move for initial write, File.Replace for overwrites). The writers are still racing each other, but at least the file on disk will always be intact.

If you want to eliminate writer races, you can do some cross-process synchronization -- either a [named mutex](https://learn.microsoft.com/en-us/dotnet/api/system.threading.mutex.-ctor?view=net-8.0#system-threading-mutex-ctor(system-boolean-system-string)) or a .lock file with FileShare.None.

src/Toolkit/Toolkit/JobManager/JobManager.cs
```diff
@@ -0,0 +1,270 @@
...
+            LoadState();
+            foreach (var job in Jobs)
+            {
+                job.StatusChanged += Job_StatusChanged;
+                job.ProgressChanged += Job_ProgressChanged;
+            }
+            ((INotifyCollectionChanged)Jobs).CollectionChanged += JobCollection_CollectionChanged;
+        }
+
+        private void Job_ProgressChanged(object? sender, EventArgs e) => SaveState();
```
mstefarov, line 117: Is it really necessary to save state for ProgressChanged events? These could be quite frequent, and would make the SaveState race conditions worse.

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,378 @@
...
+                    string taskId = taskName.Substring(34).Replace(".statusCheck", "");
+                    if (string.IsNullOrEmpty(taskId)) return;
+                    manager = JobManager.Create(taskId);
+                }
+                else
+                {
+                    return; // Unknown task id
+                }
+                manager.ResumeAllPausedJobs();
+                if (!manager.HasRunningJobs)
```
mstefarov: ResumeAllPausedJobs is async-void and returns right away, before jobs actually start running.  So if the only jobs are paused, I would expect the HasRunningJobs check to fail, and this BackgroundTask will exit right away.  Is that a problem?

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,378 @@
...
+                        JobManager._exitEvent.Set();
+                    }
+                    disposed = true;
+                }
+            }
+
+            ~BackgroundTask()
+            {
+                Dispose(false);
+            }
```
mstefarov, line 343: If I am reading this right, a BackgroundTask that gets GC'd (rather than disposed) will never set  `_exitEvent`. Wouldn't that cause the BGTaskServer to hang?  Should we perhaps set `_exitEvent` after completing the deferal?

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,378 @@
...
+                return System.IO.File.ReadAllText(GetStateFilename());
+            return null;
+        }
+
+        private string GetStateFilename()
+        {
+            if (IsAppPackaged)
+            {
+                var folderPath = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+                return Path.Combine(folderPath, $"{DefaultsKey}.json");
```
mstefarov: DefaultsKey can contain arbitrary `id` prodived by the user.  It might contain illegal characters or even slashes.  I'm not sure if I'd paste them directly into a filename.  What do you think about sanitizing or hashing it?  Or, if we actually need IDs to be simple strings, adding an argument check to `JobManager.Create(string)`

src/Toolkit/Toolkit/JobManager/JobManager.cs
dotMorten: Yup it is, otherwise when restoring we don't have the latest correct progress.

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
dotMorten: It does work but still a good point. I'll change it

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,412 @@
...
+        {
+            using (var hasher = System.Security.Cryptography.SHA256.Create())
+            {
+                StringBuilder hash = new StringBuilder();
+                byte[] bytes = hasher.ComputeHash(new UTF8Encoding().GetBytes(input));
+                for (int i = 0; i < bytes.Length / 2; i++) // Only grab the first half to keep path name short
+                {
+                    hash.Append(bytes[i].ToString("x2"));
+                }
+                return hash.ToString();
```
mstefarov: Optional: There is an easier way to get string hashes in .NET 8+


```suggestion
            byte[] bytes = System.Security.Cryptography.SHA256.HashData(Encoding.UTF8.GetBytes(input));
            return Convert.ToHexStringLower(bytes.AsSpan(0, bytes.Length / 2)); // Only grab the first half to keep path name short
```

src/Toolkit/Toolkit/JobManager/JobManager.Windows.cs
```diff
@@ -0,0 +1,412 @@
...
+            {
+                public uint CreateInstance(IntPtr objectAsUnknown, in Guid interfaceId, out IntPtr objectPointer)
+                {
+                    if (objectAsUnknown != IntPtr.Zero)
+                    {
+                        objectPointer = IntPtr.Zero;
+                        return CLASS_E_NOAGGREGATION;
+                    }
+
+                    if ((interfaceId != typeof(BackgroundTask).GUID) && (interfaceId != new Guid(IID_IUnknown)))
```
mstefarov: I believe the IClassFactory should be checking for the interface ID here -- not the class ID.

```suggestion
                    if ((interfaceId != typeof(IBackgroundTask).GUID) && (interfaceId != new Guid(IID_IUnknown)))
```

src/Toolkit/Toolkit/JobManager/JobManager.cs
prathameshnarkhede: Ah. I overlooked that the property is returning the same field instance with `get;`. Thanks for clarifying.

src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// <summary>
+        /// The preferred schedule for performing status checks while the application is in the
+        /// background. This allows an application to check to see if jobs have completed and optionally
+        /// post a local notification to update the user. The default value is `disabled`.
+        /// When the value of this property is not `disabled`, this setting is just a preference.
+        /// The operating system ultimately decides when to allow a background task to run.
+        /// If you enable background status checks then you must also make sure to have enabled
+        /// the "Background fetch" background mode in your application settings.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown if a schedule is set and the task identifier is not added to <c>info.plist</c>.</exception>
```
prathameshnarkhede: Should this be updated as we are not throwing but returning false instead now?

prathameshnarkhede (approved):
Looks good to me. Just had 1 comment.

src/Samples/Toolkit.SampleApp.Maui/Samples/JobManagerSample.xaml.cs
mstefarov: I see a few other places have been fixed, but this one still has a problem -- this should be `Guid.NewGuid()` not `new Guid()`.

src/Samples/Toolkit.SampleApp.WinUI/Samples/Jobs/JobViewControl.xaml.cs
```diff
@@ -0,0 +1,110 @@
...
+        set { SetValue(JobProperty, value); }
+    }
+
+    public static readonly DependencyProperty JobProperty =
+        DependencyProperty.Register("Job", typeof(IJob), typeof(JobViewControl), new PropertyMetadata(null, (s, e) => ((JobViewControl)s).OnJobPropertyChanged(e.OldValue as IJob, e.NewValue as IJob)));
+
+    private void OnJobPropertyChanged(IJob oldJob, IJob newJob)
+    {
+        if (oldJob is not null)
+            oldJob.ProgressChanged -= Job_ProgressChanged;
```
mstefarov, line 58: Should we unsubscribe from StatusChanged here too?

src/Samples/Toolkit.SampleApp.WPF/Samples/Jobs/JobViewControl.xaml.cs
```diff
@@ -0,0 +1,114 @@
...
+            set { SetValue(JobProperty, value); }
+        }
+
+        public static readonly DependencyProperty JobProperty =
+            DependencyProperty.Register("Job", typeof(IJob), typeof(JobViewControl), new PropertyMetadata(null, (s, e) => ((JobViewControl)s).OnJobPropertyChanged(e.OldValue as IJob, e.NewValue as IJob)));
+
+        private void OnJobPropertyChanged(IJob oldJob, IJob newJob)
+        {
+            if (oldJob is not null)
+                oldJob.ProgressChanged -= Job_ProgressChanged;
```
mstefarov, line 59: And here also, should we unsubscribe from StatusChanged too?

src/Toolkit/Toolkit/JobManager/JobManager.iOS.cs
```diff
@@ -0,0 +1,244 @@
...
+            _isBackgroundStatusChecksScheduled = true;
+            var request = new BGAppRefreshTaskRequest(StatusChecksTaskIdentifier)
+            {
+                EarliestBeginDate = NSDate.Now.AddSeconds(PreferredBackgroundStatusCheckSchedule.Interval)
+            };
+            BGTaskScheduler.Shared.Submit(request, out var Error);
+            if (Error != null)
+            {
+                System.Diagnostics.Trace.WriteLine($"Error scheduling background status check: {Error.LocalizedDescription}", "ArcGIS Toolkit");
+            }
```
mstefarov, line 108: Should this reset `_isBackgroundStatusChecksScheduled` back to false (or maybe don't even set it unless Submit succeeds)?

src/Toolkit/Toolkit/JobManager/JobManager.cs
```diff
@@ -0,0 +1,270 @@
...
+        private void Init()
+        {
+            LoadState();
+            foreach (var job in Jobs)
+            {
+                job.StatusChanged += Job_StatusChanged;
+                job.ProgressChanged += Job_ProgressChanged;
+            }
+            ((INotifyCollectionChanged)Jobs).CollectionChanged += JobCollection_CollectionChanged;
+        }
```
mstefarov, line 115: When I start a Windows process and it loads some jobs from persistence, will the background task ever be registered?

As far as I can tell, JobsRunningChanged only gets called when Jobs collection changes (after load) or when a job Status changes.  Maybe Init should call CheckAnyJobRunningStateChanged() to make sure that JobsRunningChanged gets called if there are any persisted jobs?

dotMorten: JobsRunningChanged just raises an event. We don't need to raise events during initialization, since there won't be any subscribers yet.

PR #722, opened 2026-03-04
UI test framework foundation

imalcolm1:
Adds an initial Appium-based UI Testing framework to the repo. Please refer to the [README](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/ian/uitests/src/Tests/UITests/README.md) for guidance, or reach out/suggest updates if the README is not sufficient.

# Notes
Feedback I have received so far that will probably come in subsequent PRs after this is merged:
1. Make running tests easier
    - Ideally all a user would ever have to do is click the start button and everything would get set up automatically. That is, they wouldn't even need to install Appium
2. Make building tests easier
    - Making tests is somewhat complicated
    - The first step would be to make a template to automatically create the mirrored test xaml files and code-behind
    - An llm skill would also be useful


mandel-capuchin (changes requested):
Overall, I think this is a really good PR and a fantastic way to start this. There are a few things that I would like to see done here:

## Higher abstractions

Keep the shape that you currently have, but we could split this in two 4 more clearly differentiated layers:

- *Runner/session layer:* Appium startup, build/install rules, crash handling, DPI normalization.
- *Host shell layer:* a stable contract for TestSearchBox, app metadata, and page loading.
- *Page driver layer:* ScaleLinePageDriver, CompassPageDriver, etc. Tests talk to drivers, not raw IDs.
- *Control contract layer:* shared automation IDs, parsed state objects, and visual analyzers per control.

Concretely, I’d make new tests follow this rule:

- Add mirrored XAML page per framework.
- Add shared code-behind if behavior is common.
- Add one page driver in Toolkit.UITests.Shared.
- Add one control contract/analyzer if the control needs parsing or image work.
- Keep actual [TestMethod] code short and behavior-focused.

This is obviously easier said than done, we should have a call on how to approach these

imalcolm1:
Merged this now so we can start adding test coverage. Further improvements will be their own PRs

PR #375, opened 2021-08-25
BasemapGallery implementation

nCastle1:
OMD generated for UWP:

image

OMD generated for Forms:

image

PR #724, opened 2026-03-06
Ian/basemap item equality main

imalcolm1:
The basemap gallery samples demonstrate adding and removing example BasemapItems (for example here: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/main/src/Samples/Toolkit.SampleApp.WPF/Samples/BasemapGallery/BasemapGalleryAppearanceSample.xaml.cs#L39-L58). The remove button claims to drop the "last" item in the gallery, but that is not actually what happens when the basemaps from the "add" button were involved. It turns out `BasemapItem.Equals()` currently only compares the items' `Basemap`s against each other, and ignores `.Name`, `.Tooltip`, and `.Thumbnail`. This meant the two items added by the "add" button were considered equal, so when `Gallery.AvailableBasemaps.Remove(Gallery.AvailableBasemaps.Last());` was called the first of the added items was being removed, not the actual last item.

This pull request updates the `.Equals()` function to add the metadata properties. It also changes the samples to explicitly remove the last item since even with this fix clicking the "add" button multiple times will still add items the new `.Equals()` function considers equal.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
@@ -322,14 +322,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
         /// <inheritdoc />
         public override int GetHashCode()
         {
-            unchecked
-            {
-                int hashCode = 17;
-                hashCode = 397 * (hashCode ^ (Name?.GetHashCode() ?? 0));
-                hashCode = 397 * (hashCode ^ (Tooltip?.GetHashCode() ?? 0));
-                hashCode = 397 * (hashCode ^ (Thumbnail?.GetHashCode() ?? 0));
-                return hashCode;
-            }
+            return Basemap?.Name?.GetHashCode() ?? -1;
         }
 
         /// <summary>
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
@@ -242,7 +242,9 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
                 return false;
             }
 
-            return EqualsBasemap(other.Basemap);
+            return Name == other.Name && Tooltip == other.Tooltip
+                                      && Thumbnail == other.Thumbnail
+                                      && EqualsBasemap(other.Basemap);
         }
 
         /// <summary>
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
@@ -322,7 +322,14 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
         /// <inheritdoc />
         public override int GetHashCode()
         {
-            return Basemap?.Name?.GetHashCode() ?? -1;
+            unchecked
+            {
+                int hashCode = 17;
+                hashCode = 397 * (hashCode ^ (Name?.GetHashCode() ?? 0));
+                hashCode = 397 * (hashCode ^ (Tooltip?.GetHashCode() ?? 0));
+                hashCode = 397 * (hashCode ^ (Thumbnail?.GetHashCode() ?? 0));
+                return hashCode;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -242,7 +242,9 @@ public bool Equals(BasemapGalleryItem? other)
                 return false;
             }
 
-            return EqualsBasemap(other.Basemap);
+            return Name == other.Name && Tooltip == other.Tooltip
+                                      && Thumbnail == other.Thumbnail
+                                      && EqualsBasemap(other.Basemap);
```
mstefarov: I am noticing that this class actually violates another important [equality overriding rule](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type): two objects that are equal must have equal hashcodes. If you really want to fix equality for this class, GetHashCode would also need a refactor. For example (using [classic FNV hash algo](https://stackoverflow.com/questions/102742/why-is-397-used-for-resharper-gethashcode-override)):

```cs
        /// <inheritdoc />
        public override int GetHashCode()
        {
            unchecked
            {
                int hashCode = 17;
                hashCode = (hashCode * 397) ^ (Name?.GetHashCode() ?? 0);
                hashCode = (hashCode * 397) ^ (Tooltip?.GetHashCode() ?? 0);
                hashCode = (hashCode * 397) ^ (Thumbnail?.GetHashCode() ?? 0);
                return hashCode;
            }
        }
```

(You don't _have_ to do detailed Basemap comparisons to keep GetHashCode fast and simple; Name/Tooltip/Thumbnail should be sufficient to sort items into buckets in most realistic scenarios.  Only the "equal means same hash" requirement exists &mdash; there is no inverse "unequal means different hashes" requirement)

imalcolm1: Ah good to know! Implemented

mstefarov (approved):
The remove-by-index change makes sense to me.

I am a bit less certain about Equals/HashCode change because I don't fully understand why they were implemented this specific way in the first place (in https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/375). I see that the internal `EqualsBasemap` method is used for comparing GeoModel's current basemap to the selected BasemapGallery item. But as far as I can tell nothing uses Equals/GetHashCode overrides directly.

Either way, the new behavior is a better implementation of equality/gethashcode. I'm just not sure if this code is actually used, or if it's a leftover from earlier draft of the code that ended up using EqualsBasemap instead. Maybe @dotMorten remembers.

SamhithaBurugupalli (approved):

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
prathameshnarkhede: @imalcolm1 Or instead, we could use `HashCode.Combine(Name, Tooltip, Thumbnail)`. This produces high-quality hashes and is JIT-optimized.

```C#

public override int GetHashCode()
    => HashCode.Combine(Name, Tooltip, Thumbnail);
```

mstefarov: We can use HashCode.Combine in `v.next` but not in `main` -- this is a `main` PR (has to compile for .NET Framework, UWP, and netstandard2.0)

prathameshnarkhede: Ah, didn’t catch that it is targeting main.

dotMorten (changes requested):
My gut feeling here is that two gallery items that has the same basemap should be considered the same, so I'm not quite sure that the change to Equals is desired.
The real issue here is in the sample adding two identical basemaps which doesn't really make any sense.
However it is important that the basemap you have currently might not actually be matching the basemapgalleryitem in the list, but they are equivalent, ie lets say the map opens with a imagery basemap defined in the webmap. That one needs to be equal to the imagery basemap item that is in the list, even though they came from two different sources, and aren't actually the same reference. Otherwise it might not show up as the currently selected basemap.

prathameshnarkhede (approved):

imalcolm1:
After further discussion it seems that a full fix will involve some sdk-side work to expose layer equality. Closing this PR until more work can be done

PR #737, opened 2026-04-13
Handle nullable DatePicker.Date for .NET 10 compatibility

SamhithaBurugupalli:
Issue: The feature form with a date picker does not load in in .net 10

In .NET 10, DatePicker.Date is nullable. Casted to DateTime? before converting to UTC to safely handle null values and keep the code compatible with both .NET 9 and .NET 10.

Workaround for breaking change introduced in https://github.com/dotnet/maui/pull/27921


dotMorten (approved):

src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.cs
mstefarov, line 57: Once we drop .NET 9 support, we can get rid of most of this code. We are only checking platform views because MAUI's DatePicker was NOT nullable before.

mstefarov (approved):

PR #738, opened 2026-04-14
FeatureForm .net10 fix

dotMorten:
DatePicker and TimePicker changed their properties [to be nullable in .NET MAUI](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/738), which is a breaking binary change.
This is a workaround that avoids the MissingMethodExceptions this change cause.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Maui.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Maui.cs
@@ -6,6 +6,7 @@ using Microsoft.Maui;
 using Microsoft.Maui.Controls.Internals;
 using Microsoft.Maui.Platform;
 using System.ComponentModel;
+using System.Runtime.CompilerServices;
 
 namespace Esri.ArcGISRuntime.Toolkit.Maui.Primitives
 {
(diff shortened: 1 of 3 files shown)


mstefarov (approved):
🤢

src/Toolkit/Toolkit/UI/Controls/FeatureForm/DateTimePickerFormInputView.Maui.cs
```diff
@@ -152,6 +153,61 @@ private void TimePicker_PropertyChanged(object? sender, PropertyChangedEventArgs
         }
 
         private void DatePicker_DateSelected(object? sender, DateChangedEventArgs e) => UpdateValue();
+
+
+#if MAUI && !NET10_0_OR_GREATER // Work around for breaking change in .NET MAUI 10.0
```
SamhithaBurugupalli, line 158: Shouldn't the condition `!NET10_0_OR_GREATER` be removed as it has both .net9 and .net10 related helper methods

dotMorten: No this is to remind us to remove all this code once we start compiling for net10 - it's a compile-time check and not a runtime check and we currently compile for net9

PR #741, opened 2026-04-21
a11y: [WPF][WinUI] Sets Automation properties on focusable controls of SearchViewControl

SamhithaBurugupalli:
Related issues: 
https://devtopia.esri.com/runtime/dotnet-api/issues/13368
https://devtopia.esri.com/runtime/dotnet-api/issues/14236

Sets Automation Names on the following:
- List of Suggestions
- List of Results
- List of Sources
- List view items of the lists mentioned above
- Search Button
- Clear search button
- Edit Text box 
- Toggle button which is used to display the list of sources

Note: I added title and subtitle to be part of Automation name on search results list view item. This is because, the title can be same in some cases (ex: if searched for starbucks), then the subtitle (which has address) can be used to identify the uniqueness. I explored the possibility of using Helptext for Subtitle but not all screen readers will read out the help text. So it would be difficult to choose from the list just based on the Title.


src/Toolkit/Toolkit.WinUI/UI/Controls/SearchView/SearchView.Theme.xaml
```diff
@@ -630,7 +639,8 @@
                             ItemTemplate="{StaticResource SourceTemplate}"
                             ItemsSource="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=SearchViewModel.Sources, Mode=OneWay}"
                             SelectedItem="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=SearchViewModel.ActiveSource, Mode=TwoWay}"
-                            Style="{StaticResource ListStyle}" />
+                            Style="{StaticResource ListStyle}"
+                            AutomationProperties.Name="{converter:LocalizedString Key=SearchSources}"/>
```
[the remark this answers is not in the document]
SamhithaBurugupalli: Screenshot 2026-04-22 114719
Not required for WinUI

src/Toolkit/Toolkit.WinUI/UI/Controls/SearchView/SearchView.Theme.xaml
```diff
@@ -563,15 +572,24 @@
...
                             ItemsSource="{Binding Source={StaticResource GroupedSuggestionsEx}, Mode=OneWay}"
                             Style="{StaticResource ListStyle}"
+                            AutomationProperties.Name="{converter:LocalizedString Key=SearchViewSearchSuggestions}"
+                            AutomationProperties.AutomationId="SearchSuggestionsList"
                             Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.SourceSelectVisibility, Mode=OneWay}">
                             <ListView.GroupStyle>
                                 <GroupStyle>
+                                    <GroupStyle.ContainerStyle>
+                                            <Style TargetType="GroupItem">
+                                                <Setter Property="AutomationProperties.Name" Value="{Binding Name.DisplayName}" />
```
mstefarov: Does this work?  [GroupStyle.ContainerStyle](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.itemscontrol.groupstyle?view=windows-app-sdk-1.8#microsoft-ui-xaml-controls-itemscontrol-groupstyle) has been deprecated for a long time, and the doc says that it's ignored for ItemsControls (such as ListView).

And I am not sure if there is a `Name` in scope to bind to.  AFAIK a GroupItem's DataContext in WinUI is an [ICollectionViewGroup](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.data.icollectionviewgroup?view=windows-app-sdk-1.8), which only has "Group" and "GroupItems" properties.

src/Toolkit/Toolkit.WPF/UI/Controls/SearchView/SearchView.Theme.xaml
```diff
@@ -352,20 +373,23 @@
...
+                                    AutomationProperties.AutomationId="SourceList"/>
                             </StackPanel>
                         </Border>
                     </Popup>
                     <TextBox
                         x:Name="QueryEntry"
                         Grid.Column="1"
                         Style="{StaticResource QueryEntryStyle}"
-                        Text="{Binding RelativeSource={RelativeSource Mode=TemplatedParent}, Path=SearchViewModel.CurrentQuery, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" />
+                        Text="{Binding RelativeSource={RelativeSource Mode=TemplatedParent}, Path=SearchViewModel.CurrentQuery, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" AutomationProperties.Name="{Binding Path=SearchViewModel.ActivePlaceholder}"
```
mstefarov: Do we need `RelativeSource={RelativeSource Mode=TemplatedParent}` for the AutomationProperties binding, as we have for the existing `Text` binding? I'm guessing we do.

src/Toolkit/Toolkit.WinUI/UI/Controls/SearchView/SearchView.Theme.xaml
```diff
@@ -514,7 +517,9 @@
                         x:Name="QueryEntry"
                         Grid.Column="1"
                         Style="{StaticResource QueryEntryStyle}"
-                        Text="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=SearchViewModel.CurrentQuery, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" />
+                        Text="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=SearchViewModel.CurrentQuery, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
+                        AutomationProperties.Name="{Binding Path=SearchViewModel.ActivePlaceholder}"
```
mstefarov: Same here. Probably needs `RelativeSource={RelativeSource Mode=TemplatedParent}` to find the SearchView Model.

src/Toolkit/Toolkit.WinUI/UI/Controls/SearchView/SearchView.Theme.xaml
```diff
@@ -563,15 +572,24 @@
                             ItemsSource="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=SearchViewModel.Suggestions, Mode=OneWay}"
                             SelectedItem="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=SelectedSuggestion, Mode=TwoWay}"
                             Style="{StaticResource ListStyle}"
+                            AutomationProperties.Name="{converter:LocalizedString Key=SearchViewSearchSuggestions}"
+                            AutomationProperties.AutomationId="SearchSuggestionsList"
```
mstefarov: Having two controls with the same AutomationId in the same scope, even if only one is visible at a time, could cause problems. In WinUI the invisible elements still have an automation peer, they just set IsOffScreen flag. And [AutomationId docs](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.automationproperties.automationid?view=winrt-26100) say that the value should be unique among sibling elements.

So let's use two different IDs for the two lists. For example, it can be "SearchSuggestionsListFlat" and "SearchSuggestionsListGrouped".

src/Toolkit/Toolkit/UI/Controls/SearchView/ISearchDisplayInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Provides display information for search-related items - Internal use only.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public interface ISearchDisplayInfo
+    {
+        /// <inheritdoc/>
+        string DisplayTitle { get; }
+
+        /// <inheritdoc/>
+        string? DisplaySubtitle { get; }
```
mstefarov: `<inheritdoc/>` doesn't work when the interface does not inherit anything. These need real `<summary>` tags.

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
```diff
@@ -373,6 +373,22 @@
     <value>Search</value>
     <comment>The tooltip text to display for the search button.</comment>
   </data>
+  <data name="SearchViewSelectSearchSource" xml:space="preserve">
+    <value>Select search source</value>
+    <comment>UIA AutomationProperties.Name for the button to display search sources, used by screen readers to identify the control</comment>
```
mstefarov, line 378: Nit: I'm not sure if the translation team are expected to know UIA jargon. I would simplify these comments to start with "Accessible name" or "Accessible label" instead of "UIA AutomationProperties.Name"

dotMorten: They aren't reading these comments anyway 🤷

src/Toolkit/Toolkit.WinUI/UI/Controls/SearchView/SearchView.Theme.xaml
SamhithaBurugupalli: Thanks for pointing this out. I might have tested only for WPF and missed to test the same for WinUI earlier. You are right this is not working. Found a similar issues reported in the WinUI repo, 
-https://github.com/microsoft/microsoft-ui-xaml/issues/4486
-https://github.com/microsoft/microsoft-ui-xaml/issues/8569
Implemented the suggested approach in these issues using `ChoosingGroupHeaderContainer` so that right automation name is set on group.

mstefarov (approved):

dotMorten (approved):

PR #753, opened 2026-05-12
Add copilot instructions

dotMorten:
Also update tests to all support `dotnet test` for the agent to use.


src/Analyzers/Toolkit.Maui.Analyzers.UnitTests/UseToolkitInitializationAnalyzerTests.cs
```diff
@@ -73,7 +73,7 @@ public static MauiApp CreateMauiApp()
     }
 
     [TestMethod]
-    public async Task CodeFixer()
+    public async Task CodeFixer()
```
mstefarov: Nit: unnecessary whitespace change
```suggestion
    public async Task CodeFixer()
```

mstefarov (approved):

PR #755, opened 2026-05-19
UITests: Windows CI Tests

imalcolm1:
Scripts for running CI tests for windows. All of the scripts are running on the Jenkins VMs when RDP is connected. Because of this tests are expected to fail until we get a physical machine to test on.

## Notes for reviewers
**Structure**
Most of the logic is stored in the `utils.psm1` module. The 3 platform-specific scripts just call `Invoke-WindowsUITests` with different parameters.

**Changes to test runners**
To simplify the ci logic I added the ability to configure app paths for the runner projects at runtime using environment variables. This has been documented in the UITest readme.

## Testing Locally
To test locally run the following commands in powershell. The NUGET_REPO directory should contain the nuget package from a daily build or a release, otherwise leave it unset.
```powershell
mkdir C:\uitest-workspace\
cd C:\uitest-workspace\
git clone https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit.git -b ian/uitests-cibuild-wpf-wap
$env:WORKSPACE = 'C:\uitest-workspace\'
$env:RELEASE_VERSION = '300.1.0' # This may need to change if NUGET_REPO used packages from a different version
$env:NUGET_REPO = 'C:\uitest-workspace\nuget-source' # This requires first copying nuget files to the given folder
& C:\uitest-workspace\arcgis-maps-sdk-dotnet-toolkit\src\Tests\UITests\cibuild\cibuild-wpf.ps1 
```

### DIFF
--- a/src/Tests/UITests/cibuild/utils.psm1
+++ b/src/Tests/UITests/cibuild/utils.psm1
@@ -23,7 +23,7 @@ function Install-Dotnet {
 
   if ([string]::IsNullOrWhiteSpace($dotnet_cache_folder) -or !(Test-Path -Path $dotnet_exe)) {
     $installerPath = Join-Path $workspace 'dotnet-install.ps1'
-    curl -L https://dot.net/v1/dotnet-install.ps1 -o $installerPath
+    & curl.exe -L https://dot.net/v1/dotnet-install.ps1 -o $installerPath
     & $installerPath -Version $dotnet_version -InstallDir $dotnet_install_folder -NoPath
     if ($LASTEXITCODE -ne 0) {
      exit 1
@@ -56,7 +56,7 @@ function Install-Nodejs {
     $node_url = "https://nodejs.org/dist/v$($node_version)/node-v$($node_version)-win-x64.zip"
     $node_zip = Join-Path $env:WORKSPACE "node.zip"
 
-    curl $node_url -o $node_zip
+    & curl.exe -L $node_url -o $node_zip
     Expand-Archive -Path $node_zip -Destination $env:WORKSPACE
 
     if (!($?) -or ($LASTEXITCODE -ne 0)) {
--- a/src/Tests/UITests/cibuild/utils.psm1
+++ b/src/Tests/UITests/cibuild/utils.psm1
@@ -8,11 +8,11 @@ function Invoke-WindowsUITests {
 
     [Parameter(Mandatory)]
     [ValidateNotNullOrEmpty()]
-    [string]$runner_project,
+    [string]$runner_name,
 
     [Parameter(Mandatory)]
     [ValidateNotNullOrEmpty()]
-    [string]$app_project,
+    [string]$app_name,
 
     [Parameter(Mandatory)]
     [string[]]$build_parameters,
@@ -25,6 +25,8 @@ function Invoke-WindowsUITests {
   )
 
   $config_file = Join-Path $PSScriptRoot "variables.yml"
+  $runner_project = Join-Path $PSScriptRoot "..\${runner_name}\${runner_name}.csproj"
+  $app_project = Join-Path $PSScriptRoot "..\${app_name}\${app_name}.csproj"
 
   # Install dotnet
   $dotnet_version = Get-YamlValue $config_file 'dotnet-version'
@@ -66,6 +68,10 @@ function Invoke-WindowsUITests {
     }
   }
 
+  # Ensure predictable artifacts output layout for setting env:UITEST_APP_PATH later
+  $build_params_artifacts = @('-p:ArtifactsPivots=TestBuild', '-p:UseArtifactsOutput=true')
+  $build_parameters += $build_params_artifacts
+
   # Build app and runner projects
   & $dotnet_exe build $app_project @build_parameters
   if ($LASTEXITCODE -ne 0) {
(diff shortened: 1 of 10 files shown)


src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,266 @@
...
+  if (![string]::IsNullOrWhiteSpace($nuget_repo)) {
+    $toolkit_src_dir = Join-Path $PSScriptRoot '..\..\..'
+    Set-NugetSource $toolkit_src_dir $dotnet_exe $nuget_repo
+    if (!$?) {
+      exit 1
+    }
+  }
+
+  # Start appium before build so it has time to initialize in the background
+  $appium_server_process = Start-Process -FilePath $node_exe -ArgumentList @($appium_entry) -PassThru
```
mstefarov: Start-Process for appium is now before the build, but the two `exit $LASTEXITCODE` paths in the app-build and runner-build error branches don't Stop-Process the appium handle. Same problem might happen if `dotnet test` fails. I think this can "leak" a process in Jenkins after the job ends, complicating cleanup/retrying.

Consider wrapping everything from this point on in a try-finally.

src/Tests/UITests/cibuild/WinAppDriver1.2.1.zip
mstefarov, line 1: I would NOT put this binary into our repo. Aside from bloating the repo size, the license/EULA from WinAppDriver is not redistribution-friendly. Can we just download it on-demand from https://github.com/microsoft/WinAppDriver/releases/download/v1.2.1/WindowsApplicationDriver_1.2.1.msi ?

If "unattended" install (`msiexec /i ... /qn`) doesn't work for Jenkins, we can use [administrative extract](https://superuser.com/a/788227/713655) to "unzip" the MSI package to a folder without installing.

src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,266 @@
...
+
+  # Run tests
+  $results_dir = Join-Path $workspace 'TestResults'
+  $test_run_params = @('--no-build', '--report-trx', '--results-directory', $results_dir)
+  if (![string]::IsNullOrWhiteSpace($trx_filename)) {
+    $test_run_params += @('--report-trx-filename', $trx_filename)
+  }
+
+  $toolkit_src_root = Join-Path $PSScriptRoot '..\..\..'
+  Set-Location -Path $toolkit_src_root
```
mstefarov: This will change the caller's working directory after the script returns, which is usually unexpected for a script.  Prefer `Push-Location` and a matching `Pop-Location` (in a finally) instead.

src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,266 @@
...
+  # Set nuget source if provided
+  if (![string]::IsNullOrWhiteSpace($nuget_repo)) {
+    $toolkit_src_dir = Join-Path $PSScriptRoot '..\..\..'
+    Set-NugetSource $toolkit_src_dir $dotnet_exe $nuget_repo
+    if (!$?) {
+      exit 1
+    }
+  }
+
+  # Start appium before build so it has time to initialize in the background
```
mstefarov: Whether Appium is ready is timing-based (i.e. luck-based) in the current script.  What do you think about polling Appium's [`/status` REST endpoint](https://appium.io/docs/en/latest/reference/api/webdriver/#getstatus) to see if it's _actually_ ready?  Something like...

```pws
$deadline = (Get-Date).AddSeconds(60)
while ((Get-Date) -lt $deadline) {
    try {
        $r = Invoke-RestMethod -Uri 'http://127.0.0.1:4723/status' -TimeoutSec 2
        if ($r.value.ready) { break }
    } catch { Start-Sleep -Milliseconds 500 }
}
```

...I might be over-engineering this though. Is Appium _always_ and _significantly_ faster to start than the code between here and `dotnet test`? If so, this is moot.

src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,266 @@
...
+  $appium_entry = Join-Path $node_workspace 'node_modules\appium\index.js'
+  $env:APPIUM_HOME = Join-Path $workspace '.appium'
+
+  & $npm_exe install appium --prefix $node_workspace
+  if ($LASTEXITCODE -ne 0) {
+    exit $LASTEXITCODE
+  }
+
+  & $node_exe $appium_entry driver install windows
+
```
mstefarov: This would benefit from an exit-code check, like other operations in this script

src/Tests/UITests/cibuild/utils.psm1
imalcolm1: I think it probably isn't necessary, but it's not much more code to add, and it means I can put it back to being started right before the test run rather than right before build. I'll test this approach on the machines

src/Tests/UITests/cibuild/WinAppDriver1.2.1.zip
imalcolm1: Thanks, that administrative extract was the missing piece! Rebasing so that commit can hopefully get garbage collected

imalcolm1:
@dotMorten @mstefarov @SamhithaBurugupalli The new machine is running and the tests seem to be performing well. ~~After talking with Sarat I've made an entry script on the internal repo that I've also tagged you as reviewers for. It should make running this a bit more straight forward too.~~

imalcolm1:
Commit history got corrupted. New PR here: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/760

PR #757, opened 2026-05-21
Offline Map Areas View

dotMorten:
Port of the OfflineMapAreas from Swift and Kotlin.
Reference source:
 - Swift: https://github.com/Esri/arcgis-maps-sdk-swift-toolkit/blob/main/Sources/ArcGISToolkit/Components/Offline/
 - Kotlin: https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/tree/main/toolkit/offline/src/main/java/com/arcgismaps/toolkit/offline

https://github.com/user-attachments/assets/81911c84-11f7-44c4-b40a-87a26c4e1dda


src/Toolkit/Toolkit/Internal/DispatcherExtensions.cs
```diff
@@ -32,6 +32,8 @@ internal static void Dispatch(this BindableObject bindable, Action action)
 #elif WPF
         internal static void Dispatch(this System.Windows.Threading.DispatcherObject dObject, Action action)
         {
+            if (dObject.Dispatcher.HasShutdownFinished || dObject.Dispatcher.HasShutdownFinished)
```
mstefarov: Double-checking just to be safe? 🙂 Or did you mean to

```suggestion
            if (dObject.Dispatcher.HasShutdownStarted || dObject.Dispatcher.HasShutdownFinished)
```

src/Tests/Toolkit.Tests/OfflineMapManagerTests.cs
```diff
@@ -0,0 +1,286 @@
...
+            }
+        }
+
+        private static async Task AwaitDownload(PreplannedMapModel model)
+        {
+            TaskCompletionSource downloadCompletionSource = new TaskCompletionSource();
+            System.ComponentModel.PropertyChangedEventHandler? handler = null;
+            handler = (s, e) =>
+            {
+                if (e.PropertyName == nameof(PreplannedMapModel.Status) || model.Status == PreplannedMapModelStatus.Downloaded)
```
mstefarov: Currently this returns on _any_ status change. I'm guessing this was supposed to be...


```suggestion
                if (e.PropertyName == nameof(PreplannedMapModel.Status) && model.Status == PreplannedMapModelStatus.Downloaded)
```

src/Tests/Toolkit.Tests/OfflineMapManagerTests.cs
```diff
@@ -0,0 +1,286 @@
...
+            try
+            {
+                await vm.LoadModelsAsync();
+                if (vm.PreplannedMapModels[0].Status == PreplannedMapModelStatus.Downloaded)
+                {
+                    vm.PreplannedMapModels[0].RemoveDownloadedArea();
+                }
+                if (vm.PreplannedMapModels[0].AllowsDownload == false)
+                {
+                    System.Diagnostics.Debugger.Break();
```
mstefarov: Did you mean to check this in -- and the second Debugger.Break below?

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,531 @@
...
+                if (Directory.Exists(_mmpkDirectoryPath))
+                {
+                    try
+                    {
+                        Directory.Delete(_mmpkDirectoryPath, true);
+                    }
+                    catch
+                    {
+                        Directory.Move(_mmpkDirectoryPath, _mmpkDirectoryPath + ".old");
+                    }
```
mstefarov: There are quite a few places in this PR where a directory has to be deleted / moved out of the way / emptied out.  I see that most are already centralized in `OfflineMapAreaUtilities.TryDeleteDirectory` but a few more are not (e.g. `PreplannedMapModel.DownloadPreplannedMapAreaAsync` and `OfflineManager.RemoveDownloads`).

I actually like _this_ particular implementation better than the current TryDeleteDirectory, because this one does not silently leave an undeletable directory.  Though it still risks throwing on _second retry_ (when a .old directory exists).

You could make it more robust by e.g. adding a retry loop (old1 / old2 / old3 etc).  Or by giving it a GUID name and moving it to a designated ".deleteme" subdirectory (akin to how you have a ".pending" subdirectory).  You could run best-effort cleanup on the ".deleteme" folder on OfflineManager init and after JobCompleted.

src/Toolkit/Toolkit/OfflineManager.cs
```diff
@@ -0,0 +1,605 @@
...
+            catch (Exception ex)
+            {
+                Trace.WriteLine($"ArcGIS Toolkit - Error saving pending offline map info: {ex}");
+            }
+        }
+
+        /// <summary>
+        /// Removes all downloads for all offline maps.
+        /// </summary>
+        public void RemoveAllDownloads()
```
mstefarov, line 312: This method does some pretty aggressive cleanup. If there are jobs active, something will definitely blow up. Do you think it's worth adding some kind of interlock -- e.g. cancel all jobs and wait for cancellation?

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapAreasView.cs
```diff
@@ -0,0 +1,545 @@
...
+            new OnDemandScaleOption("OfflineMapAreasScaleCountriesSmall", 18489297.737236),
+            new OnDemandScaleOption("OfflineMapAreasScaleCountriesBig", 36978595.474472),
+            new OnDemandScaleOption("OfflineMapAreasScaleContinent", 73957190.948944),
+            new OnDemandScaleOption("OfflineMapAreasScaleWorldSmall", 147914381.897889),
+            new OnDemandScaleOption("OfflineMapAreasScaleWorldBig", 295828763.795777),
+            new OnDemandScaleOption("OfflineMapAreasScaleWorld", 591657527.591555),
+        };
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FeatureFormView"/> class.
```
mstefarov:
```suggestion
        /// Initializes a new instance of the <see cref="OfflineMapAreasView"/> class.
```

src/Toolkit/Toolkit/OfflineManager.cs
```diff
@@ -0,0 +1,605 @@
...
+    /// <item>Run download jobs while the app is in the background.</item>
+    /// <item>Get notified when the jobs complete via the <see cref="JobCompleted"/> event.</item>
+    /// </list>
+    /// </para>
+    /// <para>
+    /// The component is useful both for building custom UI with the provided APIs,
+    /// and for supporting workflows that require retrieving offline map areas
+    /// information from the device. By using the `OfflineManager`, you can create
+    /// an `OfflineMapAreasView` using the ``OfflineMapAreasView/init(offlineMapInfo:selection:)``
+    /// initializer.
```
mstefarov: Doc: I spy Swift syntax!  We don't have an equivalent `OfflineMapAreasView(OfflineMapInfo)` constructor, do we?

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                            <Setter TargetName="ButtonBorder" Property="Background" Value="#FFE6E6E6" />
+                            <Setter TargetName="ButtonBorder" Property="BorderBrush" Value="#FFC6C6C6" />
+                        </Trigger>
+                        <Trigger Property="IsEnabled" Value="False">
+                            <Setter Property="Opacity" Value="0.5" />
+                        </Trigger>
+                    </ControlTemplate.Triggers>
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>-->
```
mstefarov: Can this dead code be deleted?

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,531 @@
...
+            Func<Task<OfflineMapTask>> offlineMapTaskFactory,
+            PreplannedMapArea preplannedMapArea,
+            string portalItemId,
+            string preplannedMapAreaId,
+            Action<PreplannedMapModel> onRemoveDownload,
+            System.Windows.Input.ICommand openMapCommand,
+            Action<Action> dispatcher) : this(offlineMapTaskFactory, portalItemId, preplannedMapAreaId,
+                preplannedMapArea.PortalItem?.Title ?? preplannedMapAreaId, preplannedMapArea.PortalItem?.Description, true, onRemoveDownload, openMapCommand,  dispatcher)
+        {
+            _preplannedMapArea = preplannedMapArea;;
```
mstefarov:
```suggestion
            _preplannedMapArea = preplannedMapArea;
```

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,531 @@
...
+#nullable enable
+
+#pragma warning disable CS1591
+#pragma warning disable SA1402
+
+using System.IO;
+using Esri.ArcGISRuntime.Mapping;
+using Esri.ArcGISRuntime.Tasks.Offline;
+using Esri.ArcGISRuntime.Toolkit.Internal;
+using Microsoft.VisualBasic.FileIO;
```
mstefarov:
```suggestion
```

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OnDemandMapModel.cs
```diff
@@ -0,0 +1,473 @@
...
+                });
+            }
+        }
+
+        public bool AllowsDownload => Status == OnDemandMapModelStatus.Initialized;
+
+        public bool IsDownloaded => Status == OnDemandMapModelStatus.Downloaded;
+
+        static IValueConverter _converter = new FileSizeConverter();
+        public string Description => SizeInBytes == 0 ? "" : $"Size: {_converter.Convert(SizeInBytes, typeof(string), null, null!)}";
```
mstefarov: Feels like this should be a localizable string

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapViewModel.cs
```diff
@@ -0,0 +1,476 @@
...
+            }
+
+            var model = new OnDemandMapModel(GetOfflineMapTaskAsync, configuration, PortalItemId, OnRemoveDownloadOfOnDemandArea, _openMapCommand, Dispatcher);
+            Dispatcher(() => InsertSorted(_onDemandMapModels, model, static item => item.Title));
+            await model.DownloadOnDemandMapAreaAsync().ConfigureAwait(false);
+        }
+
+        public string NextOnDemandAreaTitle()
+        {
+            static string MakeTitle(int index) => $"Area {index}";
```
mstefarov: Same here, this should probably be localizable

src/Toolkit/Toolkit/OfflineManager.cs
```diff
@@ -0,0 +1,605 @@
...
+                var base64Hash = Convert.ToBase64String(hash).Replace('/', '_').Replace('+', '-'); // filesystem-safe
+                var foldername = base64Hash[..16];
+               
+                _cacheFolder = Path.Combine(ArcGISRuntimeEnvironment.TempPath, foldername);
+            }
+            else
+                _cacheFolder = Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData);
+#endif
+
+            _synchronizationContext = SynchronizationContext.Current;
```
mstefarov: Design: Capturing SynchronizationContext in a singleton's constructor is a little risky. First one to touch this class gets to init the context for the lifetime of the process. Might be worth enforcing "only use this from the UI thread" here in the ctor, or at least documenting that this type should only be touched from the UI thread.

src/Toolkit/Toolkit/OfflineManager.cs
```diff
@@ -0,0 +1,605 @@
...
+        /// <summary>
+        /// Gets the completed job.
+        /// </summary>
+        public IJob Job { get; }
+    }
+
+    /// <summary>
+    /// The configuration properties for the <see cref="OfflineManager"/>.
+    /// </summary>
+    public sealed class OfflineManagerConfiguration
```
mstefarov, line 571: Design: OfflineManager's Configuration setter applies this once, but changing the configuration later will have no effect. Perhaps this should be immutable / have init-only properties?

src/Toolkit/Toolkit/OfflineManager.cs
```diff
@@ -0,0 +1,605 @@
...
+        /// <summary>
+        /// Gets or sets the root folder where offline maps are stored.
+        /// </summary>
+        /// <remarks>
+        /// By default for packaged apps this will be the <see cref="Environment.SpecialFolder.LocalApplicationData" /> folder, and for unpackaged apps this will be a temp folder unique to the <see cref="Environment.ProcessPath" />.
+        /// </remarks>
+        public string CacheFolder
+        {
+            get => _cacheFolder;
+            set
```
mstefarov, line 161: Design: Changing the CacheFolder while jobs are running would break things, wouldn't it?  Jobs would write to the old folder but OfflineMapInfos/models will read from the new folder.  Maybe it's worth "user-proofing" the setter by gating it with a "can't set while jobs running" check (or even a "can't set after a job was started" check).

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OnDemandMapModel.cs
```diff
@@ -0,0 +1,473 @@
...
+            Action<OnDemandMapModel> onRemoveDownload,
+            ICommand openMapCommand,
+            Action<Action> dispatcher) : base(dispatcher)
+        {
+            _areaId = areaId;
+            _portalItemId = portalItemId;
+            _title = title ?? string.Empty;
+            _thumbnailData = thumbnailData;
+            _onRemoveDownloadAction = onRemoveDownload;
+            _downloadCommand = new DelegateCommand((o) => _ = DownloadOnDemandMapAreaAsync(), () => false);
```
mstefarov: If I am reading this right, the DownloadCommand on IOfflineMapAreaItem is _uncallable_ (CanExecute always false) &mdash; is that right?  If this is supposed to be a placeholder / not implemented yet, let's document it as such.  Or remove it from the public API.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,531 @@
...
+                        area,
+                        portalItemId,
+                        area.PortalItem!.ItemId!,
+                        onRemoveDownload, openCommand, dispatcher))
+                    .ToList();
+
+                foreach (var model in models)
+                {
+                    await model.LoadAsync().ConfigureAwait(false);
+                }
```
mstefarov: Perf: Can we load these in parallel! [Parallel.ForEachAsync](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.parallel.foreachasync?view=net-8.0) would be perfect for this.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,531 @@
...
+            Action<PreplannedMapModel> onRemoveDownload, System.Windows.Input.ICommand openCommand, Action<Action> dispatcher)
+        {
+            var directory = OfflineMapAreaStorage.GetPreplannedAreasDirectory(portalItemId);
+            if (!Directory.Exists(directory))
+            {
+                return Array.Empty<PreplannedMapModel>();
+            }
+
+            var models = new List<PreplannedMapModel>();
+            foreach (var mapAreaDirectory in Directory.GetDirectories(directory))
```
mstefarov, line 491: Perf: Do you think _these_ can be loaded in parallel too?  Change `models` from a plain list to a ConcurrentBag and it should be thread-safe to put the whole loop body into Parallel.ForEachAsync.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,531 @@
...
+                    continue;
+                }
+
+                MobileMapPackage? mobileMapPackage = null;
+                try
+                {
+                    mobileMapPackage = new MobileMapPackage(mapAreaDirectory);
+                    await mobileMapPackage.LoadAsync().ConfigureAwait(false);
+
+                    var model = new PreplannedMapModel(
```
mstefarov, line 505: Perf: If I am understanding this code correctly, we end up creating and loading _two_ MobileMapPackages for each directory: first here in this loop, and second inside `PreplannedMapModel.LoadAsync()`. Feels inefficient!

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                                    AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasOpen}"
+                                    Command="{Binding OpenCommand}"
+                                    CommandParameter="{Binding Map}"
+                                    Content="{internal:LocalizedString Key=OfflineMapAreasOpen}" />
+                                <Button Margin="8,0,0,0" VerticalAlignment="Center"
+                                    AutomationProperties.HelpText="{Binding Error.Message, Mode=OneWay}"
+                                    ToolTipService.ToolTip="{Binding Error.Message, Mode=OneWay}"
+                                    Command="{Binding RemoveDownloadCommand, Mode=OneWay}" Visibility="{Binding Error, Converter={StaticResource OfflineMapAreasViewVisibilityConverter}, Mode=OneWay}">
+                                    <TextBlock FontFamily="{StaticResource ToolkitIconsFontFamily}" FontSize="16" Text="&#xE844;" />
+                                </Button>
```
SamhithaBurugupalli: Should this button be moved out of `stackpanel` as stackpanel contents are only visible when the download is successful. Also having a suitable Automation name like "Remove" or "Error occurred tap to remove" can be added.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
dotMorten: Isn't Task.WhenAll better?

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+
+                        <TextBlock
+                            Grid.Row="2" Grid.Column="1"
+                            Foreground="#FF666666"
+                            Text="{internal:LocalizedString Key=OfflineMapAreasDownloaded}" Visibility="{Binding IsDownloaded, Converter={StaticResource OfflineMapAreasViewVisibilityConverter}, Mode=OneWay}">
+                        </TextBlock>
+
+                        <Grid Grid.RowSpan="3" Grid.Column="2">
+                            <Button
+                                  AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasStop}"
```
SamhithaBurugupalli:
```suggestion
                                  AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasStop}"
```

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                                  Command="{Binding StopDownloadCommand}" Style="{StaticResource OfflineMapAreasViewIconButtonStyle}" 
+                                  ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasStop}" Visibility="{Binding IsDownloading, Converter={StaticResource OfflineMapAreasViewVisibilityConverter}, Mode=OneWay}">
+                                <TextBlock FontFamily="{StaticResource ToolkitIconsFontFamily}" FontSize="14" Text="&#xE7B3;" />
+                            </Button>
+
+                            <Button
+                                 AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasDownload}"
```
SamhithaBurugupalli:
```suggestion
                                 AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasDownload}"
```

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                                 Command="{Binding DownloadCommand}"
+                                 ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasDownload}"
+                                 Style="{StaticResource OfflineMapAreasViewIconButtonStyle}" 
+                                 Visibility="{Binding AllowsDownload, Converter={StaticResource OfflineMapAreasViewVisibilityConverter}, Mode=OneWay}">
+                                <TextBlock FontFamily="{StaticResource ToolkitIconsFontFamily}" FontSize="16" Text="&#xE81C;" />
+                            </Button>
+
+                            <StackPanel Orientation="Horizontal" Visibility="{Binding IsDownloaded, Converter={StaticResource OfflineMapAreasViewVisibilityConverter}, Mode=OneWay}">
+                                <Button
+                                    AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasRemove}"
```
SamhithaBurugupalli:
```suggestion
                                    AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasRemove}"
```

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
mstefarov: `Task.WhenAll(models.Select(...))` would work as well.  The only advantage to `ForEachAsync` might be limiting the degree of parallelism.

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                                        </Style>
+                                    </ProgressBar.Style>
+                                </ProgressBar>
+
+                                <Button
+                                    x:Name="AddMapAreaButton"
+                                    Grid.Row="1"
+                                    Margin="0,12,0,0"
+                                    HorizontalAlignment="Center"
+                                    AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasAddMapArea}">
```
SamhithaBurugupalli:
```suggestion
                                    AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasAddMapArea}">
```

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                        <Grid>
+                            <ScrollViewer VerticalScrollBarVisibility="{TemplateBinding VerticalScrollBarVisibility}" Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.IsAddOnDemandMode,Converter={StaticResource OfflineMapAreasViewVisibilityConverter}, ConverterParameter=Reverse}">
+                            <Grid>
+                                <Grid.RowDefinitions>
+                                    <RowDefinition />
+                                    <RowDefinition Height="Auto" />
+                                </Grid.RowDefinitions>
+
+                                <ItemsControl
+                                    x:Name="MapAreasView"
```
SamhithaBurugupalli, line 159:
```suggestion
                                    x:Name="MapAreasView"
                                    AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasMapAreasView}"
                                    IsTabStop="False"
```
There is currently a tab stop at the list of map areas, and the screen reader is reading out the type of it (Something like Esri.ArcGisRuntime.Toolkit. ...). As list view itself is not interactive we can disable the tabstop and set proper automation name something like "Map Areas View".  It should also be added to the localization

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                                    Grid.Row="0"
+                                    HorizontalAlignment="Stretch"
+                                    HorizontalContentAlignment="Stretch"
+                                    ItemsSource="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.MapAreas}"
+                                    ItemTemplate="{TemplateBinding ItemTemplate}"/>
```
SamhithaBurugupalli:
```suggestion
                                    ItemTemplate="{TemplateBinding ItemTemplate}">
                                        <ItemsControl.ItemContainerStyle>
                                            <Style TargetType="{x:Type ContentPresenter}">
                                                <Setter Property="AutomationProperties.Name" Value="{Binding Title}" />
                                                <Setter Property="AutomationProperties.HelpText" Value="{Binding Description}" />
                                            </Style>
                                        </ItemsControl.ItemContainerStyle>
                                    </ItemsControl>
```

Same for the list items, the screen reader is reading out the type of it.

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                            <TextBlock Text="{internal:LocalizedString Key=OfflineMapAreasSelectArea}" HorizontalAlignment="Center" VerticalAlignment="Center" FontWeight="Bold" />
+                            <Grid Grid.Row="1" Margin="0,8">
+                                <esri:MapView xmlns:esri="http://schemas.esri.com/arcgis/runtime/2013" ViewInsets="20" MinHeight="100"
+                                    x:Name="AddAreaMapView" Grid.Row="1" IsAttributionTextVisible="False" BorderBrush="Black" BorderThickness="1" />
+                                <Border BorderBrush="Black" BorderThickness="20" IsHitTestVisible="False" Opacity=".5" />
+                            </Grid>
+                            <TextBlock Grid.Row="2" Text="{internal:LocalizedString Key=OfflineMapAreasScaleLabel}" />
+                            <ComboBox Grid.Row="3" x:Name="AddOnDemandAreaScaleSelector" DisplayMemberPath="DisplayName" HorizontalAlignment="Stretch" />
+                            <TextBlock Grid.Row="4" Text="{internal:LocalizedString Key=OfflineMapAreasAreaNamePlaceholder}" />
+                            <TextBox Grid.Row="5" x:Name="AddOnDemandAreaNameTextBox" />
```
SamhithaBurugupalli:
```suggestion
                            <TextBlock x:Name="OfflineMapAreasScaleLabel" Grid.Row="2" Text="{internal:LocalizedString Key=OfflineMapAreasScaleLabel}" />
                            <ComboBox AutomationProperties.LabeledBy="{Binding ElementName=OfflineMapAreasScaleLabel}" Grid.Row="3" x:Name="AddOnDemandAreaScaleSelector" DisplayMemberPath="DisplayName" HorizontalAlignment="Stretch" />
                            <TextBlock x:Name="OfflineMapAreasAreaNamePlaceholder" Grid.Row="4" Text="{internal:LocalizedString Key=OfflineMapAreasAreaNamePlaceholder}" />
                            <TextBox Grid.Row="5" x:Name="AddOnDemandAreaNameTextBox" AutomationProperties.LabeledBy="{Binding ElementName=OfflineMapAreasAreaNamePlaceholder}"/>
```

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,428 @@
...
+                            <Button Content="{internal:LocalizedString Key=OfflineMapAreasCancel}" HorizontalAlignment="Right" x:Name="CancelAddOnDemandAreaButton" />
+
+                        </Grid>
+                        </Grid>
+                    </Border>
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>
+        <Setter Property="HorizontalAlignment" Value="Stretch" />
+        <Setter Property="VerticalAlignment" Value="Stretch" />
```
SamhithaBurugupalli, line 410:
```suggestion
        <Setter Property="VerticalAlignment" Value="Stretch" />
        <Setter Property="IsTabStop" Value="False" />
```
Extra tab stop at the control level

src/Samples/Toolkit.SampleApp.WinUI/Samples/Jobs/OfflineMapAreasSample.xaml
```diff
@@ -0,0 +1,56 @@
...
+                      SelectionChanged="OfflineMapSelection_Changed">
+                <ComboBox.ItemTemplate>
+                    <DataTemplate x:DataType="tkcommon:OfflineMapInfo">
+                        <StackPanel Orientation="Horizontal" Width="250">
+                            <Image Source="{x:Bind local:OfflineMapAreasSample.BytesToImage(ThumbnailData)}" MaxWidth="64" Height="64" />
+                            <TextBlock Text="{x:Bind Title}" TextTrimming="CharacterEllipsis" Style="{StaticResource BodyStrongTextBlockStyle}" />
+                        </StackPanel>
+                    </DataTemplate>
+                </ComboBox.ItemTemplate>
+            </ComboBox>
```
SamhithaBurugupalli, line 39: Is this part added intentionally? I couldn't understand the importance of this. Also, there is not any equivalent in WPF/ MaUI samples

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+        <Setter Property="VerticalAlignment" Value="Center" />
+        <Setter Property="HorizontalAlignment" Value="Center" />
+        <Setter Property="CornerRadius" Value="4" />
+    </Style>
+    <Style TargetType="controls:OfflineMapAreasView">
+        <!--  Default item template can be overriden by the user.  -->
+        <Setter Property="ItemTemplate">
+            <Setter.Value>
+                <DataTemplate x:DataType="controls:IOfflineMapAreaItem">
+                    <Grid Height="64" Margin="0,4" AutomationProperties.HelpText="{x:Bind Title, Mode=OneWay}" AutomationProperties.FullDescription="{x:Bind Description, Mode=OneWay}">
```
SamhithaBurugupalli:
```suggestion
                    <Grid Height="64" Margin="0,4" AutomationProperties.Name="{x:Bind Title, Mode=OneWay}" AutomationProperties.FullDescription="{x:Bind Description, Mode=OneWay}">
```

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+                        </Border>
+                        <TextBlock Grid.Row="0" Grid.Column="1" Text="{x:Bind Title, Mode=OneWay}" TextTrimming="CharacterEllipsis" Style="{StaticResource BodyStrongTextBlockStyle}" />
+                        <TextBlock Grid.Row="1" Grid.Column="1" Text="{x:Bind Description, Mode=OneWay}" TextTrimming="CharacterEllipsis" TextWrapping="Wrap" MaxLines="3" Style="{StaticResource CaptionTextBlockStyle}" />
+                        <TextBlock Text="{internal:LocalizedString Key=OfflineMapAreasDownloaded}" Grid.Row="2" Grid.Column="1" Visibility="{x:Bind IsDownloaded, Mode=OneWay}" Style="{StaticResource CaptionTextBlockStyle}" />
+                        <Grid Grid.Column="2" Grid.RowSpan="3">
+                            <Button Style="{StaticResource OfflineMapAreasViewButtonStyle}" AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasStop}" ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasStop}" Command="{x:Bind StopDownloadCommand, Mode=OneWay}" Visibility="{x:Bind IsDownloading, Mode=OneWay}">
```
SamhithaBurugupalli:
```suggestion
                            <Button Style="{StaticResource OfflineMapAreasViewButtonStyle}" AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasStop}" ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasStop}" Command="{x:Bind StopDownloadCommand, Mode=OneWay}" Visibility="{x:Bind IsDownloading, Mode=OneWay}">
```

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+                                <SymbolIcon Symbol="Stop" />
+                            </Button>
+                            <Button Style="{StaticResource OfflineMapAreasViewButtonStyle}" AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasDownload}" ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasDownload}" Command="{x:Bind DownloadCommand, Mode=OneWay}" Visibility="{x:Bind AllowsDownload, Mode=OneWay}">
```
SamhithaBurugupalli:
```suggestion
                            <Button Style="{StaticResource OfflineMapAreasViewButtonStyle}" AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasDownload}" ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasDownload}" Command="{x:Bind DownloadCommand, Mode=OneWay}" Visibility="{x:Bind AllowsDownload, Mode=OneWay}">
```

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+                                <SymbolIcon Symbol="Download" />
+                            </Button>
+                            <ProgressRing Width="40" Height="40" HorizontalAlignment="Center" VerticalAlignment="Center" IsHitTestVisible="False" Minimum="0" Maximum="1" IsIndeterminate="False" Value="{x:Bind DownloadProgress, Mode=OneWay}" Visibility="{x:Bind IsDownloading, Mode=OneWay}" />
+                            <StackPanel Orientation="Horizontal" Visibility="{x:Bind IsDownloaded, Mode=OneWay}" >
+                                <Button Style="{StaticResource OfflineMapAreasViewButtonStyle}" ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasRemove}" AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasRemove}" Command="{x:Bind RemoveDownloadCommand, Mode=OneWay}">
```
SamhithaBurugupalli:
```suggestion
                                <Button Style="{StaticResource OfflineMapAreasViewButtonStyle}" ToolTipService.ToolTip="{internal:LocalizedString Key=OfflineMapAreasRemove}" AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasRemove}" Command="{x:Bind RemoveDownloadCommand, Mode=OneWay}">
```

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+                                    <SymbolIcon Symbol="Delete" />
+                                </Button>
+                                <Button Content="{internal:LocalizedString Key=OfflineMapAreasOpen}" AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasOpen}" Margin="8,0,0,0" Command="{x:Bind OpenCommand, Mode=OneWay}" CommandParameter="{x:Bind Map, Mode=OneWay}" />
+                            </StackPanel>
+                            <Button Style="{StaticResource OfflineMapAreasViewButtonStyle}" AutomationProperties.HelpText="{x:Bind Error.Message, Mode=OneWay}" ToolTipService.ToolTip="{x:Bind Error.Message, Mode=OneWay}"
```
SamhithaBurugupalli: Automation name to be set on button, something like "Remove" or "Error occurred tap to remove"

src/Toolkit/Toolkit/OfflineManager.cs
dotMorten: Pending tasks won't have any data (it'll be in the pending folder at this point).

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
dotMorten: While I see your point, these will be quick calls on a limited set of areas, and such a change would have to deal with order of results being different each time and then keeping that constant. For simplicity I'll keep it the way it is currently, since this would never really be a slow set of tasks.

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+                                                <TextBlock Text="{internal:LocalizedString Key=OfflineMapAreasRefresh}"/>
+                                            </StackPanel>
+                                        </Button>
+
+                                    </StackPanel>
+
+                                    <ProgressRing Visibility="{x:Bind TemplateSettings.IsLoadingModels, Mode=OneWay}" IsHitTestVisible="False" />
+
+                                    <Grid Grid.Row="1" Visibility="{x:Bind local:OfflineMapAreasViewResources.FalseToVisible(TemplateSettings.IsInternetNotAvailable), Mode=OneWay}" >
+                                        <Button HorizontalAlignment="Center" Visibility="{x:Bind TemplateSettings.IsOnDemandMode, Mode=OneWay}" AutomationProperties.HelpText="{internal:LocalizedString Key=OfflineMapAreasAddMapArea}" x:Name="AddMapAreaButton" Margin="4" Padding="24,4">
```
SamhithaBurugupalli:
```suggestion
                                        <Button HorizontalAlignment="Center" Visibility="{x:Bind TemplateSettings.IsOnDemandMode, Mode=OneWay}" AutomationProperties.Name="{internal:LocalizedString Key=OfflineMapAreasAddMapArea}" x:Name="AddMapAreaButton" Margin="4" Padding="24,4">
```

src/Toolkit/Toolkit/OfflineManager.cs
dotMorten: I've made the background schedule one init (and ios only) - the rest will take effect, since they are pulled dynamically when a job is created.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapAreasView.Maui.cs
```diff
@@ -0,0 +1,563 @@
...
+            {
+                WidthRequest = 32,
+                HeightRequest = 32,
+                BackgroundColor = Colors.Transparent,
+                BorderWidth = 0,
+                Padding = 0,
+            };
+            if (!string.IsNullOrEmpty(hint))
+            {
+                SemanticProperties.SetHint(button, hint);
```
SamhithaBurugupalli:
```suggestion
                SemanticProperties.SetDescription(button, hint);
```

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapAreasView.Maui.cs
```diff
@@ -0,0 +1,563 @@
...
+            progressBar.SetBinding(IsVisibleProperty, static (IOfflineMapAreaItem item) => item.IsDownloading);
+            progressBar.SetBinding(ProgressBar.ProgressProperty, static (IOfflineMapAreaItem item) => item.DownloadProgress);
+            Grid.SetColumn(progressBar, 1);
+            Grid.SetColumnSpan(progressBar, 2);
+            Grid.SetRow(progressBar, 2);
+            root.Children.Add(progressBar);
+
+            CalciteImageButton failedButton = CreateIconButton(ToolkitIcons.ExclamationMarkCircle, null);
+            failedButton.SetBinding(ToolTipProperties.TextProperty, static (IOfflineMapAreaItem item) => item.Error?.Message);
+            failedButton.SetBinding(SemanticProperties.DescriptionProperty, static (IOfflineMapAreaItem item) => item.Error?.Message);
```
SamhithaBurugupalli:
```suggestion
            failedButton.SetBinding(SemanticProperties.HintProperty, static (IOfflineMapAreaItem item) => item.Error?.Message);
```

Ideally Automation name should be short, can use hint for conveying the error message. DescriptionProperty should be something like "Remove" or "Failed to downlaod"

src/Toolkit/Toolkit/OfflineManager.cs
dotMorten: Removed the sync entirely. Instead the VMs which does have access to the dispatcher will ensure it uses the APIs that rely on INPC to do so on the UI Thread

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+                            <SolidColorBrush x:Key="ButtonBackgroundPointerOver" Color="{ThemeResource ControlFillColorDefault}" />
+                        </Border.Resources>
+                        <Grid>
+                            <ScrollViewer VerticalScrollBarVisibility="{TemplateBinding VerticalScrollBarVisibility}" Visibility="{x:Bind local:OfflineMapAreasViewResources.FalseToVisible(TemplateSettings.IsAddOnDemandMode), Mode=OneWay}">
+                                <Grid>
+                                    <Grid.RowDefinitions>
+                                        <RowDefinition/>
+                                        <RowDefinition Height="Auto" />
+                                    </Grid.RowDefinitions>
+                                    <ItemsControl x:Name="MapAreasView" HorizontalAlignment="Stretch" HorizontalContentAlignment="Stretch"
```
SamhithaBurugupalli:
```suggestion
                                    <ItemsControl x:Name="MapAreasView" HorizontalAlignment="Stretch" HorizontalContentAlignment="Stretch" IsTabStop="False"

```

src/Toolkit/Toolkit.WinUI/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,177 @@
...
+                                <Button Content="{internal:LocalizedString Key=OfflineMapAreasCancel}" HorizontalAlignment="Right" x:Name="CancelAddOnDemandAreaButton" />
+
+                            </Grid>
+                        </Grid>
+                    </Border>
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>
+        <Setter Property="HorizontalAlignment" Value="Stretch" />
+        <Setter Property="VerticalAlignment" Value="Stretch" />
```
SamhithaBurugupalli, line 185:
```suggestion
        <Setter Property="VerticalAlignment" Value="Stretch" />
        <Setter Property="IsTabStop" Value="False" />
```

src/Tests/Toolkit.Tests/OfflineMapManagerTests.cs
```diff
@@ -0,0 +1,282 @@
...
+                Assert.AreEqual(PreplannedMapModelStatus.Packaged, vm.PreplannedMapModels[0].Status);
+                Assert.IsGreaterThan(0, vm.PreplannedMapModels[0].SizeInBytes);
+                await vm.PreplannedMapModels[0].DownloadPreplannedMapAreaAsync();
+                Assert.AreEqual(PreplannedMapModelStatus.Downloading, vm.PreplannedMapModels[0].Status);
+                Assert.IsNotNull(vm.PreplannedMapModels[0].Job);
+                await AwaitDownload(vm.PreplannedMapModels[0]);
+                if (vm.PreplannedMapModels[0].Error is not null)
+                {
+                    System.Diagnostics.Debugger.Break();
+                }
```
mstefarov: Did you mean to leave this `Debugger.Break()` in the test code? Might be a problem in CI, halting the test run instead of just failing the test.

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,401 @@
...
+                                <RowDefinition MaxHeight="400" />
+                                <RowDefinition Height="Auto" />
+                                <RowDefinition Height="Auto" />
+                                    <RowDefinition Height="Auto" />
+                                    <RowDefinition Height="Auto" />
+                                    <RowDefinition Height="Auto" />
+                            </Grid.RowDefinitions>
+                            <TextBlock Text="{internal:LocalizedString Key=OfflineMapAreasSelectArea}" HorizontalAlignment="Center" VerticalAlignment="Center" FontWeight="Bold" />
+                            <Grid Grid.Row="1" Margin="0,8">
+                                <esri:MapView xmlns:esri="http://schemas.esri.com/arcgis/runtime/2013" ViewInsets="20" MinHeight="100"
```
mstefarov, line 392: Nit: This is a bit unconventional, defining `xmlns:esri` down on a MapView itself. It would be more consistent to declare at the top, like all other xamls.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapViewModel.cs
```diff
@@ -0,0 +1,445 @@
...
+        }
+    }
+
+    internal abstract class OfflineBindableObject : INotifyPropertyChanged
+    {
+        public event PropertyChangedEventHandler? PropertyChanged;
+        
+        protected Action<Action> Dispatcher { get; }
+
+        protected Task DispatchAsync(Task task)
```
mstefarov: I think we can get rid of this DispatchAsync impl.

It doesn't really do what the name implies (the `task` is already started by the time DispatchAsync sees it) -- it just hops to UI thread to _attach_ a ContinueWith, which isn't useful in itself.

Both callers (PreplannedMapModel and OnDemandMapModel) are invoked from XAML command handlers, so they're already on the UI thread. So even if it worked right, it wouldn't be necessary.

src/Toolkit/Toolkit/OfflineManager.cs
```diff
@@ -0,0 +1,672 @@
...
+            {
+                ArgumentNullException.ThrowIfNullOrWhiteSpace(value, nameof(value));
+                if (_cacheFolder != value)
+                {
+                    if (_observedJobs.Count > 0)
```
mstefarov: HashSet<T>.Count is not thread-safe. I would take _observedJobsLock here.

Actually, I think CacheFolder property as designed doesn't work if a prior session left paused jobs.
Resolving OfflineManager.Shared triggers the constructor, which calls InitManager() against the default folder. So _offlineMapInfos is populated and any persisted jobs in _jobManager.Jobs are observed against paths baked under the default folder before the setter is reachable.

Now, if we keep this check, the setter will throw. But even if we remove the check, it still won't work correctly. Setting CacheFolder re-runs InitManager against the new folder, but the loaded jobs still write to the old folder when they resume. Their completion runs HandlePendingMapInfo against the new folder, finds nothing, and silently fails to register. And the mmpk ends up orphaned under the old folder.

If a configurable cache folder is a real requirement, the API needs a different shape. Currently it only behaves well if there are no completed downloads or paused jobs from a prior session to be orphaned.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapViewModel.cs
```diff
@@ -0,0 +1,445 @@
...
+            if (onlineMap is null)
+            {
+                throw new ArgumentNullException(nameof(onlineMap));
+            }
+
+            string? itemId = onlineMap.Item?.ItemId;
+            if (string.IsNullOrWhiteSpace(itemId) && onlineMap.Uri is not null)
+            { 
+                // If map isn't loaded yet, look for Itme ID parameter in URL and extract it:
+                itemId = System.Web.HttpUtility.ParseQueryString(onlineMap.Uri.Query)["id"];
```
mstefarov: This seems a bit fragile. What if the URL is `.../sharing/rest/content/items/XYZ` or `/apps/mapviewer/index.html?webmap=XYZ` etc? We [support quite a few URL patterns](https://developers.arcgis.com/net/api-reference/api/netwin/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Mapping.Map.-ctor.html#Esri_ArcGISRuntime_Mapping_Map__ctor_System_Uri_) on Map.

src/Samples/Toolkit.SampleApp.WinUI/Samples/Jobs/OfflineMapAreasSample.xaml
dotMorten: Yes this was to test the offline workflow where you start out offline but already have data offline you can start with. I just didn't implement this specifically in the other platforms, mostly due to screen space issues on MAUI mobile devices, and also wouldn't add a lot of functionality testing, since that's all shared code across all platforms.

src/Toolkit/Toolkit/OfflineManager.cs
dotMorten: The doc clearly state to set this upfront before queuing any jobs. I don't see you ever really changing this on-the-fly during a running app as a real scenario, and not something that is intended to be supported, so I don't think it's worth the added complexity to support a scenario that likely isn't realistic.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapViewModel.cs
dotMorten: Good call. Since we don't have unloaded portal items and we need the item id, we'll need way to do this. I'm ok with it works in most cases, and you can always fallback to using a map with a loaded portal item.

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
dotMorten: I move this to the item template instead. WinUI already had the right properties set there, but WPF was setting HintText instead of Name

dotMorten: I didn't see the above of my approach to have any effect. Instead improve the narration for each button to specify which item the button acts on.

src/Toolkit/Toolkit.WPF/UI/Controls/OfflineMapAreasView/OfflineMapAreasView.Theme.xaml
```diff
@@ -0,0 +1,410 @@
...
+    <Style TargetType="{x:Type controls:OfflineMapAreasView}">
+        <!--  Default item template can be overriden by the user.  -->
+        <Setter Property="ItemTemplate">
+            <Setter.Value>
+                <DataTemplate>
+                    <Grid
+                        Height="64"
+                        Margin="0,4"
+                        AutomationProperties.Name="{Binding Title}"
+                        AutomationProperties.HelpText="{Binding Decription}">
```
mstefarov: Typo in a binding

```suggestion
                        AutomationProperties.HelpText="{Binding Description}">
```

src/Toolkit/Toolkit/OfflineManager.cs
mstefarov: I agree that changing configuration mid-session is not worth supporting. BUT even the documented "set upfront before queuing any jobs" path might not work when there's state from a previous session.

I see no way to use the CacheFolder setter without triggering `InitManager()` against the default folder first. Just referencing `OfflineManager.Shared` triggers `InitManager`, which populates `_observedJobs` and `_offlineMapInfos`.

Maybe I am totally misunderstanding the code, but it's worth testing "set CacheFolder on startup when there's previous state" scenario.

src/Toolkit/Toolkit/OfflineManager.cs
```diff
@@ -0,0 +1,685 @@
...
+        private void LoadOfflineMapInfos()
+        {
+            _offlineMapInfos.Clear();
+            var offlineManagerDirectory = GetOfflineManagerDirectory();
+            if (!Directory.Exists(offlineManagerDirectory))
+            {
+                return;
+            }
+
+            foreach (var file in Directory.GetFiles(offlineManagerDirectory, "*.json"))
```
mstefarov, line 441: `LoadOfflineMapInfos()` scans the cache folder for `*.json` and parses _every match_ as a map-info file. But the JobManager state file (jobs.json) lives in that same folder, so it gets parsed and throws. Since this runs from the OfflineManager.Shared static initializer, the whole type faults for the rest of the process.

To repro on Windows: start a map-area download and relaunch the app while a job is still persisted (the resume-after-restart functionality). On next launch the first use of OfflineManager.Shared throws TypeInitializationException and it re-throws on every subsequent launch until the cache is wiped. Existing tests don't catch it because a fresh process has no `jobs.json`

This can be fixed by keeping the job state file out of the scanned directory or skipping the jobs file specifically. Adding a try-catch around each FromFile call might also be a good idea, so one bad file doesn't ruin the process.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,517 @@
...
+        {
+            try
+            {
+                var result = (DownloadPreplannedOfflineMapResult)await job.GetResultAsync().ConfigureAwait(false);
+                Error = null;
+                if (result.MobileMapPackage is not null)
+                {
+                    await LoadAndUpdateMobileMapPackageAsync(result.MobileMapPackage).ConfigureAwait(false);
+                }
+                Status = PreplannedMapModelStatus.Downloaded;
```
mstefarov: `LoadAndUpdateMobileMapPackageAsync` already sets status to _either_ "Downloaded" or "MmpkLoadFailure". We shouldn't unconditionally set it back to "Downloaded".

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OnDemandMapModel.cs
```diff
@@ -0,0 +1,475 @@
...
+            ICommand openMapCommand,
+            Action<Action> dispatcher) : base(dispatcher)
+        {
+            _areaId = areaId;
+            _portalItemId = portalItemId;
+            _title = title ?? string.Empty;
+            _thumbnailData = thumbnailData;
+            _onRemoveDownloadAction = onRemoveDownload;
+            _downloadCommand = new DelegateCommand((o) => { }, () => false); // Currently doesn't allow re-download
+            _removeDownloadCommand = new DelegateCommand((o) => RemoveDownloadedArea(), () => IsDownloaded || Error != null);
```
mstefarov: This allows the currently opened map to be deleted, which attempts to close and delete the MMPK and might break something.  I accidentally did this in our sample and got spammed with

> LayerViewState Error: ArcGISVectorTiledLayer: File error.

The common-toolkit design suggests disabling the remove button for the currently-opened map. Let's do that.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,517 @@
...
+            Action<Action> dispatcher): base(dispatcher)
+        {
+            _offlineMapTaskFactory = offlineMapTaskFactory;
+            _portalItemId = portalItemId;
+            _preplannedMapAreaId = preplannedMapAreaId;
+            _title = title;
+            _description = HtmlUtility.StripHtml(description);
+            _supportsRedownloading = supportsRedownloading;
+            _downloadCommand = new DelegateCommand((o) => _ = DownloadPreplannedMapAreaAsync(), () => _preplannedMapArea != null && AllowsDownload && Status != PreplannedMapModelStatus.Downloading);
+            _removeDownloadCommand = new DelegateCommand((o) => RemoveDownloadedArea(), () => IsDownloaded);
```
mstefarov: Same here. The command allows currently-opened map to be removed.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OnDemandMapModel.cs
```diff
@@ -0,0 +1,475 @@
...
+            _ = ObserveJobAsync(job);
+        }
+
+        private async Task ObserveJobAsync(GenerateOfflineMapJob job)
+        {
+            try
+            {
+                var result = (GenerateOfflineMapResult)await job.GetResultAsync().ConfigureAwait(false);
+                Error = null;
+                Status = OnDemandMapModelStatus.Downloaded;
```
mstefarov: Similar to the Status problem in the other model. This unconditionally sets the Downloaded status before MMPK actually loads, so it might flip to "MmpkLoadFailure". 

Let that method take care of the status changes when MobileMapPackage is not null.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapViewModel.cs
```diff
@@ -0,0 +1,433 @@
...
+                }
+            });
+            await tcs.Task.ConfigureAwait(false);
+        }
+
+        private async Task LoadOnDemandMapModelsAsync()
+        {
+            var models = await OnDemandMapModel.LoadOnDemandMapModelsAsync(PortalItemId, OnRemoveDownloadOfOnDemandArea, _openMapCommand, Dispatcher).ConfigureAwait(false);
+            Dispatcher(() => ReplaceCollection(_onDemandMapModels, models));
+        }
```
mstefarov, line 303: One of these methods has a TCS to wait for ReplaceCollection to finish and handles errors, but the other just fires-and-forgets.  Is there a reason?

I'm guessing DetermineModeAsync should only run after ReplaceCollection completed. So LoadOnDemandMapModelsAsync should probably wait too.

mstefarov (commented):
The trying-to-open-`jobs.json` bug is the only big one left. We're almost there!

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OnDemandMapModel.cs
```diff
@@ -0,0 +1,492 @@
...
+            Status = OnDemandMapModelStatus.Downloading;
+            Error = null;
+
+            try
+            {
+                var offlineMapTask = await _offlineMapTaskFactory().ConfigureAwait(false);
+                var parameters = await offlineMapTask.CreateDefaultGenerateOfflineMapParametersAsync(
+                    _configuration.AreaOfInterest,
+                    _configuration.MinScale,
+                    _configuration.MaxScale).ConfigureAwait(false);
```
mstefarov: We actually _want_ to stay on the UI thread here, so that `OfflineManager.StartJobAsync` gets called from the correct thread, and `OfflineMapInfos` collection ends up updated from the correct thread, etc. The WinUI sample actually crashes on me in the on-demand workflow because the combobox gets updated off-thread.

Can be fixed by either removing `ConfigureAwait(false)` or by explicitly calling `ConfigureAwait(true)` (so that we are not tempted to accidentally add it later).

```suggestion
                var offlineMapTask = await _offlineMapTaskFactory().ConfigureAwait(true);
                var parameters = await offlineMapTask.CreateDefaultGenerateOfflineMapParametersAsync(
                    _configuration.AreaOfInterest,
                    _configuration.MinScale,
                    _configuration.MaxScale).ConfigureAwait(true);
```

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/PreplannedMapModel.cs
```diff
@@ -0,0 +1,531 @@
...
+            Action<Action> dispatcher): base(dispatcher)
+        {
+            _offlineMapTaskFactory = offlineMapTaskFactory;
+            _portalItemId = portalItemId;
+            _preplannedMapAreaId = preplannedMapAreaId;
+            _title = title;
+            _description = HtmlUtility.StripHtml(description);
+            _supportsRedownloading = supportsRedownloading;
+            _downloadCommand = new DelegateCommand((o) => _ = DownloadPreplannedMapAreaAsync(), () => _preplannedMapArea != null && AllowsDownload && Status != PreplannedMapModelStatus.Downloading);
+            _removeDownloadCommand = new DelegateCommand((o) => RemoveDownloadedArea(), () => IsDownloaded && !IsOpen);
```
mstefarov: If a preplanned download fails, the area gets stuck at `Status=DownloadFailure`. The error/remove button is displayed (`Error != null` is true) but its command remains disabled (`IsDownloaded && !IsOpen` is false). The download button is hidden.  So there's no way to clear out the failed area without restarting.

Image

OnDemandMapModel already includes `|| Error != null` in the same predicate, let's do the same here.

```suggestion
            _removeDownloadCommand = new DelegateCommand((o) => RemoveDownloadedArea(), () => IsDownloaded && !IsOpen || Error != null));
```

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapAreasView.cs
```diff
@@ -0,0 +1,550 @@
...
+                if (GetTemplateChild("AddAreaMapView") is MapView mv)
+                {
+                    var vp = mv.GetCurrentViewpoint(ViewpointType.BoundingGeometry);
+                    if (vp != null)
+                    {
+                        try
+                        {
+                            // Subtract the 40px dark buffer around the view.
+                            double buffer = mv.UnitsPerPixel * mv.ViewInsets.Left; // Assume insets are equidistant
+                            var clippedArea = GeometryEngine.Buffer(vp.TargetGeometry, -buffer);
```
mstefarov: I was going to point out that on WinUI the shaded area is too wide (40px vs 20px on all other platforms):

Image

But then I realized that all platforms adjust for insets _twice_ in this method, so WinUI is actually the only accurate one.

`ViewInsets` already reserves the dark-frame margin, and `GetCurrentViewpoint(BoundingGeometry)` already returns the inset-reduced region (see our `MapView_Insets_Tests.TestMapViewInsets_ExtentTests`). After subtracting buffer, the exported area ends up smaller than the selection rectangle that the user sees.

You should be able to export `vp.TargetGeometry` directly, without the Buffer. And shrink the WinUI border. Then all platforms will display the selected area accurately.

src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OnDemandMapModel.cs
```diff
@@ -0,0 +1,492 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                if (OfflineMapAreaUtilities.IsCancellation(job, ex))
+                {
+                    Error = null;
+                    Status = OnDemandMapModelStatus.DownloadCancelled;
+                    RemoveDownloadedArea();
+                }
```
mstefarov, line 354: Cancelling an in-progress on-demand download (Stop) removes the whole row. This is a but different from common-toolkit design: it shows a cancelled area listed in DownloadCancelled status with a button to remove it.

The cancel path sets status to "DownloadCancelled" but then immediately calls `RemoveDownloadedArea()`, which overwrites it with "Initialized" and drops the row.  So user never gets to observe a DownloadCancelled status.

If removal-on-cancel is intended, the DownloadCancelled assignment is dead. If not, leave the row in DownloadCancelled status and something in the template to display it with a button.

mstefarov (approved):

PR #759, opened 2026-06-01
UITests: Android CI Tests

imalcolm1:
Scripts for running CI tests for maui android.

## Notes for reviewers
**Structure**
`cibuild.sh` is the entry that makes sure dotnet is installed so `cibuild.cs` can be run. `cibuild.cs` is intended to be semi-cross-platform, though for now it assumes it is being run on Mac.

**Changes to test runners/apps**
Like in the windows PR, the android test runner can now be configured with an environment variable for the apk path. This PR also includes a fix for the device dpi metadata not being recognized by appium properly.

## Testing Locally
To test locally run the following commands in zsh. The NUGET_REPO directory should contain the nuget package from a daily build or a release, otherwise leave it unset.
```bash
export WORKSPACE=$(realpath ~/Developer)/uitest-workspace
mkdir $WORKSPACE
cd $WORKSPACE
git clone https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit.git -b ian/uitests-cibuild-android
export RELEASE_VERSION=300.1.0
export NUGET_REPO=${WORKSPACE}/nuget-source # Nuget files must be placed here

# Once the nuget repo has been copied, run this in the same terminal
${WORKSPACE}/arcgis-maps-sdk-dotnet-toolkit/src/Tests/UITests/cibuild/cibuild.sh MauiAndroid
```


src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -0,0 +1,377 @@
...
+            dependencies.AppiumEntry,
+            appiumStandardOutput,
+            appiumStandardError
+        );
+
+        // Configure appium cleanup
+        Action cleanup = () => {
+            Console.WriteLine("\nStarting test cleanup...");
+
+            appiumProcess.Kill();
```
mstefarov: Two issues with shutting Appium down here:

- `appiumProcess.Kill()` stops only the Appium process itself, not the other programs it launches (the Android driver, adb). Those keep running and can interfere with the next run. Use `Kill(entireProcessTree: true)`.
- The log lists are still being filled in the background while this code reads them, so we might get a "Collection was modified" exception or miss the last few lines. Wait for the process to fully stop before reading:

```suggestion
            appiumProcess.Kill(entireProcessTree: true);
            appiumProcess.WaitForExit();
```

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -0,0 +1,377 @@
...
+            var nodeUrl = $"https://nodejs.org/dist/v{nodeVersion}/node-v{nodeVersion}-darwin-arm64.tar.gz";
+            var nodeTarGz = Path.Combine(workspace, "node.tar.gz");
+
+            using var httpClient = new HttpClient();
+            using var response = httpClient.GetAsync(nodeUrl).GetAwaiter().GetResult();
+            response.EnsureSuccessStatusCode();
+
+            using (var output = File.Create(nodeTarGz))
+            {
+                response.Content.CopyToAsync(output).GetAwaiter().GetResult();
```
mstefarov: Calling GetAwaiter/Wait/Result on tasks is _almost always_ a bad idea. Check out [David Fowler’s Async Guide](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#avoid-using-taskresult-and-taskwait) for a good explanation why, with examples. It's a handy reference to bookmark.

In this case it's avoidable: you can change the `Main` method to be `async Task<int>`. It works!

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -0,0 +1,377 @@
...
+        BuildSettings buildSettings = testPlatform switch
+        {
+            "MauiAndroid" => SetupAndroid(dependencies, nodeWorkspace, toolkitSrc, workspace),
+            _ => throw new ArgumentException($"The test platform '{testPlatform}' was not recognized. Aborting tests.")
+        };
+
+        // Run appium in background
+        var appiumStandardOutput = new List<string>();
+        var appiumStandardError = new List<string>();
+        var appiumProcess = RunBinaryBackground(
```
mstefarov: Same feedback I had on the Windows PR: After starting Appium the code goes straight into building and running. You could port the appium-status check from powershell to C#:

```cs
private static async Task WaitForAppiumReadyAsync(string statusUrl, TimeSpan timeout)
{
    using var http = new HttpClient { Timeout = TimeSpan.FromSeconds(2) };
    var sw = Stopwatch.StartNew();
    while (sw.Elapsed < timeout)
    {
        try
        {
            var body = await http.GetStringAsync(statusUrl);
            if (JsonNode.Parse(body)?["value"]?["ready"]?.GetValue<bool>() == true)
                return;
        }
        catch { /* not up / not ready yet */ }
        await Task.Delay(500);
    }
    throw new Exception($"Appium did not become ready within {timeout.TotalSeconds:0}s.");
}
```

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -0,0 +1,377 @@
...
+
+        // Install node
+        var nodeWorkspace = Path.Join(workspace, ".node");
+        var nodeVersion = ReadYamlValue(yamlConfig, "node-version");
+        (dependencies.NodeExe, dependencies.NpmExe) = InstallNode(nodeWorkspace, nodeVersion);
+        dependencies.AppiumEntry = Path.Join(nodeWorkspace, "node_modules", "appium", "index.js");
+
+        // Install appium
+        Environment.SetEnvironmentVariable("APPIUM_HOME", Path.Join(workspace, ".appium"));
+        RunBinary(dependencies.NpmExe, $"install appium --prefix \"{nodeWorkspace}\"");
```
mstefarov: Totally optional: If you run into problems with escaping/quoting command-line parameters, consider having RunBinary use [ProcessStartInfo.ArgumentList](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.argumentlist?view=net-10.0) (which quotes and combines individual args for you) as an alternative to building a single Arguments string. So in this example it could be...

```cs
RunBinary(dependencies.NpmExe, ["install", "appium", "--prefix", nodeWorkspace]);
```

You don't have to convert everything, I just want to point out that this is an available option.

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -0,0 +1,377 @@
...
+        using (StreamReader sr = File.OpenText(yamlFile)) {
+            string? s;
+            while ((s = sr.ReadLine()) != null) {
+                s = s.Trim();
+                if (s.StartsWith(name)) {
+                    return s.Substring(name.Length+3, s.Length-name.Length-4);
+                }
+            }
+        }
+        throw new Exception($"Could find variable ${name} in ${yamlFile}");
```
mstefarov:
```suggestion
        throw new Exception($"Could not find variable {name} in {yamlFile}");
```

src/Tests/UITests/cibuild/cibuild.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+test_platform=$1
+if [ -z "${test_platform}" ]; then
+  echo "Error: No test platform argument was provided to cibuild.sh, the build will fail." 1>&2
+fi
```
mstefarov: When no platform is given, this prints an error but keeps going, so the run continues and fails later in a less obvious place. Add `exit 1` right after the message, like you already have elsewhere.

imalcolm1: My thinking there was that`cibuild.cs` has the properly-descriptive error messages that will lead to a failure. I'll probably just remove this particular echo entirely since it's really just noise

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -0,0 +1,454 @@
...
+        var stdErr = string.Empty;
+        process.OutputDataReceived += (sender, e) => {
+            if (!string.IsNullOrEmpty(e.Data)) {
+                if (captureOutput) { stdOut += e.Data + "\n"; }
+                else { Console.WriteLine(e.Data); }
+            }
+        };
+        process.ErrorDataReceived += (sender, e) => {
+            if (!string.IsNullOrEmpty(e.Data)) {
+                if (captureOutput) { stdOut += e.Data + "\n"; }
```
mstefarov:
```suggestion
        process.ErrorDataReceived += (sender, e) => {
            if (!string.IsNullOrEmpty(e.Data)) {
                if (captureOutput) { stdErr += e.Data + "\n"; }
```

If you ever have to deal with long output, you might want to switch from strings to StringBuilders inside RunBinary.  It's much cheaper to concatenate.

src/Tests/UITests/Toolkit.UITests.MauiAndroid/AppiumSetup.MauiAndroid.cs
```diff
@@ -11,7 +11,15 @@ public static void AssemblyInitialize(TestContext testContext)
     {
         var settings = GetBuildSettings();
 
-        if (settings["usePreinstalledApp"] == "true")
+        if (!String.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("TKUITEST_APP")))
+        {
+            // If TKUITEST_APP is set it is assumed to be Release apk, and will be fully uninstalled an reinstalled each run
```
mstefarov:
```suggestion
            // If TKUITEST_APP is set it is assumed to be Release apk, and will be fully uninstalled and reinstalled each run
```

mstefarov (approved):
Couple minor comments, but no blockers. Well done!

src/Tests/UITests/cibuild/cibuild.cs
imalcolm1: Thanks for the catch! The mac test has to process a fairly long output at one point so I've added the StringBuilder suggestion too

SamhithaBurugupalli (approved):

PR #760, opened 2026-06-01
UITests: Windows CI Build

imalcolm1:
The commit history got corrupted in the [last PR](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/755), hence this copy. Changes are exactly the same.

Scripts for running CI tests for windows. All of the scripts are running on the Jenkins VMs when RDP is connected. Because of this tests are expected to fail until we get a physical machine to test on.

## Notes for reviewers
**Structure**
Most of the logic is stored in the `utils.psm1` module. The 3 platform-specific scripts just call `Invoke-WindowsUITests` with different parameters.

**Changes to test runners**
To simplify the ci logic I added the ability to configure app paths for the runner projects at runtime using environment variables. This has been documented in the UITest readme.

## Testing Locally
To test locally run the following commands in powershell. The NUGET_REPO directory should contain the nuget package from a daily build or a release, otherwise leave it unset.
```powershell
mkdir C:\uitest-workspace\
cd C:\uitest-workspace\
git clone https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit.git -b ian/uitests-cibuild-wpf-clean
$env:WORKSPACE = 'C:\uitest-workspace\'
$env:RELEASE_VERSION = '300.1.0' # This may need to change if NUGET_REPO used packages from a different version
$env:NUGET_REPO = 'C:\uitest-workspace\nuget-source' # This requires first copying nuget files to the given folder
& C:\uitest-workspace\arcgis-maps-sdk-dotnet-toolkit\src\Tests\UITests\cibuild\cibuild-wpf.ps1 
```

### DIFF
--- a/src/Tests/UITests/cibuild/utils.psm1
+++ b/src/Tests/UITests/cibuild/utils.psm1
@@ -23,7 +23,7 @@ function Install-Dotnet {
 
   if ([string]::IsNullOrWhiteSpace($dotnet_cache_folder) -or !(Test-Path -Path $dotnet_exe)) {
     $installerPath = Join-Path $workspace 'dotnet-install.ps1'
-    curl -L https://dot.net/v1/dotnet-install.ps1 -o $installerPath
+    & curl.exe -L https://dot.net/v1/dotnet-install.ps1 -o $installerPath
     & $installerPath -Version $dotnet_version -InstallDir $dotnet_install_folder -NoPath
     if ($LASTEXITCODE -ne 0) {
      exit 1
@@ -56,7 +56,7 @@ function Install-Nodejs {
     $node_url = "https://nodejs.org/dist/v$($node_version)/node-v$($node_version)-win-x64.zip"
     $node_zip = Join-Path $env:WORKSPACE "node.zip"
 
-    curl $node_url -o $node_zip
+    & curl.exe -L $node_url -o $node_zip
     Expand-Archive -Path $node_zip -Destination $env:WORKSPACE
 
     if (!($?) -or ($LASTEXITCODE -ne 0)) {
--- a/src/Tests/UITests/cibuild/utils.psm1
+++ b/src/Tests/UITests/cibuild/utils.psm1
@@ -8,11 +8,11 @@ function Invoke-WindowsUITests {
 
     [Parameter(Mandatory)]
     [ValidateNotNullOrEmpty()]
-    [string]$runner_project,
+    [string]$runner_name,
 
     [Parameter(Mandatory)]
     [ValidateNotNullOrEmpty()]
-    [string]$app_project,
+    [string]$app_name,
 
     [Parameter(Mandatory)]
     [string[]]$build_parameters,
@@ -25,6 +25,8 @@ function Invoke-WindowsUITests {
   )
 
   $config_file = Join-Path $PSScriptRoot "variables.yml"
+  $runner_project = Join-Path $PSScriptRoot "..\${runner_name}\${runner_name}.csproj"
+  $app_project = Join-Path $PSScriptRoot "..\${app_name}\${app_name}.csproj"
 
   # Install dotnet
   $dotnet_version = Get-YamlValue $config_file 'dotnet-version'
@@ -66,6 +68,10 @@ function Invoke-WindowsUITests {
     }
   }
 
+  # Ensure predictable artifacts output layout for setting env:UITEST_APP_PATH later
+  $build_params_artifacts = @('-p:ArtifactsPivots=TestBuild', '-p:UseArtifactsOutput=true')
+  $build_parameters += $build_params_artifacts
+
   # Build app and runner projects
   & $dotnet_exe build $app_project @build_parameters
   if ($LASTEXITCODE -ne 0) {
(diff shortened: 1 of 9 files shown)


src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,311 @@
...
+      if (![string]::IsNullOrWhiteSpace($trx_filename)) {
+        $test_run_params += @('--report-trx-filename', $trx_filename)
+      }
+
+      $env:TKUITEST_APP = Join-Path $PSScriptRoot "..\artifacts\bin\${app_name}\TestBuild\${app_name}.exe"
+      & $dotnet_exe test --project $runner_project @build_parameters @test_run_params
+    }
+    finally {
+      # Kill appium and return to original location
+      Stop-Process -InputObject $appium_server_process
```
mstefarov: Two suggestions:
- Guard against the process never having started: if `Start-Process` above failed, `$appium_server_process` is empty and this line throws. Wrap it: `if ($appium_server_process) { ... }`
- This stops the Appium (node) process, but Appium also launches WinAppDriver.exe and the test app as separate processes.  Powershell's `Stop-Process` will not kill those.  To kill the whole process tree you can do `taskkill /T /F /PID $($appium_server_process.Id)` (`/T` also stops the child processes).

src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,311 @@
...
+        exit 1
+      }
+
+      $deadline = (Get-Date).AddSeconds(60)
+      while ((Get-Date) -lt $deadline) {
+          try {
+              $r = Invoke-RestMethod -Uri 'http://127.0.0.1:4723/status' -TimeoutSec 2
+              if ($r.value.ready) { break }
+          } catch { Start-Sleep -Milliseconds 500 }
+      }
```
mstefarov, line 143: I think we forgot to check whether whether Appium is actually ready, or if we just ran out of time.

```suggestion
      $ready = $false
      while ((Get-Date) -lt $deadline) {
          try {
              $r = Invoke-RestMethod -Uri 'http://127.0.0.1:4723/status' -TimeoutSec 2
              if ($r.value.ready) { $ready = $true; break }
          } catch { Start-Sleep -Milliseconds 500 }
      }
      if (-not $ready) { Write-Error 'Appium did not become ready within 60s.'; exit 1 }
```

src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,311 @@
...
+  # Install koffi
+  & $npm_exe install koffi --prefix $node_workspace
+  if ($LASTEXITCODE -ne 0) {
+    Write-Error "Error installing koffi node module. Koffi is necessary to avoid reliance on node-ffi, which in turn requires a Visual Studio install."
+    exit $LASTEXITCODE
+  }
+
+  # Extract and configure WindowsAppDriver for appium
+  $wad_installer = Join-Path $workspace 'WinAppDriver.msi'
+  & curl.exe -sSL https://github.com/microsoft/WinAppDriver/releases/download/v1.2.1/WindowsApplicationDriver_1.2.1.msi -o $wad_installer
```
mstefarov: By default curl can exit with success (0) even when the download fails (e.g. github down again). By default it just writes the error page into the file. So the `$LASTEXITCODE` check below passes, and the problem only shows up later as a "corrupt file" error. Add [--fail](https://curl.se/docs/manpage.html#--fail) so curl reports it immediately. Same applies to the other curl downloads in this file (the Node.js zip and dotnet-install.ps1).

```suggestion
  & curl.exe -fsSL https://github.com/microsoft/WinAppDriver/releases/download/v1.2.1/WindowsApplicationDriver_1.2.1.msi -o $wad_installer
```

src/Tests/UITests/cibuild/utils.psm1
```diff
@@ -0,0 +1,311 @@
...
+  $runner_project = Join-Path $PSScriptRoot "..\${runner_name}\${runner_name}.csproj"
+  $app_project = Join-Path $PSScriptRoot "..\${app_name}\${app_name}.csproj"
+
+  # Install dotnet
+  $dotnet_version = Get-YamlValue $config_file 'dotnet-version'
+  $dotnet_exe = Install-Dotnet $workspace $dotnet_version $env:DOTNET_CACHE_FOLDER
+
+  # Install maui workload if required
+  if ($install_maui) {
+    & $dotnet_exe workload install maui-windows
```
mstefarov, line 43: Add a `$LASTEXITCODE` check here too, in case workload install fails.
```suggestion
    & $dotnet_exe workload install maui-windows
    if ($LASTEXITCODE -ne 0) { Write-Error 'maui workload install failed.'; exit $LASTEXITCODE }
```

imalcolm1:
@dotMorten @SamhithaBurugupalli cc: @mstefarov
The current intention is for these scripts to eventually ported to the C# script framework introduced in the [Android PR](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/759). After discussion with Matvei, we've decided that from this point only major problems with this PR will need addressing since these scripts will eventually be rewritten to C# anyway

mstefarov (approved):

dotMorten (approved):

PR #765, opened 2026-06-03
UITests: iOS CI Tests

imalcolm1:
Scripts for running CI tests for maui android.

## Notes for reviewers
**Branches off Android PR**
The Android PR is the base for this one: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/759. I will likely rebase this branch as I receive feedback for the other PR, but the contents that is different here should not change drastically.

**Real Devices Only**
Currently this is only set up to support real devices. Running on a simulator is likely not to work.

**Changes to test runners/apps**
Somewhat significant changes to how the MauiiOS test setups run. The automatic builds were causing problems so those are toggleable now. The MauiiOS test setup also allows wildcard settings from the shell environment now.

## Testing Locally
To test locally run the following commands in zsh. The NUGET_REPO directory should contain the nuget package from a daily build or a release, otherwise leave it unset.
```bash
export WORKSPACE=$(realpath ~/Developer)/uitest-workspace
mkdir $WORKSPACE
cd $WORKSPACE
git clone https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit.git -b ian/uitests-cibuild-ios
export RELEASE_VERSION=300.1.0
export NUGET_REPO=${WORKSPACE}/nuget-source # Nuget files must be placed here

# Target device UDID (required)
export TKUITEST_DEVICE='your device udid'

# Parameters for automatic signing
export TKUITEST_PARAM_xcodeOrgId='Our apple org id'
export TKUITEST_PARAM_xcodeSigningId='Apple Development'
export TKUITEST_PARAM_updatedWDABundleId='com.esri.wda'

# Alternate parameters for manual signing used by the actual jenkins job
export BUILD_WDA_MANUAL=true
export DEVELOPMENT_TEAM='our apple developer org id'
export PROVISIONING_PROFILE_SPECIFIER='our team developer profile name'

# Once the nuget repo has been copied, run this in the same terminal
${WORKSPACE}/arcgis-maps-sdk-dotnet-toolkit/src/Tests/UITests/cibuild/cibuild.sh MauiiOS
```


src/Tests/UITests/Toolkit.UITests.Shared/AppiumSetup.cs
```diff
@@ -121,14 +121,12 @@ private static MacDriver MakeMacDriver(string bundleId, string port = "4723", in
...
-            ?? throw new InvalidOperationException("Could not determine test assembly directory.");
-
-        var pathFile = Path.Combine(testAssemblyDir, "AppBuildInfo.txt");
-        if (!File.Exists(pathFile))
+        var pathFile = GetBuildFile();
+        if (pathFile == null)
         {
             throw new FileNotFoundException(
-                $"Missing '{pathFile}'. Ensure the 'BuildTestApp' MSBuild target ran before tests.");
+                $"Missing '{pathFile}'. Ensure the 'BuildTestApp' MSBuild target ran before tests."
```
mstefarov, line 128: This will always print `Missing ''` (blank name), won't it? The exception is only reachable when it's null.

I would keep path-creation logic in `GetBuildFile` (it helps to avoid duplication), but move `File.Exists` checks to the point of use.

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -213,6 +214,73 @@ public CommonDependencies(string dotnetExe)
...
+        if (buildManual != null && buildManual == "true")
+        {
+            Console.WriteLine("\nStarting manual WDA clone and build...");
+
+            var deviceUdid = Environment.GetEnvironmentVariable("TKUITEST_DEVICE");
+            var developmentTeam = Environment.GetEnvironmentVariable("DEVELOPMENT_TEAM");
+            var provisioningProfile = Environment.GetEnvironmentVariable("PROVISIONING_PROFILE_SPECIFIER");
+            if (deviceUdid == null || developmentTeam == null || provisioningProfile == null)
+            {
+                throw new ArgumentNullException("To run the MauiiOS cibuild with a manual WDA build, please set the TKUITEST_DEVICE, DEVELOPMENT_TEAM, and PROVISIONING_PROFILE_SPECIFIER environment variables.");
```
mstefarov: Nit: `ArgumentNullException(string)` constructor actually treats the string as a parameter name. So your error would look like

> Value cannot be null. (Parameter 'To run the MauiiOS cibuild...

Stick to `ArgumentException(string)` or `InvalidOperationException(string)` instead.

src/Tests/UITests/Toolkit.UITests.Shared/AppiumSetup.cs
imalcolm1: Ah good point!

mstefarov (approved):

SamhithaBurugupalli (approved):

PR #768, opened 2026-06-05
UIToolkit: Use floating version for MagickNet

imalcolm1:
MagickNet seems to constantly get new security updates which can lead to sometimes hundreds of warnings when all the test runners are built together. Setting it to a floating version off the current major version should make this less of a problem.

### DIFF
--- a/src/Tests/UITests/Directory.Build.props
+++ b/src/Tests/UITests/Directory.Build.props
@@ -46,6 +46,6 @@
   <PropertyGroup>
     <TestRunnerNetVersion>net10.0</TestRunnerNetVersion>
     <AppiumWebDriverVersion>8.2.0</AppiumWebDriverVersion>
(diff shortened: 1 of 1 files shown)


mstefarov (approved):
Thanks for doing this! I was about to propose the same change; got tired of looking at nuget vulnerability warnings :)

dotMorten (changes requested):
I don’t think running with a floating version is a good idea. Tests could suddenly be failing and we wouldn’t know why just because they shipped a new version.

imalcolm1:
> I don’t think running with a floating version is a good idea. Tests could suddenly be failing and we wouldn’t know why just because they shipped a new version.

@dotMorten Is the preferred fix to suppress the warnings or just manually update the repo whenever a new warning appears?

imalcolm1:
Closing in favor of #769

PR #771, opened 2026-06-09
Default search view UI tests 

SamhithaBurugupalli:
- Adds UI tests for default Search view control
- Adds/ updates automation properties
- Adds/ updates appium helpers

### DIFF
--- a/src/Tests/UITests/Toolkit.UITests.Shared/Appium/AppiumTestBase.Utils.cs
+++ b/src/Tests/UITests/Toolkit.UITests.Shared/Appium/AppiumTestBase.Utils.cs
@@ -83,8 +83,6 @@ public abstract partial class AppiumTestBase
     {
 #if ANDROID_TEST
         var action = () => Driver.FindElement(MobileBy.AndroidUIAutomator($"new UiSelector().description(\"{name}\")"));
-#elif WINDOWS_TEST
-        var action = () => Driver.FindElement(MobileBy.Name(name));
 #elif MAC_TEST || IOS_TEST
         var action = () => Driver.FindElement(MobileBy.XPath($"//*[@label=\"{name}\" or @name=\"{name}\"]"));
 #else
--- a/src/Tests/UITests/Toolkit.UITests.Shared/Appium/AppiumTestBase.Utils.cs
+++ b/src/Tests/UITests/Toolkit.UITests.Shared/Appium/AppiumTestBase.Utils.cs
@@ -82,7 +82,11 @@ public abstract partial class AppiumTestBase
     protected AppiumElement FindElementByName(string name, TimeSpan? timeout = null)
     {
 #if ANDROID_TEST
-        var action = () => Driver.FindElement(MobileBy.AndroidUIAutomator($"new UiSelector().text(\"{name}\")"));
+        var action = () => Driver.FindElement(MobileBy.AndroidUIAutomator($"new UiSelector().description(\"{name}\")"));
+#elif WINDOWS_TEST
+        var action = () => Driver.FindElement(MobileBy.Name(name));
+#elif MAC_TEST || IOS_TEST
+        var action = () => Driver.FindElement(MobileBy.XPath($"//*[@label=\"{name}\" or @name=\"{name}\"]"));
 #else
         var action = () => Driver.FindElement(MobileBy.Name(name));
 #endif
@@ -97,7 +101,65 @@ public abstract partial class AppiumTestBase
         }
     }
 
-    protected string GetElementText(AppiumElement element, TimeSpan? timeout = null)
+    protected AppiumElement FindElementByText(string text, TimeSpan? timeout = null)
+    {
+#if ANDROID_TEST
+        var action = () => Driver.FindElement(MobileBy.AndroidUIAutomator($"new UiSelector().text(\"{text}\")"));
+#elif MAC_TEST || IOS_TEST
+        var action = () => Driver.FindElement(MobileBy.XPath($"//*[@label=\"{text}\" or @value=\"{text}\" or @name=\"{text}\"]"));
+#else
+        var action = () => Driver.FindElement(MobileBy.Name(text));
+#endif
+        try
+        {
+            return OptionalWaitCall(action, timeout);
+        }
+        catch (Exception)
+        {
+            TestContext.WriteLine($"No elements found with text \"{text}\". See exception for details.");
+            throw;
+        }
+    }
+
+    protected bool ElementExistsByName(string name, TimeSpan? timeout = null)
+    {
+        try
+        {
+            FindElementByName(name, timeout);
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+    protected bool ElementExistsById(string id, TimeSpan? timeout = null)
+    {
+        try
+        {
+            FindElement(id, timeout);
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+
+    protected bool ElementExistsByText(string text, TimeSpan? timeout = null)
+    {
+        try
+        {
+            FindElementByText(text, timeout);
+            return true;
+        }
+        catch
+        {
+            return false;
+        }
+    }
+
+    protected string GetLabelText(AppiumElement element, TimeSpan? timeout = null)
     {
 #if WINDOWS_TEST
         var action = () => element.GetAttribute("Name");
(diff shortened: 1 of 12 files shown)


src/Tests/UITests/Toolkit.UITests.TestPages.Shared/SearchView/SearchViewControlMap.xaml.cs
```diff
@@ -0,0 +1,65 @@
...
+            return;
+        }
+
+        var center = new MapPoint(longitude, latitude, SpatialReferences.Wgs84);
+        MyMapView.SetViewpoint(new Viewpoint(center, scale));
+
+        UpdateCoordinateAndScaleTextBoxes();
+    }
+
+    private void UpdateCoordinateAndScaleTextBoxes()
```
mstefarov: The viewpoint has a race condition and the tests barely need it. The ViewpointChanged handler rewrites the same text boxes that the test types into. If the map takes a few seconds to load, it overwrites the typed values and the test navigates to the wrong place. I reproduced this locally with just a 1.5 second load delay.

But, rather than patching that, you can simplify the setup. The tests only ever use 4 fixed viewpoints, never read lon/lat back, and read scale exactly once (in SearchViewControl_SearchSuggestions). So you can trim the UI a lot:

- Replace the Scale/Log/Lat boxes and Update button with one button per preset viewpoint, that calls SetViewpoint with hardcoded values in code-behind. One click instead of three typed fields = less Apium actions and less to go wrong.
- Add a read-only scale label, to use it for the zoom assertion in SearchSuggestions.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewControlMapTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        var selectedResult = FindElementByName("Colorado Springs Airport, Colorado Springs, Colorado", TimeSpan.FromSeconds(5));
+        selectedResult.Click();
+
+        // Verify the selected search result displays the expected callout content.
+        Assert.IsTrue(ElementExistsByText("Colorado Springs Airport", TimeSpan.FromSeconds(5)), "Callout title Colorado Springs Airport is not visible");
+        Assert.IsTrue(ElementExistsByText("Colorado Springs, Colorado", TimeSpan.FromSeconds(5)), "Callout Description is not visible");
+
+        // After selecting a result, the results list should be hidden and repeat search should not be shown yet.
+        Assert.IsFalse(ElementExistsById("SearchResultsList"));
+        Assert.IsFalse(ElementExistsByName("Repeat Search Here"), "Expected 'Repeat Search Here' button to be hidden initially");
```
mstefarov: Would this button benefit from having an automation ID too? You've done a nice job of unifying IDs for other elements (QueryEntry, SearchButton, ClearSearchButton, SearchResultsList) but this one gets looked up by the label text 3 times instead.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewControlMapTests.cs
```diff
@@ -0,0 +1,162 @@
...
+#if MAC_TEST
+        Assert.AreEqual("Find a place or address", GetEntryText(FindElement("QueryEntry", TimeSpan.FromSeconds(5))), "Expected the search input placeholder to be visible.");
+#else
+        Assert.IsTrue(ElementExistsByText("Find a place or address", TimeSpan.FromSeconds(5)), "Expected the search input placeholder to be visible.");
+#endif
+        Assert.IsTrue(ElementExistsByName("Search"), "Expected the Search button to be visible.");
+        Assert.IsFalse(ElementExistsByName("Clear Search"), "Expected the Clear Search button to be hidden before search text is entered.");
+        Assert.IsFalse(ElementExistsById("SearchResultsList"), "Expected search results to be hidden before a search is performed.");
+        Assert.IsFalse(ElementExistsById("SearchSuggestionsList"), "Expected search suggestions to be hidden before search text is entered.");
+        Assert.IsFalse(ElementExistsById("SearchSourcesList"), "Expected search sources to be hidden initially.");
```
mstefarov, line 26: It's called `SearchSourcesList` on MAUI, but on WPF/WinUI the same element has a different ID (`SourceList`) so this assertion will never catch a problem on those platforms.

Since you are already normalizing automation IDs across platforms, this one might be worth fixing too.

src/Tests/UITests/Toolkit.UITests.WinUI.App/Toolkit.UITests.WinUI.App.csproj
```diff
@@ -17,6 +17,7 @@
   </PropertyGroup>
   <ItemGroup>
     <None Remove="TestPages\ScaleLines.xaml" />
+	<None Remove="TestPages\SearchViewControlMap.xaml" />
   </ItemGroup>
```
imalcolm1, line 20: This ItemGroup is actually vestigial and was replaced by the wildcard group below. I'll probably put up a PR to remove it and possibly do some other tweaks, but for now this line can be undone I think.

src/Tests/UITests/Toolkit.UITests.Shared/Appium/AppiumTestBase.Utils.cs
```diff
@@ -82,7 +82,11 @@ protected AppiumElement FindElement(AppiumElement parent, string id, TimeSpan? t
     protected AppiumElement FindElementByName(string name, TimeSpan? timeout = null)
     {
 #if ANDROID_TEST
-        var action = () => Driver.FindElement(MobileBy.AndroidUIAutomator($"new UiSelector().text(\"{name}\")"));
+        var action = () => Driver.FindElement(MobileBy.AndroidUIAutomator($"new UiSelector().description(\"{name}\")"));
+#elif WINDOWS_TEST
+        var action = () => Driver.FindElement(MobileBy.Name(name));
```
imalcolm1: This is already covered by the `#else` condition, so we might as well remove it
```suggestion
```

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewControlMapTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        // After selecting a result, the results list should be hidden and repeat search should not be shown yet.
+        Assert.IsFalse(ElementExistsById("SearchResultsList"));
+        Assert.IsFalse(ElementExistsByName("Repeat Search Here"), "Expected 'Repeat Search Here' button to be hidden initially");
+
+        // Move the map to a new location so the previous search can be repeated in the new visible extent.
+        UpdateViewpoint(107962, -117.593468, 34.069134);
+
+        // Wait for the map viewpoint update to complete and for the repeat search button to become available.
+        await Task.Delay(1000);
+        Assert.IsTrue(ElementExistsByName("Repeat Search Here", TimeSpan.FromSeconds(5)), "Expected 'Repeat Search Here' button to be visible after moving the map");
```
imalcolm1: This await may not be necessary given there's already a timeout on the ElementExistsByName() call

imalcolm1 (approved):
Looks good!

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewControlMapTests.cs
imalcolm1: Also the clear button maybe?

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewControlMapTests.cs
```diff
@@ -0,0 +1,146 @@
...
+        Assert.AreEqual("Find a place or address", GetEntryText(FindElement("QueryEntry", TimeSpan.FromSeconds(5))), "Expected the search input placeholder to be visible.");
+#else
+        Assert.IsTrue(ElementExistsByText("Find a place or address", TimeSpan.FromSeconds(5)), "Expected the search input placeholder to be visible.");
+#endif
+        Assert.IsTrue(ElementExistsById("SearchButton"), "Expected the Search button to be visible.");
+        Assert.IsFalse(ElementExistsById("ClearSearchButton"), "Expected the Clear Search button to be hidden before search text is entered.");
+        Assert.IsFalse(ElementExistsById("SearchResultsList"), "Expected search results to be hidden before a search is performed.");
+        Assert.IsFalse(ElementExistsById("SearchSuggestionsList"), "Expected search suggestions to be hidden before search text is entered.");
+        Assert.IsFalse(ElementExistsById("SearchSourcesList"), "Expected search sources to be hidden initially.");
+        Assert.IsFalse(ElementExistsById("SelectSearchSourceButton"), "Expected the source selector to be hidden initially.");
```
mstefarov, line 27: The button is actually named "SourceSelectToggle" on MAUI and "SelectSearchSourceToggle" on WPF/WinUI.

mstefarov (approved):
One last assertion to fix up, otherwise this is good to go 🚀

PR #772, opened 2026-06-10
UITests: Add api key configuration

imalcolm1:
This PR exposes a settable api key property in the test Directory.Build.props file by way of a small source code generator. It touches a lot of files but the diffs are small.


src/Tests/UITests/Toolkit.UITests.Generators/ApiKeyGenerator.cs
```diff
@@ -0,0 +1,37 @@
...
+            if (string.IsNullOrWhiteSpace(key)) {
+                var descriptor = new DiagnosticDescriptor(
+                    id: "TKUITEST001",
+                    title: "Missing Toolkit.UITests api key",
+                    messageFormat: "The Toolkit.UITests are missing an API key, some tests will fail.",
+                    category: "Toolkit.UITests",
+                    defaultSeverity: DiagnosticSeverity.Warning,
+                    isEnabledByDefault: false);
+                productionContext.ReportDiagnostic(Diagnostic.Create(descriptor, null, new string[0]));
+            }
```
imalcolm1, line 24: This is admittedly a pretty useless error since it's disabled by default. Enabling it would mean subjecting all users to 3 warnings whenever they open the project without setting the api key property, but it seems like good information to have somewhere, I'm just not sure how to do it in a minimally disruptive way

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -44,11 +44,12 @@ static async Task<int> Main(string[] args)
         var workspace = Environment.GetEnvironmentVariable("WORKSPACE");
         var dotnetExe = Environment.GetEnvironmentVariable("DOTNET_PATH");
         var toolkitSrc = Environment.GetEnvironmentVariable("TOOLKIT_SRC");
-        if (string.IsNullOrWhiteSpace(workspace) || string.IsNullOrEmpty(dotnetExe) || string.IsNullOrEmpty(toolkitSrc)) {
-            throw new ArgumentException("Environment variables WORKSPACE, DOTNET_PATH, and TOOLKIT_SRC must all be set.");
+        var apiKey = Environment.GetEnvironmentVariable("ARCGIS_API_KEY");
+        if (string.IsNullOrWhiteSpace(workspace) || string.IsNullOrEmpty(dotnetExe) || string.IsNullOrEmpty(toolkitSrc) || string.IsNullOrEmpty(apiKey)) {
+            throw new ArgumentException("Environment variables WORKSPACE, DOTNET_PATH, ARCGIS_API_KEY, and TOOLKIT_SRC must all be set.");
```
mandel-capuchin: I know it sounds verbose (and maybe a little pedantic) but I think we want to return exactly which one was not set so that it is easier to check, what about using a look, something like:

```csharp
foreach(var (key, val) in new Dictionary<string, string> {"WORKSPACE", worjsapce}){
    if (val is null)
        // throw here
}
```
Of create a string builder with the final exception message that contains which env vars are missing

src/Tests/UITests/cibuild/cibuild.cs
```diff
@@ -44,11 +44,12 @@ static async Task<int> Main(string[] args)
...
         }
         if (!Path.Exists(workspace) || !Path.Exists(dotnetExe) || !Path.Exists(toolkitSrc)) {
-            throw new ArgumentException("Workspace and dotnet directory must be existing paths.");
+            throw new ArgumentException("Workspace, dotnet executable, and toolkit source directory must be existing paths.");
```
mandel-capuchin: Same, this will be read from CI probably so we want to be super explicit with what is missing

src/Tests/UITests/Toolkit.UITests.Generators/ApiKeyGenerator.cs
```diff
@@ -0,0 +1,37 @@
...
+/// This generator exposes the TestAppApiKey property for use in the UITest test apps.
+/// </summary>
+[Generator(LanguageNames.CSharp)]
+public class ApiKeyGenerator : IIncrementalGenerator
+{
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        // Get API key property, see https://github.com/dotnet/roslyn/blob/eb789e2741f6f22d9e283e2049dc1378871323e0/docs/features/incremental-generators.cookbook.md#consume-msbuild-properties-and-metadata
+        IncrementalValueProvider<string?> keyProvider = context.AnalyzerConfigOptionsProvider.Select((provider, ct) =>
+            provider.GlobalOptions.TryGetValue("build_property.TestAppApiKey", out var key) ? key : null);
```
mandel-capuchin: Hmm what happens if key is null? Looks to me that we can always return the value in key, since if it is false, is null nevertheless

src/Tests/UITests/Toolkit.UITests.Generators/ApiKeyGenerator.cs
```diff
@@ -0,0 +1,37 @@
...
+        context.RegisterSourceOutput(keyProvider, (productionContext, key) =>
+        {
+            if (string.IsNullOrWhiteSpace(key)) {
+                var descriptor = new DiagnosticDescriptor(
+                    id: "TKUITEST001",
+                    title: "Missing Toolkit.UITests api key",
+                    messageFormat: "The Toolkit.UITests are missing an API key, some tests will fail.",
+                    category: "Toolkit.UITests",
+                    defaultSeverity: DiagnosticSeverity.Warning,
+                    isEnabledByDefault: false);
```
mandel-capuchin: You want you descriptors to be static and use them later in the calls, create a new descriptor for every error is too much work

src/Tests/UITests/Toolkit.UITests.Wpf.App/App.xaml.cs
```diff
@@ -10,6 +11,9 @@ public partial class App : Application
     public App()
     {
         ControlPatcher.ApplyPatches();
-        Esri.ArcGISRuntime.ArcGISRuntimeEnvironment.Initialize();
+        ArcGISRuntimeEnvironment.Initialize((config) =>
+        {
+            config.UseApiKey(ApiKeyProvider.Key);
+        });
```
mandel-capuchin: we can make the lamda static, right?

mandel-capuchin (changes requested):

src/Tests/UITests/cibuild/cibuild.cs
imalcolm1: Makes sense, I've split this process into a couple new methods

SamhithaBurugupalli (approved):

mandel-capuchin (approved):

PR #774, opened 2026-06-16
FeatureFormView: Add ability to remove associations from Utility Network

dotMorten:
Also adds a "more information" details page about the association according to figma design.


src/Toolkit/Toolkit.WinUI/UI/Controls/FeatureForm/FeatureForm.Theme.xaml
```diff
@@ -466,11 +467,211 @@
...
+                                Padding="0">
+                            <FontIcon Glyph="&#xE712;" FontSize="18" />
+                        </Button>
                     </Grid>
                 </ControlTemplate>
             </Setter.Value>
         </Setter>
     </Style>
 
+    <Style TargetType="primitives:UtilityAssociationResultDetailsView" >
```
dotMorten, line 487: This and above is the main change. The rest is just moved over from the .WinUI file

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationResultDetailsView.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            if (GetTemplateChild("FromElementText") is TextBlock fromElementText)
+            {
+                fromElementText.Text = ffv?.CurrentFeatureForm?.Title;
+            }
+
+            if (GetTemplateChild("ToElementText") is TextBlock toElementText)
+            {
+                toElementText.Text = AssociationResult?.AssociatedFeature is null ? "" : new FeatureForm(AssociationResult?.AssociatedFeature!)?.Title;
+            }
```
mstefarov, line 182: This codes assumes that FromElement is the current feature, but it can actually be _either_ the FromElement or the ToElement. For example in a Containment association we could be looking at the container (FromElement) or the content (ToElement).

Looks like `UtilityAssociationResultView.GetAssociationProperty` already handles this correctly by matching AssociatedFeature's `GlobalId` against both FromElement and ToElement rather than assuming a direction. The details view probably needs the same logic.

src/Toolkit/Toolkit.WPF/UI/Controls/FeatureForm/FeatureFormView.Theme.xaml
```diff
@@ -668,17 +728,27 @@
                             <DataTemplate>
                                 <StackPanel VerticalAlignment="Center">
                                     <TextBlock Text="{Binding Filter.Title, Mode=OneWay}" Style="{StaticResource FeatureFormViewHeaderStyle}" Visibility="{Binding Filter.Title, Converter={StaticResource FeatureFormViewVisibilityConverter}}" />
-                                    <TextBlock Text="{Binding Filter.Description, Mode=OneWay}" Style="{StaticResource FeatureFormViewTitleStyle}" Visibility="{Binding Filter.Description, Converter={StaticResource FeatureFormViewVisibilityConverter}}" />
+                                    <TextBlock Text="{Binding AdditionalContent.Title, Mode=OneWay, RelativeSource={RelativeSource AncestorType=primitives:NavigationSubView}}" Style="{StaticResource FeatureFormViewTitleStyle}"
+                                               Visibility="{Binding AdditionalContent.Title, Mode=OneWay, RelativeSource={RelativeSource AncestorType=primitives:NavigationSubView}, Converter={StaticResource FeatureFormViewVisibilityConverter}}" />
```
mstefarov, line 742: The navigation breadcrumb subtitles look good on WPF, but missing from MAUI.  I don't see `additionalContent` anywhere in MAUI's filter result and group result views. Do they need template updates too?

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
```diff
@@ -237,6 +237,134 @@
...
+    <comment>Text Label accepting utility association deletion.</comment>
+  </data>
+  <data name="FeatureFormDeleteAssociationConfirmationCancel" xml:space="preserve">
+    <value>Cancel</value>
+    <comment>Text Label cancelling utility association deletion.</comment>
+  </data>
+  <data name="FeatureFormUtilityAssociationsAddAssociation" xml:space="preserve">
+    <value>Add Association</value>
+    <comment>Button text for starting or completing the add association workflow in FeatureFormView.</comment>
+  </data>
```
mstefarov, line 279: I see a whole lot of unused resource strings related to adding associations. I'm guessing they are for the next PR?

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationResultDetailsView.cs
```diff
@@ -0,0 +1,201 @@
...
+                toTerminalText.Text = AssociationResult?.Association?.ToElement?.Terminal?.Name;
+#if MAUI
+                toTerminalText.IsVisible = !string.IsNullOrEmpty(toTerminalText.Text);
+#else
+                toTerminalText.Visibility = string.IsNullOrEmpty(toTerminalText.Text) ? Visibility.Collapsed : Visibility.Visible;
+#endif
+            }
+        }
+
+        private async System.Threading.Tasks.Task<bool> ConfirmDeleteAssociationAsync()
```
mstefarov: Is this dead code, or was it supposed to be wired up? Currently nothing calls this.

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
dotMorten: yup for the add work. I figured I'd keep them rather than throw them away

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationResultDetailsView.cs
dotMorten: Found out we have to show it here as-is

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
```diff
@@ -237,6 +237,138 @@
...
+    <comment>Title for the utility association settings/details in FeatureFormView.</comment>
+  </data>
+  <data name="FeatureFormUtilityAssociationsFromElement" xml:space="preserve">
+    <value>From Element</value>
+    <comment>Text Label for the utility association from connection.</comment>
+  </data>
+  <data name="FeatureFormUtilityAssociationsToElement" xml:space="preserve">
+    <value>To Element</value>
+    <comment>Text Label for the utility association to connection.</comment>
+  </data>
```
mstefarov: There are two duplicate definitions for each of these keys in Resources.resx

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationResultDetailsView.cs
```diff
@@ -0,0 +1,228 @@
...
+            string? toTerminal = null;
+
+            if (AssociationResult is not null)
+            {
+                // If association Type is (Connectivity or JunctionEdgeObjectConnectivityMidspan or JunctionEdgeObjectConnectivityFromSide or JunctionEdgeObjectConnectivityToSide) and the terminal on the UN Element(feature) is not null, this terminal will also be displayed
+                if (AssociationResult.Association.AssociationType == UtilityAssociationType.Connectivity ||
+                   AssociationResult.Association.AssociationType == UtilityAssociationType.JunctionEdgeObjectConnectivityMidspan ||
+                   AssociationResult.Association.AssociationType == UtilityAssociationType.JunctionEdgeObjectConnectivityFromSide ||
+                   AssociationResult.Association.AssociationType == UtilityAssociationType.JunctionEdgeObjectConnectivityToSide &&
+                   AssociationResult.AssociatedFeature is not null)
```
mstefarov: Is the precedence correct here? The AssociatedFeature would only be checked for JunctionEdgeObjectConnectivityToSide, as written

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationResultDetailsView.cs
```diff
@@ -0,0 +1,228 @@
...
+        public override void OnApplyTemplate()
+#endif
+        {
+            base.OnApplyTemplate();
+            if (GetTemplateChild("RemoveAssociationButton") is Button button)
+            {
+#if MAUI
+                button.Clicked += (s, e) => RemoveAssociation();
+#else
+                button.Click += (s, e) => RemoveAssociation();
```
mstefarov, line 69: Do these click handlers ever need to be unsubscribed?  The ResultView keeps track and unsubscribes its  _detailsButton.

dotMorten: It's honestly not super important - it would ever only be an issue if a new template keeps getting applied after rendering (really doesn't happen). Not unsubscribing has the benefit of not having to keep the reference to the button as well.

src/Toolkit/Toolkit/UI/Controls/FeatureForm/UtilityAssociationResultDetailsView.cs
dotMorten: That null check isn't needed any longer. Removing

src/Toolkit/Toolkit/LocalizedStrings/Resources.resx
dotMorten: Good eye!

PR #776, opened 2026-06-17
catalyst: Stabilize compass UI tests

mandel-capuchin:
Wait for Compass_Rotates to observe the expected orientation instead of sampling the first screenshot that contains red pixels. This gives Catalyst and mac2 time to reflect visual rotation changes before assertions run.

Normalize the standalone heading bound to the test compass so equivalent angles avoid the raw 270 -> -270 visual rotation path that has been unreliable on Catalyst. The map still receives the original rotation value.

Fix the Compass_AutoHides timeout assertion and wait for the reset test's map rotation label through the platform-specific text accessor.


imalcolm1 (approved):
Thanks for the fix!

PR #777, opened 2026-06-17
UITest maintenance

imalcolm1:
A couple quick fixes/improvements:
1. Remove redundant build item in winui test runner (the xaml file it included is already covered by a wildcard item)
2. Add option to pass mstest filter argument to cibuild script for mac/ios/android (not windows for now since the same functionality should be easy to accomplish with the test explorer)
3. No longer throw exception when tests fail since it was showing up in the test results as an error when they did
4. Fixes visual studio not finding the code generator when doing the auto build for WPF/MauiWinUI
5. Address a nullability warning in the iOS runner


imalcolm1:
Just added another quick fix for the automatic app builds when running tests for Wpf and MauiWinUI from visual studio (VS didn't recognize the dependency on the generator unless it is explicitly added to the sln)

SamhithaBurugupalli (approved):

mandel-capuchin (approved):

PR #778, opened 2026-06-17
UITests ScaleLine updates

imalcolm1:
This PR does has two parts:
1. Adds a timeout for finding the scale line elements which should help with the occasional flakiness in the ci tests
2. Uses buttons instead of entries for updating the viewpoint since there are only 3 viewpoints and it is quicker for appium to find and press buttons than multiple entries

### DIFF
--- a/src/Tests/UITests/Toolkit.UITests.Shared/Tests/ScaleLine/ScaleLineTests.cs
+++ b/src/Tests/UITests/Toolkit.UITests.Shared/Tests/ScaleLine/ScaleLineTests.cs
@@ -118,9 +118,9 @@ public class ScaleLineTests : AppiumTestBase
     private AppiumElement GetScaleElement(ScaleLineType type)
     {
         if (type == ScaleLineType.Advanced)
-            return FindElement("AdvancedScaleLine");
+            return FindElement("AdvancedScaleLine", TimeSpan.FromSeconds(5));
         else if (type == ScaleLineType.Simple)
-            return FindElement("SimpleScaleLine");
+            return FindElement("SimpleScaleLine", TimeSpan.FromSeconds(5));
         else
             throw new ArgumentOutOfRangeException(nameof(type), type, "Invalid scale line type.");
     }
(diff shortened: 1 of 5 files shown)


SamhithaBurugupalli (approved):

src/Tests/UITests/Toolkit.UITests.Shared/Tests/ScaleLine/ScaleLineTests.cs
```diff
@@ -118,23 +118,32 @@ private ScaleLineInfo GetScaleLineInfo(ScaleLineType type)
     private AppiumElement GetScaleElement(ScaleLineType type)
     {
         if (type == ScaleLineType.Advanced)
-            return FindElement("AdvancedScaleLine");
+            return FindElement("AdvancedScaleLine", TimeSpan.FromSeconds(5));
         else if (type == ScaleLineType.Simple)
-            return FindElement("SimpleScaleLine");
+            return FindElement("SimpleScaleLine", TimeSpan.FromSeconds(5));
```
mandel-capuchin: What about moving that TimeSpan as a private variable so that you can re-use it. That way if we need to bump it, we just do  it in a single location

src/Tests/UITests/Toolkit.UITests.Shared/Tests/ScaleLine/ScaleLineTests.cs
```diff
@@ -118,23 +118,32 @@ private ScaleLineInfo GetScaleLineInfo(ScaleLineType type)
...
+    private void SelectViewpoint(ViewpointPreset preset)
     {
-        var scaleInputElement = FindElement("ScaleTextBox");
-        SubmitText(scaleInputElement, scale.ToString());
+        var buttonAutomationId = preset switch
+        {
+            ViewpointPreset.WorldEquator => "WorldEquatorViewpoint",
+            ViewpointPreset.DetailEquator => "EquatorDetailViewpoint",
+            ViewpointPreset.DetailHighLatitude => "HighLatitudeDetailViewpoint",
+            _ => throw new ArgumentOutOfRangeException(nameof(preset), preset, "Invalid viewpoint preset.")
```
mandel-capuchin: you are capturing preset, is that intended? Will that be a memory issue?

mandel-capuchin (approved):
some small comments that will not stop the PR from landing

mstefarov (approved):
Looks great 👍

PR #779, opened 2026-06-18
UI testing: Automated tests for Search view customization workflow

SamhithaBurugupalli:
- Adds tests for searchview with multiple geocoders

- Adds tests to check bindings (Geoview, EnableIndividualResultDisplay, EnableRepeatSearchHereButton, EnableResultListView)

- Adds tests for text customizations (DefaultPlaceholder, NoResultMessage, RepeatSearchButtonText, SearchTooltipText, ClearSearchTooltipText, AllSourceSelectText)

- Adds tests for Selection and deselection events of search results

- Fixes a null reference exception in Maui

- Adds/ updates automation properties


src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+    public record SearchViewTextValues(
+        TextConfiguration Type,
+        string SearchTooltip,
+        string ClearSearchTooltip,
+        string AllSourcesButtonText,
+        string Placeholder,
+        string NoResultsMessage,
+        string RepeatSearchButtonText);
+
+    private static readonly SearchViewTextValues DefaultValues = new(
```
imalcolm1: nit: It might be slightly easier to read `SearchViewCustomization_Text` if `DefaultValues` was renamed to `DefaultTextValues` (and same for `CustomValues` -> `CustomTextValues`)

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+
+        // Check the text values
+#if MAC_TEST
+        Assert.AreEqual(expected.Placeholder, GetEntryText(FindElement("QueryEntry", TimeSpan.FromSeconds(5))), $"Expected the placeholder text to be visible and equal to {expected.Type} value.");
+#else
+        var placeholderElement = FindElementByText(expected.Placeholder, TimeSpan.FromSeconds(5));
+        Assert.IsTrue(placeholderElement.Displayed, $"Expected the placeholder text to be visible and equal to {expected.Type} value.");
+#endif
+#if MAUI_TEST
+        if (!string.Equals(expected.Type.ToString(), "Custom", StringComparison.OrdinalIgnoreCase))
```
imalcolm1: I think we can just compare the enum here rather than converting to a string
```suggestion
        if (expected.Type != TextConfiguration.Custom)
```

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+#endif
+        {
+            // Check the Automation Names on Search and Clear Search buttons
```
imalcolm1: nit: This comment should probably go above both blocks that it covers. Also a note about why they are not run for maui would make it more clear that the if-defs were an intentional choice
```suggestion
// Check the Automation Names on Search and Clear Search buttons (customization of these elements is not supported on Maui)
#if MAUI_TEST
        if (!string.Equals(expected.Type.ToString(), "Custom", StringComparison.OrdinalIgnoreCase))
#endif
        {
```

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+        var repeatSearchHereButton = FindElement("RepeatSearchHereButton", TimeSpan.FromSeconds(5));
+        Assert.AreEqual(expected.RepeatSearchButtonText, repeatSearchHereButton.Text, $"Expected the repeat search here button text to be set to {expected.Type} value.");
+        Assert.AreEqual(expected.RepeatSearchButtonText, GetAutomationName(repeatSearchHereButton), $"Expected the automation name of repeat search here button to be set to {expected.Type} value.");
+        FindElement("ClearSearchButton").Click();
+        FindElement("AddEventTestSourceButton").Click();
+        await ShowAllSourcesButton();
+#if !MAUI_TEST
+        var allSourcesButton = FindElement("AllSourcesButton", TimeSpan.FromSeconds(5));
+        Assert.AreEqual(expected.AllSourcesButtonText, allSourcesButton.Text, $"Expected the all sources button text to be set to {expected.Type} value.");
+        Assert.AreEqual(expected.AllSourcesButtonText, GetAutomationName(allSourcesButton), $"Expected the automation name of all sources button to be set to {expected.Type} value.");
```
imalcolm1:
```suggestion
        Assert.AreEqual(expected.AllSourcesButtonText, GetAutomationName(allSourcesButton), $"Expected the automation name of the all sources button to be set to {expected.Type} value.");
```

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
+using OpenQA.Selenium;
+using OpenQA.Selenium.Appium.Enums;
```
imalcolm1: Not sure whether this is actually true for all platforms, but visual studio is telling me these are unused

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
imalcolm1: Actually, you could also do something like
```c#
bool skipUnsupportedMaui = false;
#if MAUI_TEST
skipUnsupportedMaui  = expected.Type != TextConfiguration.Custom;
#endif

...

if (!skipUnsupportedMaui)
{
    // Check search/clear button text
}
```
To reduce the number of preprocessor statements

src/Tests/UITests/Toolkit.UITests.TestPages.Shared/SearchView/SearchViewCustomization.xaml.cs
```diff
@@ -0,0 +1,162 @@
...
+        MyMapView.SetViewpoint(new Viewpoint(center, scale));
+    }
+
+#if WPF_APP || WINUI_APP
+    private void GeoViewConnection_Checked(object sender, RoutedEventArgs e)
+    {
+        if (EnableGeoViewBindingCheck.IsChecked ?? false)
+#elif MAUI_APP
+
+    private void GeoViewConnection_Checked(object sender, CheckedChangedEventArgs e)
```
imalcolm1: If we have a case where we need platform-specific function declarations to get around different event types, I would follow the same pattern as `ClickedEventArgs` where we define one name for the different event types through our `using` statements. You may still need platform checks inside the function, but it's a bit more readable in my opinion.

I would even suggest replacing the checkbox entirely in favor of two* normal buttons: one that only removes the `GeoView` and one that only adds it. That way we don't need to know anything about the state of the page to be sure what the outcome of clicking a particular button will be.
*It can just be one button if that's all the tests need

src/Tests/UITests/Toolkit.UITests.TestPages.Shared/SearchView/SearchViewCustomization.xaml.cs
```diff
@@ -0,0 +1,162 @@
...
+    private void UpdateToCustomValuesButton_Click(object sender, ClickEventArgs e)
+    {
+#if WINUI_APP || WPF_APP
+        MySearchView.SearchTooltipText = "Custom Search";
+        MySearchView.ClearSearchTooltipText = "Custom Clear Search";
+        MySearchView.AllSourceSelectText = "Custom All Sources";
+#elif MAUI_APP
+        MySearchView.AllSourcesSelectText = "Custom All Sources";
+#endif
+        MySearchView.SearchViewModel?.DefaultPlaceholder = "Custom Find a place or address";
```
imalcolm1: It looks as though this could be simplified slightly
```suggestion
#if !MAUI_APP
        MySearchView.SearchTooltipText = "Custom Search";
        MySearchView.ClearSearchTooltipText = "Custom Clear Search";
#endif
        MySearchView.AllSourcesSelectText = "Custom All Sources";
        MySearchView.SearchViewModel?.DefaultPlaceholder = "Custom Find a place or address";
```

src/Tests/UITests/Toolkit.UITests.TestPages.Shared/SearchView/SearchViewCustomization.xaml.cs
```diff
@@ -0,0 +1,162 @@
...
+using Esri.ArcGISRuntime.Mapping;
+using Esri.ArcGISRuntime.Symbology;
+using Esri.ArcGISRuntime.Tasks.Geocoding;
+using Esri.ArcGISRuntime.UI;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Windows;
```
imalcolm1: A bunch of the imports on this file are unused according to visual studio

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+#if WINUI_TEST
+        Assert.IsTrue(ElementExistsById("SearchSuggestionsListGrouped", TimeSpan.FromSeconds(5)));
+#else
+        Assert.IsTrue(ElementExistsById("SearchSuggestionsList", TimeSpan.FromSeconds(5)));
+#endif
+
+        // Verify the world geocoder result is displayed.
+        Assert.IsTrue(
+            ElementExistsByName(
+                "2000 E Convention Center Way, Ontario, CA, 91764, USA",
```
imalcolm1: Looks like there's already a constant for this address (`OntarioAddress`). I noticed two other places where it's spelled out manually as well

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+        // Toggle the binding setting that controls whether the result list view is enabled.
+        FindElement("EnableIndividualResultDisplayCheck").Click();
+
+        // Show the result list again and verify it is hidden after disabling the binding.
+        await ShowIndividualResultList();
+        Assert.IsTrue(ElementExistsById("SearchResultsList", TimeSpan.FromSeconds(5)), "Expected the results list to be visible");
+    }
+
+    [TestMethod]
+    public async Task SearchViewCustomization_MultipleSources_Rendering()
```
imalcolm1: Possible name change:
```suggestion
    public async Task SearchViewCustomization_MultipleSources_DropDownRenders()
```
Or maybe `_SourcesListRenders`?

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+                "2000 E Convention Center Way, Ontario, CA, 91764, USA"));
+
+        // Verify the event tester suggestion is displayed.
+        Assert.IsTrue(
+            ElementExistsByName(
+                "suggestion one"));
+    }
+
+    [TestMethod]
+    public async Task SearchViewCustomization_EventTestSource_EventFired()
```
imalcolm1, line 336: Given this https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/728 fix I made, this test should probably check that the right number of events are fired as well. Maybe the `VerifyEventTextBlock` could have text like `"Selected events xN; Deselected events xM"` where N and M are the number of times each event has been recieved?

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+        var selectSuggestion = FindElementByName("Airport", TimeSpan.FromSeconds(5));
+        selectSuggestion.Click();
+        await Task.Delay(2000);
+        // Move the map to a new location so the previous search can be repeated in the new visible extent.
+        FindElement("UpdateViewpointExtentToOntario").Click();
+    }
+    private async Task ShowAllSourcesButton()
+    {
+        FindElement("SourceSelectToggle").Click();
+    }
```
imalcolm1: I think this should have a clearer name

```suggestion

    private async Task ExpandSourcesDropDown()
    {
        FindElement("SourceSelectToggle").Click();
    }

```

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+        FindElement("ClearSearchButton").Click();
+        FindElement("AddEventTestSourceButton").Click();
+        await ShowAllSourcesButton();
+#if !MAUI_TEST
+        var allSourcesButton = FindElement("AllSourcesButton", TimeSpan.FromSeconds(5));
+        Assert.AreEqual(expected.AllSourcesButtonText, allSourcesButton.Text, $"Expected the all sources button text to be set to {expected.Type} value.");
+        Assert.AreEqual(expected.AllSourcesButtonText, GetAutomationName(allSourcesButton), $"Expected the automation name of all sources button to be set to {expected.Type} value.");
+#else
+        Assert.IsTrue(ElementExistsByText(expected.AllSourcesButtonText, TimeSpan.FromSeconds(5)), $"Expected the all sources button text to be set to {expected.Type} value.");
+#endif
```
imalcolm1: Why is this logic different between patterns. Can we not just do `ElementExistsByText` for both?

imalcolm1 (changes requested):
I'm very glad to see this particular control being automated! Generally looks good save for a few small things.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    private async Task ShowRepeatSearchHereButton()
+    {
+        // Zoom to the extent of Colorado so the entered query returns expected category suggestions and results.
+        FindElement("UpdateViewpointExtentToColorado").Click();
+        SubmitText(FindElement("QueryEntry"), "airport");
+        var selectSuggestion = FindElementByName("Airport", TimeSpan.FromSeconds(5));
+        selectSuggestion.Click();
+        await Task.Delay(2000);
```
imalcolm1, line 379: As you were talking about, I think a label linked to the GeoView.DrawStatusChanged event could be useful here as a more deterministic way to wait until the zoom has completed

imalcolm1 (commented):
Forgot one thing

src/Tests/UITests/Toolkit.UITests.TestPages.Shared/SearchView/SearchViewCustomization.xaml.cs
SamhithaBurugupalli: In WPF/ WinUI the parameter name is `AllSourceSelectText` and where as for Maui it is `AllSourcesSelectText` . Note that there is not that extra 's'

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
SamhithaBurugupalli: I tried using the draw status changed event. But it did not work out. I found the possible root cause in our implementation. 
After auto-navigation on results, it sets `IgnoreAreaChangesFlag = true` in [MaUI](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/502321907affa44cf866e5e7453fc2017b4c23a6/src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs#L696) and [WPF/WinUI](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/502321907affa44cf866e5e7453fc2017b4c23a6/src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs#L402) , awaits map navigation, then does `Task.Delay(1000)` before clearing the flag.
If automation pans during that 1s window, query-area updates are ignored for requery eligibility, so RepeatSearchHereButton never appears.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
SamhithaBurugupalli: Ah yes, I was initially checking for automation names on WPF and WinUI. But then figured out checking by ElementExistsByText works in Maui. So missed to updated for WPF and WinUI platforms. Thanks for catching that.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
imalcolm1, line 1: There is a bit of flakiness where timeouts are needed to let the page's buttons and other controls load when first opened. I've noted the ones I see.

Also we might want a something like:
```c#
private readonly TimeSpan DefaultTimeout = TimeSpan.FromSeconds(5);
```
as a field to make updating the value easier as well.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,417 @@
...
+        Assert.IsTrue(ElementExistsByText(expected.AllSourcesButtonText, TimeSpan.FromSeconds(5)), $"Expected the all sources button text to be set to {expected.Type} value.");
+    }
+
+    [TestMethod]
+    public async Task SearchViewCustomization_GeoViewBinding()
+    {
+        OpenSample(SearchViewCustomizationPage);
+
+        // Restrict the GeoView to Colorado so the query should not return results near Ontario, California.
+        FindElement("UpdateViewpointExtentToColorado").Click();
```
imalcolm1: This wants a timeout

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    private const string OntarioAddress =
+    "2000 E Convention Center Way, Ontario, CA, 91764, USA";
+
+    private async Task SearchWithEventSource()
+    {
+        FindElement("AddEventTestSourceButton").Click();
+        FindElement("UpdateViewpointExtentToOntario").Click();
+    }
```
imalcolm1: This is called as the first action in several tests so it needs a timeout. I'd suggest making the timeout a parameter so that it's clear to readers that this has some leeway without needing to scroll down to the implementation.
```suggestion
    private async Task SearchWithEventSource(TimeSpan timeout)
    {
        FindElement("AddEventTestSourceButton", timeout).Click();
        FindElement("UpdateViewpointExtentToOntario", timeout).Click();
    }
```

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -0,0 +1,417 @@
...
+
+    private async Task ShowIndividualResultList()
+    {
+        await ShowResultListView();
+
+        // Select a known search result from the result list
+        var selectedResult = FindElementByName("Colorado Springs Airport, Colorado Springs, Colorado", TimeSpan.FromSeconds(5));
+        selectedResult.Click();
+    }
```
imalcolm1: Same as `SearchWithEventSource()`

```suggestion
    private async Task ShowResultListView(TimeSpan timeout)
    {
        // Zoom to the extent of Colorado so the entered query returns expected category suggestions and results.
        FindElement("UpdateViewpointExtentToColorado", timeout).Click();
        SubmitText(FindElement("QueryEntry", timeout), "airport");
        var selectSuggestion = FindElementByName("Airport", timeout);
        selectSuggestion.Click();
    }

    private async Task ShowIndividualResultList(TimeSpan timeout)
    {
        await ShowResultListView(timeout);

        // Select a known search result from the result list
        var selectedResult = FindElementByName("Colorado Springs Airport, Colorado Springs, Colorado", timeout);
        selectedResult.Click();
    }
```

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
```diff
@@ -410,5 +404,14 @@ private void SubmitOntarioAddressQuery()
...
 
+    private static bool ShouldValidateButtonAutomationNames(TextConfiguration type)
+    {
+#if MAUI_TEST
+        return type != TextConfiguration.Custom;
+#else
+        return true;
+#endif
+    }
+
```
imalcolm1: I think this could be inlined. It's just 4 lines
```c#
var shouldValidateButtonAutomationNames = true;
#if MAUI_TEST
    shouldValidateButtonAutomationNames = expected.Type != TextConfiguration.Custom;
#endif
```

imalcolm1 (commented):
Just needs a couple timeouts

src/Tests/UITests/Toolkit.UITests.Shared/Tests/SearchView/SearchViewCustomization.cs
imalcolm1: Ah good find! There should probably be a comment justifying the Task.Delay in that case

imalcolm1 (approved):

src/Toolkit/Toolkit.Maui/SearchView/SearchView.Appearance.cs
mstefarov: The `repeatSearchDescription` variable is dead code now, right? XAML doesn't use it any more.

mstefarov (approved):
Very solid 👍 

I have only one little nitpick to offer, otherwise this is great

PR #780, opened 2026-06-22
UITests: Small tweak to viewpoint on one of the SearchView tests

imalcolm1:
The `SearchViewControl_SearchSuggestions` test used the `UpdateViewpointExtentToUSA` to take the map to a viewpoint that was expected to cover ontario international airport. The existing viewpoint was a bit too close to the center of the US and did not include ontario as the CI test device is very narrow, which caused the test to fail. This PR just shifts the viewpoint west.


SamhithaBurugupalli (approved):
Thanks for this fix!!

PR #781, opened 2026-06-23
Fix SearchView source race when default geocoder config reruns

SamhithaBurugupalli:
Fixes a SearchView race where custom search sources could be removed when default world geocoder configuration reran (for example after binding/GeoView changes).
ConfigureDefaultWorldGeocoder is now non-destructive and only ensures the default source exists, preserving custom sources.

- Caches the default world geocoder source in SearchViewModel
- Adds the default source only if it is missing.
- Synchronizes access to avoid race conditions during concurrent/lifecycle-triggered configuration.

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/SearchView/SearchViewModel.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/SearchView/SearchViewModel.cs
@@ -56,9 +56,11 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
 
         private bool _searchInProgress;
         private bool _suggestInProgress;
+        private LocatorSearchSource? _defaultWorldGeocoderSource;
 
         private CancellationTokenSource? _activeSearchCancellation;
         private CancellationTokenSource? _activeSuggestCancellation;
+        private readonly object _defaultSourceConfigurationSync = new();
 
         /// <summary>
         /// Gets or sets the active search source, if one is selected. If there is no selection, all sources will be used for the search.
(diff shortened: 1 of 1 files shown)


imalcolm1 (approved):

src/Toolkit/Toolkit/UI/Controls/SearchView/SearchViewModel.cs
```diff
@@ -56,9 +56,11 @@ public partial class SearchViewModel : INotifyPropertyChanged
 
         private bool _searchInProgress;
         private bool _suggestInProgress;
+        private LocatorSearchSource? _defaultWorldGeocoderSource;
 
         private CancellationTokenSource? _activeSearchCancellation;
         private CancellationTokenSource? _activeSuggestCancellation;
+        private readonly object _defaultSourceConfigurationSync = new();
```
mstefarov, line 63: One small musing, not a change request:

I just want to point out that the real fix here is dropping the `Sources.Clear()` and adding the "only add the default if it's missing" check. Caching the geocoder source is part of that same idea, since CreateDefaultSourceAsync returns a new object each time and you need a stable instance for the Contains check to work. Nice work running down the root cause! 👍 

The locking is more of a defensive layer on top. In practice this method only ever runs on the UI thread (it's all driven by bindings and GeoView events). So the "race" is really two async calls interleaving on the UI thread rather than two threads fighting, and locks don't help with that.  It's not harmful, it's just guarding against something that can't happen under normal use. The rest of the toolkit is written with "controls are meant to be used from the UI-thread" convention instead of locks for this kind of code.

mstefarov (approved):

PR #784, opened 2026-06-28
[IOS] [Bug fix]: Fixes Display of floor visibility in the control, When zoomed into a building

SamhithaBurugupalli:
Issue:  What is expected is that when I zoom in the floor filter control should expand vertically to show the floors in the current building (1, 2, 3, etc). When I zoom in and out there is a very subtle change in the height of the control like it's trying to do something, but no floors are actually shown. Otherwise the navigate site button and zoom to building button work as intended.
Fix: There is a typo in pre-processor statement for ios. So fixed that


imalcolm1 (approved):
Very happy to see this fixed!

mandel-capuchin (approved):

PR #785, opened 2026-06-29
Add OrientedImageryView (WPF only)

imalcolm1:
Adds the upcoming `OrientedImageryView` control. This control builds on the `OrientedImageDisplay`, from the `dev/orientedimageryview` branch. Where the `OrientedImageDisplay` class handles the low-level details of displaying an oriented image (either planar or 360), the `OrientedImageryView` provides a default preconfigured workflow for users mimicking the oriented imagery UIs provided by Online and Pro.

## Notes
- See the sample for how this control is meant to be used. Also see the [documentation file](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/ian/orientedimageryview-green/docs/oriented-imagery-view.md)
- Includes extracting the custom FeatureFormView.Command class and renaming its members to better align with the definition of the equivalent Maui class

## Data for testing
Esri Campus (Planar): https://services.arcgis.com/2Pv4ow3pE6NFC9SW/arcgis/rest/services/EsriCampus_GoProSample/FeatureServer/0

Hopeful Heights (Panoramic): https://services8.arcgis.com/joda3ARQ9znLf0hf/ArcGIS/rest/services/HopefulHeights_360/FeatureServer/1

There are many more OrientedImageryLayers on AGOL if you filter by layer type. Video and some raster formats are not supported.

### DIFF
--- a/docs/oriented-imagery-view.md
+++ b/docs/oriented-imagery-view.md
@@ -1,0 +1,155 @@
+# OrientedImageryView
+
+`OrientedImageryView` displays oriented imagery with a built-in toolbar for image navigation, footprint visibility, camera location markers, and image markers. It works with an `OrientedImageryLayer` and `GeoView`. The control manages footprints and marker overlays, while your app remains responsible for searching for images and choosing which image is selected.
+
+![Example Image](images/oriented-imagery.png)
+
+## Minimal implementation
+
+The following WPF example places a `MapView` next to an `OrientedImageryView`. Set `GeoView` so the control can display its managed marker overlay on the map, and set `OrientedImageryLayer` so the view model can manage footprints and image-to-map calculations.
+
+```xml
+<Grid xmlns:esri="http://schemas.esri.com/arcgis/runtime/2013">
+    <Grid.ColumnDefinitions>
+        <ColumnDefinition Width="*" />
+        <ColumnDefinition Width="400" />
+    </Grid.ColumnDefinitions>
+
+    <esri:MapView x:Name="MainMapView" />
+    <esri:OrientedImageryView x:Name="MainOrientedImageryView"
+                              Grid.Column="1"
+                              GeoView="{Binding ElementName=MainMapView}" />
+</Grid>
+```
+
+Create the map and oriented imagery layer. Search for oriented images when the map is tapped.
+
+```cs
+private const string OrientedImageryLayerUrl = "https://example.com/arcgis/rest/services/OrientedImagery/FeatureServer";
+
+private OrientedImageryLayer? _orientedImageryLayer;
+
+private async Task InitializeAsync()
+{
+    MainMapView.Map = new Map(BasemapStyle.ArcGISImageryStandard);
+
+    _orientedImageryLayer = new OrientedImageryLayer(new Uri(OrientedImageryLayerUrl));
+    MainMapView.Map.OperationalLayers.Add(_orientedImageryLayer);
+
+    MainOrientedImageryView.OrientedImageryLayer = _orientedImageryLayer;
+
+    await _orientedImageryLayer.LoadAsync();
+
+    if (_orientedImageryLayer.FullExtent is not null)
+    {
+        await MainMapView.SetViewpointGeometryAsync(_orientedImageryLayer.FullExtent);
+    }
+
+    MainMapView.GeoViewTapped += MainMapView_GeoViewTapped;
+}
+
+private async void MainMapView_GeoViewTapped(object? sender, GeoViewInputEventArgs e)
+{
+    if (_orientedImageryLayer is null || e.Location is null)
+    {
+        return;
+    }
+
+    // Add a marker if the `OrientedImageryViewModel.AllowAddingMarkers` is enabled.
+    if (MainOrientedImageryView.ViewModel.AllowAddingMarkers)
+    {
+        MainOrientedImageryView.ViewModel.AddMarkerLocation(e.Location);
+        return;
+    }
+
+    // Otherwise search for images that view the clicked location.
+    var parameters = new OrientedImageSearchParameters { MaxResults = -1 };
+    var images = await _orientedImageryLayer.SearchImagesAsync(e.Location, parameters) ?? new List<OrientedImage>();
+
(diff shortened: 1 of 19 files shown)


imalcolm1:
Hello all, for this initial review I'm mostly hoping to get feedback on the api surface of the control, so reviews don't need to be super in depth. The best place to start is probably the sample and the [readme](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/ian/orientedimageryview-green/docs/oriented-imagery-view.md) since those demonstrate how the control is used. At least for this review we can probably ignore implementation details for the `OrientedImageView` control that this class takes advantage of.

Keeping this in draft since I have no plans to merge until the `dev/orientedimageryview` is merged.

gdestigter (commented):
I'm trying to get my head around this and have a general question.

How are you thinking about the split between `OrientedImageryView` and `OrientedImageryViewModel`? Right now consumers set things like `GeoView` and `OrientedImageryLayer` on the view, but work with images, selection, footprints, and markers through the view model. Is that division intentional, and which type do you expect to be the primary public API when this is implemented for WinUI and MAUI?

Do you expect the view model to become a shared cross-platform API, or have more of the functionality be exposed as bindable properties on the control?

SamhithaBurugupalli (commented):
I haven't done a detailed review yet, but looks ok to me from what I understand

src/Toolkit/Toolkit/UI/Controls/OrientedImagery/OrientedImageryView.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        if (newValue == null)
+        {
+            SetCurrentValue(ViewModelProperty, new OrientedImageryViewModel());
+            return;
+        }
+
+        newValue.PropertyChanged += ViewModel_PropertyChanged;
+        DataContext = newValue;
```
mandel-capuchin: can we leave DataContext alone here? this replaces the app’s inherited data context, so bindings can resolve against the wrong view model.

src/Toolkit/Toolkit/UI/Controls/OrientedImagery/OrientedImageryViewModel.cs
```diff
@@ -0,0 +1,594 @@
...
+            if (_oiLayer == value) return;
+
+            if (_oiLayer != null)
+            {
+                _oiLayer.VisibleFootprints.Clear();
+                _oiLayer.PropertyChanged -= OrientedImageryLayer_PropertyChanged;
+                _oiLayerSceneProperties!.PropertyChanged -= OrientedImageryLayer_SceneProperties_PropertyChanged;
+            }
+
+            _images.Clear();
```
mandel-capuchin, line 85: should we clear _footprints here too? setting SelectedImage to null can add the old footprints back, and they can then be added to the new layer.

src/Toolkit/Toolkit/UI/Controls/OrientedImagery/OrientedImageryViewModel.cs
```diff
@@ -0,0 +1,649 @@
...
+
+    /// <summary>
+    /// Sets the images to display in the control.
+    /// </summary>
+    /// <remarks>
+    /// Pass the <paramref name="searchPoint"/> parameter to display a marker at the location from which the images were searched.
+    /// </remarks>
+    /// <param name="images">The oriented images to display.</param>
+    /// <param name="searchPoint">The point from which the images were searched.</param>
+    public void SetImages(List<OrientedImage> images, MapPoint? searchPoint = null)
```
mandel-capuchin: can this take IReadOnlyList<OrientedImage> or IEnumerable<OrientedImage> instead? SearchImagesAsync returns an IReadOnlyList, so the documented example does not compile.”

src/Toolkit/Toolkit.WPF/UI/Controls/OrientedImagery/OrientedImagery.Theme.xaml
```diff
@@ -1,7 +1,253 @@
...
+
+    <!-- Helper styles -->
+    <Style x:Key="ToolbarButtonStyle" TargetType="{x:Type Button}">
+        <Setter Property="MinWidth" Value="40" />
+        <Setter Property="MinHeight" Value="40" />
+        <Setter Property="Padding" Value="4" />
+        <Setter Property="Background" Value="Transparent" />
+        <Setter Property="BorderBrush" Value="Transparent" />
+        <Setter Property="BorderThickness" Value="1" />
+        <Setter Property="FocusVisualStyle" Value="{x:Null}" />
```
mandel-capuchin: can we keep a visible keyboard focus state here? these styles remove it without adding a replacement. the previous/next icon buttons also need localized accessible names.

docs/oriented-imagery-view.md
```diff
@@ -0,0 +1,158 @@
...
+    MainOrientedImageryView.ViewModel.SetImages(images, e.Location);
+    MainOrientedImageryView.ViewModel.SelectedImage = images.FirstOrDefault();
+}
+```
+
+See the `OrientedImageryView` sample for a full example implementation.
+
+## Customizing the toolbar
+
+`OrientedImageryView` is structured as an `ItemsControl` whose toolbar items are defined by its `ItemsSource` property. The default toolbar items are returned by `OrientedImageryView.GetDefaultToolbarItems()`, which includes controls for auto-updating the footprint, showing selected and unselected footprints, cycling camera marker display modes, enabling marker creation, selecting the marker symbol, and clearing markers.
```
mandel-capuchin, line 78: this section needs an update after the toolbar changes. OrientedImageryView is now a Control, toolbar items are in ViewModel.ToolbarItems, and the default list does not include auto-update footprint. the examples below also use the old ItemTemplateSelector and MainViewModel names.

mandel-capuchin (changes requested):

docs/oriented-imagery-view.md
imalcolm1: Going to hold off for a couple of days since I expect at least the screenshot on that file to need changing, but I'll make sure to update it before I leave

PR #786, opened 2026-07-03
UITests: Remove old automation workaround as it was breaking the ScaleLine tests on MauiWindows

imalcolm1:
The [maui issue](https://github.com/dotnet/maui/issues/4715) was solved in 10.0.80, and the workaround seems to have broken the scale line tests once we started in MauiVersion 10.0.80. This PR avoids the automation fix when MauiVersion is >= 10.0.80. It also adds MauiVersion to the list of versions in the Maui app for help with manual troubleshooting


SamhithaBurugupalli (approved):

SamhithaBurugupalli (approved):
This will work on the daily test runs. But might not work locally if the nuget package has not been updated. So, might be better to update nuget packages as well (In the same/ a different PR)

imalcolm1:
That's a very good point, I've changed the PR to include the fix conditionally rather than just removing it

SamhithaBurugupalli (approved):
Nice way to handle any scenario!!

PR #788, opened 2026-07-06
Ignore AGOL org when comparing basemap source URLs

imalcolm1:
Ignore runtime URL when comparing layer equality

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
@@ -349,7 +349,10 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
 
             var host = uri.Host;
             return host.Equals("arcgis.com", StringComparison.OrdinalIgnoreCase)
-                || host.EndsWith(".arcgis.com", StringComparison.OrdinalIgnoreCase);
+                || host.Equals("www.arcgis.com", StringComparison.OrdinalIgnoreCase)
+                || host.Equals("cdn.arcgis.com", StringComparison.OrdinalIgnoreCase)
+                || host.Equals("maps.arcgis.com", StringComparison.OrdinalIgnoreCase)
+                || host.EndsWith(".maps.arcgis.com", StringComparison.OrdinalIgnoreCase);
         }
 
         /// <inheritdoc />
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
@@ -292,8 +292,9 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
                     return webTiledLayer.TemplateUri == ((WebTiledLayer)layer2).TemplateUri && webTiledLayer.SubDomains.SequenceEqual(((WebTiledLayer)layer2).SubDomains);
                 case ServiceImageTiledLayer serviceImageTiledLayer:
 #pragma warning disable CS0618 // Type or member is obsolete
-                    if (layer1 is not BingMapsLayer)
-                        return false;
+                    // Handle BingMapsLayer in URL comparisons below
+                    if (layer1 is BingMapsLayer)
+                        break;
 #pragma warning restore CS0618 // Type or member is obsolete
                     return serviceImageTiledLayer.TileInfo == ((ServiceImageTiledLayer)layer2).TileInfo;
                 case SubtypeFeatureLayer subtypeFeatureLayer:
@@ -335,7 +336,8 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
 
             if (layer1.Item != null && layer2.Item != null)
             {
-                if (layer1.Item.ItemId == layer2.Item.ItemId || PortalHelper.GetPortalItemId((layer1.Item as PortalItem)?.Url) == PortalHelper.GetPortalItemId((layer2.Item as PortalItem)?.Url))
+                var id1 = PortalHelper.GetPortalItemId((layer1.Item as PortalItem)?.Url);
+                if (layer1.Item.ItemId == layer2.Item.ItemId || (id1 != null && id1 == PortalHelper.GetPortalItemId((layer2.Item as PortalItem)?.Url)))
                     return true;
             }
 
--- a/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
@@ -322,9 +322,8 @@ namespace Esri.ArcGISRuntime.Toolkit.UI
                 return true;
 
             // Two AGOL items may still be equivalent if their URLs are the same but refer to different orgs. For example,
-            // https://runtime.maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405, https://maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405,
-            // and https://arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405 all point to the same item
-            // despite having different hostnames
+            // https://runtime.maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405 and
+            // https://maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405 point to the same item
             if (IsArcGisComHost(source1) && IsArcGisComHost(source2) && source1?.PathAndQuery == source2?.PathAndQuery)
             {
                 return true;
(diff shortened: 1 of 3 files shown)


src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -279,31 +279,65 @@ private static bool LayerEquals(Layer layer1, Layer layer2)
...
+        }
+
+        private static bool IsArcGisComHost(Uri? uri)
+        {
+            if (uri == null)
+                return false;
+
+            var host = uri.Host;
+            return host.Equals("arcgis.com", StringComparison.OrdinalIgnoreCase)
+                || host.EndsWith(".arcgis.com", StringComparison.OrdinalIgnoreCase);
```
mstefarov: I would recommend tightening this query a bit to exclude dev-testing/staging AGOL environments:

```suggestion
            return host.Equals("arcgis.com", StringComparison.OrdinalIgnoreCase)
                || host.Equals("www.arcgis.com", StringComparison.OrdinalIgnoreCase)
                || host.Equals("cdn.arcgis.com", StringComparison.OrdinalIgnoreCase)
                || host.Equals("maps.arcgis.com", StringComparison.OrdinalIgnoreCase)
                || host.EndsWith(".maps.arcgis.com", StringComparison.OrdinalIgnoreCase);
```

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -279,31 +279,65 @@ private static bool LayerEquals(Layer layer1, Layer layer2)
...
+
+            if (source1 == null || source2 == null)
+                return false;
+            if (source1 == source2)
+                return true;
+
+            // Two AGOL items may still be equivalent if their URLs are the same but refer to different orgs. For example,
+            // https://runtime.maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405 and
+            // https://maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405 point to the same item
+            if (IsArcGisComHost(source1) && IsArcGisComHost(source2) && source1?.PathAndQuery == source2?.PathAndQuery)
```
mstefarov, line 334: We might find a few different patterns here: `item.html?id=<guid>` or `/sharing/rest/content/items/<guid>` or webmap viewer URL. See [Map(Uri) doc for examples](https://developers.arcgis.com/net/api-reference/api/netwin/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Mapping.Map.-ctor.html#Esri_ArcGISRuntime_Mapping_Map__ctor_System_Uri_). Rather than comparing PathAndQuery, consider extracting item IDs from the URL. Actually, [OfflineMapViewModel](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/main/src/Toolkit/Toolkit/UI/Controls/OfflineMapAreas/OfflineMapViewModel.cs#L81-L98) already has code that does this, maybe we can move it to a shared helper.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -279,31 +279,65 @@ private static bool LayerEquals(Layer layer1, Layer layer2)
...
-                ArcGISTiledLayer tiledLayer => tiledLayer.Source == ((ArcGISTiledLayer)layer2).Source,
-                ArcGISVectorTiledLayer vectorTiledLayer => vectorTiledLayer.Source == ((ArcGISVectorTiledLayer)layer2).Source,
+                case FeatureCollectionLayer featureCollectionLayer:
+                    return featureCollectionLayer.FeatureCollection == ((FeatureCollectionLayer)layer2).FeatureCollection;
+                case GroupLayer groupLayer:
+                    return groupLayer.Layers.Count == ((GroupLayer)layer2).Layers.Count;
+                case OpenStreetMapLayer openStreetMapLayer:
+                    return true; // OpenStreetMap layers are considered equal if types match
+                case ServiceImageTiledLayer serviceImageTiledLayer:
+                    return serviceImageTiledLayer.TileInfo == ((ServiceImageTiledLayer)layer2).TileInfo;
```
mstefarov, line 299: This would early-return for WebTiledLayer and BingMapsLayer too (they derive from ServiceImageTiledLayer). So the corresponding branches in the second switch are not reachable.

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -279,31 +279,65 @@ private static bool LayerEquals(Layer layer1, Layer layer2)
...
+                case SubtypeFeatureLayer subtypeFeatureLayer:
+                    return subtypeFeatureLayer.FeatureTable == ((SubtypeFeatureLayer)layer2).FeatureTable;
+                case FeatureLayer featureLayer:
+                    return featureLayer.FeatureTable == ((FeatureLayer)layer2).FeatureTable;
+            }
+
+            var (source1, source2) = layer1 switch
```
mstefarov, line 306: Did you mean to drop GroupLayer recursion from this switch?

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -279,31 +279,65 @@ private static bool LayerEquals(Layer layer1, Layer layer2)
...
-                SubtypeFeatureLayer subtypeFeatureLayer => subtypeFeatureLayer.FeatureTable == ((SubtypeFeatureLayer)layer2).FeatureTable,
-                WmsLayer wmsLayer => wmsLayer.Source == ((WmsLayer)layer2).Source,
-                WmtsLayer wmtsLayer => wmtsLayer.Source == ((WmtsLayer)layer2).Source,
-                _ => false,
+                DimensionLayer dimensionLayer => (dimensionLayer.Source, ((DimensionLayer)layer2).Source),
+                IntegratedMeshLayer integratedMeshLayer => (integratedMeshLayer.Source, ((IntegratedMeshLayer)layer2).Source),
+                KmlLayer kmlLayer => (kmlLayer.Dataset?.Source, ((KmlLayer)layer2).Dataset?.Source),
+                Ogc3DTilesLayer ogc3DTilesLayer => (ogc3DTilesLayer.Source, ((Ogc3DTilesLayer)layer2).Source),
+                PointCloudLayer pointCloudLayer => (pointCloudLayer.Source, ((PointCloudLayer)layer2).Source),
+                WebTiledLayer webTiledLayer => (new Uri(webTiledLayer.TemplateUri), new Uri(((WebTiledLayer)layer2).TemplateUri)),
```
mstefarov: Note that TemplateUri strings are _not necessarily valid URIs_ and this line can throw a UriFormatException.  Templates contain placeholders like `https://{subDomain}.tile.example.com/{level}/{col}/{row}.png`

SamhithaBurugupalli (approved):

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
```diff
@@ -279,31 +280,79 @@ private static bool LayerEquals(Layer layer1, Layer layer2)
...
+
+            // Two AGOL items may still be equivalent if their URLs are the same but refer to different orgs. For example,
+            // https://runtime.maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405 and
+            // https://maps.arcgis.com/home/item.html?id=55ebf90799fa4a3fa57562700a68c405 point to the same item
+            if (IsArcGisComHost(source1) && IsArcGisComHost(source2) && source1?.PathAndQuery == source2?.PathAndQuery)
+                return true;
+
+            if (layer1.Item != null && layer2.Item != null)
+            {
+                if (layer1.Item.ItemId == layer2.Item.ItemId || PortalHelper.GetPortalItemId((layer1.Item as PortalItem)?.Url) == PortalHelper.GetPortalItemId((layer2.Item as PortalItem)?.Url))
```
mstefarov: Since GetPortalItemId returns null when it can't find an ID, this check would pass on double ID-extraction failure (`null == null` is true). Similarly, GetPortalItemId will pass through a null if a layer come from a LocalItem (`as PortalItem` became null).

Just null-check one of the IDs before comparing them, and both false-positive scenarios are fixed:

```suggestion
                var id1 = PortalHelper.GetPortalItemId((layer1.Item as PortalItem)?.Url);
                if (layer1.Item.ItemId == layer2.Item.ItemId || (id1 is not null && id1 == PortalHelper.GetPortalItemId((layer2.Item as PortalItem)?.Url)))
```

src/Toolkit/Toolkit/UI/Controls/BasemapGallery/BasemapGalleryItem.cs
mstefarov: BingMapsLayer will still hit this ServiceImageTiledLayer case and return, which makes the BingMapsLayer case below dead code. It never reaches the `Portal?.Uri` check and instead just does the `TileInfo` equality check.

I would expect TileInfo equality to always be false, actually. It does not override equals and does not implement InstanceID, so each layer will have a unique-looking TileInfo object.

So I expect that Bing layers never compare equal with the current code.

imalcolm1: Ah that was a silly mistake, fixed!

mstefarov (approved):

PR #790, opened 2026-07-07
UITests: Fix screenshot rotation on iOS and increase color tolerance for identifying Compass north arrow

imalcolm1:
The Compass_Rotates (Heading Bound) test was failing consistently on Mac in the daily tests. Increasing the tolerance helped the test pass on the CI machine. In the process of testing, it seemed that iOS was failing all compass tests on a particular device. This was apparently due to problems with rotation metadata, hence the fix being to call AutoOrient.


SamhithaBurugupalli (approved):
LGTM

PR #793, opened 2026-07-10
Fix FeatureForm attachment element rendering and header layout in WPF, WinUI, and MAUI

SamhithaBurugupalli:
Issue: Attachment form elements are not rendering on all platforms. 
Repro: Replace the map in feature form sample with https://runtimecoretest.maps.arcgis.com/home/item.html?id=7064081d2d1a4af6b871d35954417e5e and select the feature on map to open it's form

- Added explicit rendering support for `AttachmentsFormElement` in the Windows form-element host so attachment elements are no longer skipped.
- Wired attachment element templates into both top-level and grouped form element pipelines
- Added MAUI selector support for `AttachmentsFormElement`
- Fixed attachment header label/description overlap in WPF and WinUI attachment element templates.


imalcolm1 (approved):
Looks good save for the preexisting issues we talked about

dotMorten (approved):

PR #794, opened 2026-07-13
Update MeasureToolbar to use new GeometryEditor.InteractionPreviewChanged event

duffh:
Update toolkit implementation to use new `GeometryEditor.InteractionPreviewChanged` event to display measure information during ongoing interactions.

If the pointer leaves the map view or the interaction ends (for example, a vertex is hovered) then the measurement value falls back to the `GeometryEditor.Geometry`.

### DIFF
--- a/src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
@@ -400,7 +400,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         /// instance.
         /// </summary>
         /// <remarks>This method updates the state of the clear button based on the <see
-        /// cref="GeometryEditor.Geometry"/> property and displays the updated geometry result. The <paramref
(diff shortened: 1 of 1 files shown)


src/Toolkit/Toolkit.UI.Controls/MeasureToolbar/MeasureToolbar.cs
```diff
@@ -414,14 +414,38 @@ private void OnGeometryEditorPropertyChanged(object? sender, PropertyChangedEven
                 {
                     _clearButton.IsEnabled = editor?.Geometry is Geometry.Geometry geometry && !geometry.IsEmpty;
                 }
-                DisplayResult(editor?.Geometry);
```
dotMorten, line 417: I would think this should stay in case the geometry property gets set up front or other cases where it is updated programmatically and not on mouse-move.

duffh: That's a fair point, if the geometry were set up front or it is updated programmatically while the pointer is not on the map view this call would cover that. I'll replace it.

dotMorten (approved):

PR #798, opened 2026-07-15
[Feature Form] [Attachments]: DisplayFilename changes

SamhithaBurugupalli:
Changes are related to the [recent attachment enhancements](https://devtopia.esri.com/runtime/dotnet-api/issues/14577). Creating small feature wise PRs for easy review

Displays the file name in attachments thumbnail only if `DisplayFilename` value is true

Note: 
Webmap used for testing - https://runtimecoretest.maps.arcgis.com/home/item.html?id=7064081d2d1a4af6b871d35954417e5e
The changes from [this PR](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/793) should be merged into the current branch for testing

https://github.com/user-attachments/assets/78d015f4-c9e5-4b66-a92f-b573b1bc79f7

PR #799, opened 2026-07-15
[Feature Form] [Attachments]:  AllowUserRename changes

SamhithaBurugupalli:
Changes are related to the [recent attachment enhancements](https://devtopia.esri.com/runtime/dotnet-api/issues/14577). Creating small feature wise PRs for easy review

Shows/ allows the user to rename the attachments only if `AllowUserRename` value is true

Note:
Webmap used for testing - https://runtimecoretest.maps.arcgis.com/home/item.html?id=7064081d2d1a4af6b871d35954417e5e
The changes from https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/793 should be merged into the current branch for testing. Also, there is a binding issue in WinUI which is fixed in [this PR](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/798)

https://github.com/user-attachments/assets/dd1a5c15-1cce-484f-aaf4-4dad4616b230

### DIFF
--- a/src/Toolkit/Toolkit/UI/Controls/FeatureForm/FormAttachmentView.Maui.cs
+++ b/src/Toolkit/Toolkit/UI/Controls/FeatureForm/FormAttachmentView.Maui.cs
@@ -69,6 +69,16 @@ namespace Esri.ArcGISRuntime.Toolkit.Maui.Primitives
 
         [SupportedOSPlatform("windows")]
         [SupportedOSPlatform("maccatalyst")]
+        private void FormAttachmentView_PropertyChanged(object? sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(Element))
+            {
+                ConfigureFlyout();
+            }
+        }
+
+        [SupportedOSPlatform("windows")]
+        [SupportedOSPlatform("maccatalyst")]
         private void ConfigureFlyout()
         {
             MenuFlyout flyout = new MenuFlyout();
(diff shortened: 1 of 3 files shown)


src/Toolkit/Toolkit/UI/Controls/FeatureForm/FormAttachmentView.Maui.cs
```diff
@@ -233,14 +247,17 @@ private async Task<bool> DisplayAttachmentActionSheet()
...
 
             string delete = Properties.Resources.GetString("FeatureFormRemoveAttachmentMenuItem")!;
-            string rename = Properties.Resources.GetString("FeatureFormRenameAttachmentMenuItem")!;
             string open = Properties.Resources.GetString("FeatureFormOpenAttachmentMenuItem")!;
-            var result = await page.DisplayActionSheetAsync(Attachment.Name, null, null, delete, rename, open);
+            string? rename = Element?.AllowUserRename == true ? Properties.Resources.GetString("FeatureFormRenameAttachmentMenuItem")! : null;
+            var result = rename is not null
+                ? await page.DisplayActionSheetAsync(Attachment.Name, null, null, delete, rename, open)
+                : await page.DisplayActionSheetAsync(Attachment.Name, null, null, delete, open);
+
```
dotMorten, line 255: This feels a little backwards to get the string only if supported. Why not switch on ` Element?.AllowUserRename == true` instead? I think that reads a little easier what the purpose is

So this would read something like
```
 var result = Element?.AllowUserRename == true
                ? await page.DisplayActionSheetAsync(Attachment.Name, null, null, delete, rename, open)
                : await page.DisplayActionSheetAsync(Attachment.Name, null, null, delete, open);
```

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FormAttachmentView.Maui.cs
```diff
@@ -233,14 +247,17 @@ private async Task<bool> DisplayAttachmentActionSheet()
...
             if (result == delete)
             {
                 DeleteAttachment();
             }
-            else if(result == rename)
+            else if(rename is not null && result == rename)
```
dotMorten, line 260: You don't really need this check. It would never be rename if that wasn't an option.

dotMorten (commented):
Overall looks good. Just some minor logic comments

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FormAttachmentView.Windows.cs
imalcolm1, line 225: This seems like it was probably a preexisting issue, but on WinUI it seems that `Element` is always null here so the dialog never shows (debug and release)

imalcolm1: Confirmed fixed in the linked PR :+1:

imalcolm1 (approved):

src/Toolkit/Toolkit/UI/Controls/FeatureForm/FormAttachmentView.Maui.cs
SamhithaBurugupalli: The `rename is not null` guard is useful because `DisplayActionSheetAsync` can return `null `when the sheet is dismissed (for example, tapping outside/back), and we pass null for both cancel and destruction labels. If we removed the guard and used only `result == rename`, then when rename is not offered:


rename is null
result can also be null on dismiss
result == rename would be true

the rename branch would run incorrectly

Attached is the video of this behavior

https://github.com/user-attachments/assets/2183a0eb-cb4f-4a73-8eb5-e413c0439e0d

dotMorten (approved):

PR #809, opened 2026-07-22
Fix for FeatureForm datetime element update

twakim:
Fix update datatime issue.
This fixes issue where after updating a datetime element, the new time was being added old time before being set on the element.


dotMorten (approved):
OK if Matvei's fix doesn't already address this

PR #812, opened 2026-07-29
feat(localization): update pseudo translation

jona7150:
Updated translation using Jenkins. Please review and merge @mstefarov @dotMorten 

 CC: @michi6571 @MasaGithub88


dotMorten (approved):
Looks great. Will merge once we get the updated final translations.

PR #835, opened 2026-08-13
BookmarksView accessibility and tests

gdestigter:
Updates `BookmarksView` with:
- accessibility properties
- UI tests for verification

### DIFF
--- a/src/Tests/UITests/Toolkit.UITests.Maui.App/TestPages/BookmarksView/BookmarksTemplateMap.xaml
+++ b/src/Tests/UITests/Toolkit.UITests.Maui.App/TestPages/BookmarksView/BookmarksTemplateMap.xaml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<local:TestPage
+    x:Class="Toolkit.UITests.App.TestPages.BookmarksTemplateMap"
+    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    xmlns:esri="http://schemas.esri.com/arcgis/runtime/2013"
+    xmlns:local="clr-namespace:Toolkit.UITests.App.TestPages"
+    xmlns:mapping="clr-namespace:Esri.ArcGISRuntime.Mapping;assembly=Esri.ArcGISRuntime"
+    xmlns:toolkit="clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui;assembly=Esri.ArcGISRuntime.Toolkit.Maui">
+    <local:TestPage.Resources>
+        <DataTemplate x:Key="RuntimeItemTemplate" x:DataType="mapping:Bookmark">
+            <Label Text="{Binding Name, StringFormat='Runtime: {0}'}" />
+        </DataTemplate>
+    </local:TestPage.Resources>
+    <Grid>
+        <esri:MapView x:Name="MainMapView" />
+        <Border Background="{StaticResource CalciteBackground1Brush}"
+                HeightRequest="220"
+                HorizontalOptions="Center"
+                Margin="8"
+                MaximumHeightRequest="220"
+                MaximumWidthRequest="300"
+                MinimumHeightRequest="220"
+                MinimumWidthRequest="300"
+                Padding="8"
+                Stroke="{StaticResource CalciteBorder1Brush}"
+                StrokeThickness="1"
+                VerticalOptions="End"
+                WidthRequest="300">
+            <Grid RowDefinitions="Auto,*" RowSpacing="4">
+                <Button AutomationId="SetRuntimeTemplateButton" Clicked="SetRuntimeTemplateButton_Click" Text="Apply runtime" />
+                <toolkit:BookmarksView x:Name="BookmarksView" Grid.Row="1">
+                    <toolkit:BookmarksView.ItemTemplate>
+                        <DataTemplate x:DataType="mapping:Bookmark">
+                            <Border Background="Gold" Padding="4" Margin="2">
+                                <Label Text="{Binding Name, StringFormat='Custom: {0}'}" />
+                            </Border>
+                        </DataTemplate>
+                    </toolkit:BookmarksView.ItemTemplate>
+                </toolkit:BookmarksView>
+            </Grid>
+        </Border>
+    </Grid>
+</local:TestPage>
--- a/src/Tests/UITests/Toolkit.UITests.Maui.App/TestPages/BookmarksView/BookmarksTemplateMap.xaml
+++ b/src/Tests/UITests/Toolkit.UITests.Maui.App/TestPages/BookmarksView/BookmarksTemplateMap.xaml
@@ -32,7 +32,7 @@
                 <toolkit:BookmarksView x:Name="BookmarksView" Grid.Row="1">
                     <toolkit:BookmarksView.ItemTemplate>
                         <DataTemplate x:DataType="mapping:Bookmark">
-                            <Border Background="Gold" Padding="4" Margin="2">
+                            <Border Padding="4" Margin="2">
                                 <Label Text="{Binding Name, StringFormat='Custom: {0}'}" />
                             </Border>
                         </DataTemplate>
--- a/src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
+++ b/src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
@@ -1,0 +1,70 @@
+using ImageMagick;
+using OpenQA.Selenium.Appium;
+
+namespace Toolkit.UITest.Shared.BookmarksView;
+
+[TestClass]
+public class BookmarksViewTests : AppiumTestBase
+{
+    private const string BookmarksMapPage = "BookmarksMap";
+
+    [TestMethod]
+    public void BookmarksView_Map()
+    {
+        OpenSample(BookmarksMapPage);
+
+        var bookmarksList = FindElement("BookmarksListView", TimeSpan.FromSeconds(5));
+#if WPF_TEST || WINUI_TEST
+        Assert.AreEqual("Bookmarks", GetAutomationName(bookmarksList), "The bookmarks list should expose its accessible name.");
+#endif
+
+        var bookmarks = new (string name, string expectedCenterText)[]
+        {
+            ("Red bookmark", "10.0,10.0"),
+            ("Green bookmark", "30.0,20.0"),
+            ("Blue bookmark", "-30.0,-20.0"),
+        };
+        var map = FindElement("BookmarksMap", TimeSpan.FromSeconds(5));
+        AssertNoBookmarkMarkersAreVisible(map, bookmarks);
+
+        foreach (var (name, expectedCenterText) in bookmarks)
+        {
+            FindElementByName(name, TimeSpan.FromSeconds(5)).Click();
+            FindElementByText(expectedCenterText, TimeSpan.FromSeconds(5));
+            var selectedBookmarkText = FindElement("SelectedBookmarkText", TimeSpan.FromSeconds(5));
+            Assert.AreEqual(name, GetLabelText(selectedBookmarkText), "The BookmarkSelected event should identify the selected bookmark.");
+            AssertSelectedMarkerIsVisible(map, name);
+        }
+    }
+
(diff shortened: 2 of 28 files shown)


gdestigter:
@imalcolm1 / @SamhithaBurugupalli, apologies for the long PR. You should be able to review commit by commit.

Please check the accessibility / automation updates to make sure they're accurate and match the design.

Also, I'm still getting acquainted with the UI tests in this repo, so give feedback on the general test approach (too many test pages, too many tests, too few tests, etc.). Happy to take a different path based on your feedback.

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
```diff
@@ -36,15 +36,16 @@ static BookmarksView()
...
               </DataTemplate>";
         DefaultDataTemplate = Microsoft.Maui.Controls.Xaml.Extensions.LoadFromXaml(new DataTemplate(), dataTemplate);
 
+        // A semantic description makes the CollectionView an accessibility leaf on iOS, hiding its bookmark items.
         string template =
             $@"<ControlTemplate xmlns=""http://schemas.microsoft.com/dotnet/2021/maui""
                                 xmlns:x=""http://schemas.microsoft.com/winfx/2009/xaml""
                                 xmlns:esriTK=""clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui"">
-                <CollectionView x:Name=""{_presentingViewName}"" SelectionMode=""Single"" ItemTemplate=""{{TemplateBinding ItemTemplate}}"" />
+                 <CollectionView x:Name=""{_presentingViewName}"" SelectionMode=""Single"" ItemTemplate=""{{TemplateBinding ItemTemplate}}"" AutomationProperties.AutomationId=""BookmarksListView"" />
```
imalcolm1: I notice unlike WPF/WinUI, MAUI adds a description property to the individual entries but not the CollectionView as a whole. Did you find it just wasn't possible to set the description on the collection?

src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
```diff
@@ -39,6 +40,27 @@ public void BookmarksView_Map()
...
+    public void BookmarksView_Style()
+    {
+        OpenSample(BookmarksTemplateMapPage);
+
+        foreach (var bookmarkName in new[] { "Red bookmark", "Green bookmark", "Blue bookmark" })
+        {
+            FindElementByText($"Custom: {bookmarkName}", TimeSpan.FromSeconds(5));
+#if WPF_TEST || WINUI_TEST
+            FindElementByName(bookmarkName, TimeSpan.FromSeconds(5));
+#endif
```
imalcolm1: I think this needs a comment about why we're only doing this on WPF and WinUI (ie because they use `PrepareContainerForItemOverride` to set the description property always, rather than it being configured on the default style).

src/Tests/UITests/Toolkit.UITests.Maui.App/TestPages/BookmarksView/BookmarksTemplateMap.xaml
```diff
@@ -0,0 +1,44 @@
...
+                Stroke="{StaticResource CalciteBorder1Brush}"
+                StrokeThickness="1"
+                VerticalOptions="End"
+                WidthRequest="300">
+            <Grid RowDefinitions="Auto,*" RowSpacing="4">
+                <Button AutomationId="SetRuntimeTemplateButton" Clicked="SetRuntimeTemplateButton_Click" Text="Apply runtime" />
+                <toolkit:BookmarksView x:Name="BookmarksView" Grid.Row="1">
+                    <toolkit:BookmarksView.ItemTemplate>
+                        <DataTemplate x:DataType="mapping:Bookmark">
+                            <Border Background="Gold" Padding="4" Margin="2">
```
imalcolm1: minor: The changed background color implies that we're checking it in the test, which might cause some confusion given the test only actually checks the text

src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
```diff
@@ -0,0 +1,193 @@
...
+
+    [TestMethod]
+    public void BookmarksView_Map()
+    {
+        OpenSample(BookmarksMapPage);
+
+        var bookmarksList = FindElement("BookmarksListView", TimeSpan.FromSeconds(5));
+#if WPF_TEST || WINUI_TEST
+        Assert.AreEqual("Bookmarks", GetAutomationName(bookmarksList), "The bookmarks list should expose its accessible name.");
+#endif
```
imalcolm1, line 25: Following up on the Bookmarks name in Maui vs WinUI/WPF: if it isn't possible to get that working, there should probably be a comment to explain why here

src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        var bookmarks = new (string name, string expectedCenterText)[]
+        {
+            ("Red bookmark", "10.0,10.0"),
+            ("Green bookmark", "30.0,20.0"),
+            ("Blue bookmark", "-30.0,-20.0"),
+        };
+        var map = FindElement("BookmarksMap", TimeSpan.FromSeconds(5));
```
imalcolm1: It looks like on Mac the default appium function to get by automation Id is also checking for the value property, which means it picks up the search bar text which has the same name as the map automation id, so the color check screenshot is wrong and the test fails. I think I have a fix for the FindElement function, so I'll put up a PR for that separately

imalcolm1 (commented):
Looks excellent, thanks so much for doing this! I have a question about a difference between Maui vs WinUI/WPF automation names, but otherwise looks pretty much perfect

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
SamhithaBurugupalli: I was going to put this in my review, the approach used for WPF/WinUI is a solid approach. We can put the same in xaml setter as well. But when there is an override of the template, the automation name would not be preserved if the user has not set automation name explicitly as well. So, the screen reader would read out the entire type (something like Esri.arcgisruntime.toolkit......)
I would recommend  similar approach for Maui as well. 
We could something like:
  ```
public static readonly BindableProperty ItemTemplateProperty =
        BindableProperty.Create(nameof(ItemTemplate), typeof(DataTemplate), typeof(BookmarksView), DefaultDataTemplate, BindingMode.OneWay, null, propertyChanged: ItemTemplateChanged);

    private static void ItemTemplateChanged(BindableObject sender, object? oldValue, object? newValue)
    {
        BookmarksView bookmarksView = (BookmarksView)sender;
        if (bookmarksView._presentingView is not null)
        {
            bookmarksView._presentingView.ItemTemplate = CreateAccessibleItemTemplate(newValue as DataTemplate);
        }
    }

    private static DataTemplate? CreateAccessibleItemTemplate(DataTemplate? itemTemplate)
    {
        if (itemTemplate is null)
        {
            return null;
        }

        return new DataTemplate(() =>
        {
            ContentView container = new ContentView
            {
                Content = itemTemplate.CreateContent() as View
                    ?? throw new InvalidOperationException("The bookmark item template must create a View."),
            };
            container.SetBinding(SemanticProperties.DescriptionProperty, nameof(Bookmark.Name));
            return container;
        });
    }
```
    And set `_presentingView.ItemTemplate = CreateAccessibleItemTemplate(ItemTemplate);` in `OnApplyTemplate`

File: src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs [hunk 6]
```diff
@@ -0,0 +1,193 @@
...
+    private const string BookmarksScenePage = "BookmarksScene";
+    private const string BookmarksTemplateMapPage = "BookmarksTemplateMap";
+    private const string BookmarksUpdatesPage = "BookmarksUpdates";
+
+    [TestMethod]
+    public void BookmarksView_Map()
+    {
+        OpenSample(BookmarksMapPage);
+
+        var bookmarksList = FindElement("BookmarksListView", TimeSpan.FromSeconds(5));
```
SamhithaBurugupalli: Can TimeSpan.FromSeconds(5) be parameterized, like timeout = TimeSpan.FromSeconds(5)

SamhithaBurugupalli (commented):
Looks good overall! This PR is covering most of the accessibility requirements along with tests for this control. Just a few points worth fixing, either in this PR or a separate one for a complete a11y compliance. 
- As pointed by Ian earlier, I think we should set the Automation Name for Maui in a similar way done for WPF and WinUI, so that it will not be overridden when a user customizes the template.
- When in high contrast mode for WPF, the text is not visible properly.
This screenshot is for "Aquatic theme"
Image

src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
imalcolm1: Merged Friday: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/836

src/Toolkit/Toolkit.Maui/BookmarksView/BookmarksView.cs
gdestigter: Implemented in ede3e996. MAUI now wraps every item template and binds the wrapper semantic description to Bookmark.Name, so default, custom, and runtime templates preserve the bookmark name. The CollectionView itself remains unnamed because a list-level semantic description makes it an accessibility leaf on iOS and hides its bookmark children.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
gdestigter: Implemented in ede3e996. The platform conditional was removed from the custom-template item-name assertion because MAUI now preserves those names too. The remaining Windows-only list-name assertion has an explicit comment explaining the iOS accessibility-leaf behavior.

src/Tests/UITests/Toolkit.UITests.Maui.App/TestPages/BookmarksView/BookmarksTemplateMap.xaml
gdestigter: Removed the Gold background from the custom template on MAUI, WPF, and WinUI in ede3e996, so the page now signals only the text/template behavior that the test verifies.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
gdestigter: Added the explanation here in ede3e996. MAUI intentionally has no list-level accessible name because that hides the bookmark children on iOS; item names are now preserved outside the consumer template by the accessible wrapper and tested on MAUI as well.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/BookmarksView/BookmarksViewTests.cs
gdestigter: Implemented in ede3e996 as a shared DefaultTimeout and replaced the repeated positive five-second waits. It is set to ten seconds after Windows runs showed Scene and observable-collection transitions occasionally exceeding five seconds; the one-second negative checks and 30-second online-map waits remain intentional.

gdestigter:
Addressed the review feedback in `ede3e996`:

- MAUI now wraps every item template and binds the wrapper semantic description to `Bookmark.Name`, preserving accessibility names for default, custom, and runtime templates.
- The shared positive wait is parameterized as `DefaultTimeout`; short negative checks and online-map waits remain distinct.
- Removed the decorative Gold custom-template background on all platforms.
- WPF test panels now use system window/window-text brushes for high-contrast compatibility.
- Rebased onto current `v.next`, including #836.

imalcolm1 (approved):
Looks good!

SamhithaBurugupalli (approved):

PR #836, opened 2026-08-14
UITests: Make FindElement more specific on mac

imalcolm1:
The default Appium locator strategies for Id all seem to search both the `identifier` and `value` properties. `value` is also linked to text, for example the the text value of an Entry field, which can mean the wrong element gets identified.

Theoretically XPath is slower than the locator, I didn't notice a difference, and appium mac is very slow anyway.


gdestigter (approved):
👍

Issue #732, opened 2026-03-23
UtilityNetworkTraceTool Doesn't Load when Adding Utility Network As Layer, Not via Webmap


AmrEldib:
We have a client who's working with the UtilityNetworkTraceTool control. The [control](https://esri.github.io/arcgis-maps-sdk-dotnet-toolkit/un-trace.html) supports loading networks and named trace configurations from a web map.
However, when trying to [add the utility network as a layer](https://developers.arcgis.com/net/utility-network/display-a-utility-network/), the control doesn't fully load.

Does the control support this approach? What methods/properties need to be called/set to make this happen?
Or, is there a bug in this behavior?

The control is initially not loaded.
<img width="1427" height="700" alt="Image" src="https://github.com/user-attachments/assets/bc4b5b3b-561a-4a80-b7e3-659c90a64c5d" />

Then, the user clicks "Add data" to load the layer, which is loaded (as well as the utility network).
The control reacts (a loading symbol briefly appears, then the control is back). 
<img width="1427" height="699" alt="Image" src="https://github.com/user-attachments/assets/67844ee2-61cc-444a-a3a3-b3927ce93adc" />

However, the user needs to switch tabs to see that the control has loaded.
Then, you can see that no Trace Types were loaded.

<img width="1424" height="698" alt="Image" src="https://github.com/user-attachments/assets/ede543a9-5025-4b12-80d7-6a8b391c2f4e" />

Here's the code (XAML file renamed to txt to allow uploading).
Use the [dotnet-samples](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples). Under `src\MAUI\Maui.Samples\Samples`, you can add a new folder (named "UtilityNetworkTrace" for example), then place the files there.

[UtilityNetworkTraceToolSample.xaml.txt](https://github.com/user-attachments/files/26192956/UtilityNetworkTraceToolSample.xaml.txt)
```xaml
﻿<?xml version="1.0" encoding="utf-8" ?>
<ContentPage
    x:Class="Toolkit.SampleApp.Maui.Samples.UtilityNetworkTraceToolSample"
    xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
    xmlns:esriTK="clr-namespace:Esri.ArcGISRuntime.Toolkit.Maui;assembly=Esri.ArcGISRuntime.Toolkit.Maui"
[...]

PR #844, opened 2026-08-21
utilnet: add support for barriers and filtering named traces

jnery:
Also addresses enabling non-portal map to use the trace tool #732 with `UtilityNetworkTraceTool.loadAsync(Array<String>?): Future<void>`

This might be easier to review per commit

- Extracted `StartingPointModel` shared logic into `UtilityElementModel` (shared base)
- Added `BarrierModel` that inherits from this same `UtilityElementModel` base class. `UseAsFilterBarrier` is unique to `BarrierModel` which is used to determine `Barriers` or `FilterBarriers` collection.
- Controller (shared logic): Add barrier symbol, methods, graphics when collection changes (identical code to starting points)
- Windows: Add to resources, expose properties and methods
- WPF: UI in theme
- WinUI: UI in theme
- Add `LoadAsync` to filter by named traces
- MAUI: UI in code, fixed bugs in element view (missing terminal picker and fraction along slider, symbol not always visible, ability to zoom and remove all)

### DIFF
--- a/docs/un-trace.md
+++ b/docs/un-trace.md
@@ -7,6 +7,8 @@ Use named trace configurations defined in a web map to perform connected trace o
 ## Features
 
 - Load networks and named trace configurations from a web map
+- Load selected named trace configurations by name with `LoadNamedTracesAsync`
+- Add barriers interactively or programmatically with `AddBarrier`; barriers can also be configured as filter barriers
 - Support for templating
 - (WinUI, WPF) Identify starting point candidates, then use the inspection view to narrow the selection:
 
@@ -28,6 +30,7 @@ The following properties enable customization (WinUI, WPF only):
 
 - `ResultItemTemplate` - override the display of results
 - `StartingPointItemTemplate` - override the display of starting points
+- `BarrierItemTemplate` - override the display of barriers
 - `TraceTypeItemTemplate` - override the display of the trace configuration choices
 - `UtilityNetworkItemTemplate` - override the display of the Utility Network choices
 
@@ -35,9 +38,10 @@ All platforms:
 
 - `Template` - allows overriding the appearance of the entire control
 
-The following properties enable customizing symbology:
+The following properties enable customizing symbology on all platforms:
 
 - `StartingPointSymbol`
+- `BarrierSymbol`
 - `ResultFillSymbol`
 - `ResultLineSymbol`
 - `ResultPointSymbol`
--- a/src/Toolkit/Toolkit.Maui/UtilityNetworkTraceTool/UtilityNetworkTraceTool.Appearance.cs
+++ b/src/Toolkit/Toolkit.Maui/UtilityNetworkTraceTool/UtilityNetworkTraceTool.Appearance.cs
@@ -1,4 +1,4 @@
-﻿// /*******************************************************************************
+// /*******************************************************************************
 //  * Copyright 2012-2018 Esri
 //  *
 //  *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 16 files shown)


SamhithaBurugupalli (commented):
Looks good to me overall! Just a few points.

1. Observed a minor nit, the message which shows - "trying to find starting points" should be something like "trying to find barrier" right when trying to add a barrier?. Attached the video for ref

https://github.com/user-attachments/assets/ee4b7afb-3b33-400e-901f-2ff4d8985e4d

2. Thanks for correcting the close and next buttons. Also, I observed that the only way to dismiss the callout is going to the inspect (and noticed this inspect is missing on Maui even for starting point) and then clicking the close button. was this intentional? Can we dismiss the callout when clicked on the part of map with no data? 

3. Happy to see that AutomationName was set on one of the buttons in WinUI, if you are interested to add more, I would suggest, the slider and the edit box as well. Anyways this could wait and we can do it with a full accessibility check later.

SamhithaBurugupalli (approved):
LGTM!!

src/Toolkit/Toolkit/UI/Controls/UtilityNetworkTraceTool/StartingPointModel.cs
imalcolm1: I noticed the BarrierModel equality takes into account GlobalId. Should it be the same for StartingPointModel? If so maybe it could just go on the base class

src/Toolkit/Toolkit.Maui/UtilityNetworkTraceTool/UtilityNetworkTraceTool.Appearance.cs
imalcolm1, line 1: minor: The close button is in a bit of a strange place for the add barrier dialog, I would have expected it to be on the top like it is when adding a starting point. May or may not be worth changing

Image

docs/un-trace.md
imalcolm1, line 1: Not sure what file to put this on since it's all 3 platforms, but when the callout doesn't switch between selected starting and barrier points if they are already selected in their respective lists. Maybe selection should be cleared in the other list when an item from one list is selected?

https://github.com/user-attachments/assets/76858c91-b599-49ed-bdb5-52273f08ea31

src/Toolkit/Toolkit/UI/Controls/UtilityNetworkTraceTool/UtilityNetworkTraceToolController.cs
```diff
@@ -554,8 +672,140 @@ public Tuple<CalloutDefinition, MapPoint>? RequestedCallout
...
+        public void DismissCallout()
+        {
+            _calloutRequestVersion++;
+            _requestedCallout = null;
+            OnPropertyChanged(nameof(RequestedCallout));
+        }
         #endregion Starting Points
 
+        #region Barriers
+        public void AddBarrier(ArcGISFeature feature, MapPoint? location = null, bool? useAsFilterBarrier = false)
```
imalcolm1, line 693: Not sure this is worthwhile, but there's enough code shared between this and AddStartingPoint that we might consider consolidating them with an inner function that also takes the respective graphicsoverlay and observable collection?

imalcolm1 (approved):
Looks good to me, just some minor things. Also looks good on Mac, iOS, Android as far as I can tell!

src/Toolkit/Toolkit/UI/Controls/UtilityNetworkTraceTool/UtilityNetworkTraceToolController.cs
jnery: Certainly. I had this implementation originally, but I chose the more conservative approach because it was easier to review and less likely to introduce regressions. That said, this is definitely a good area for future cleanup and simplification.

src/Toolkit/Toolkit.Maui/UtilityNetworkTraceTool/UtilityNetworkTraceTool.Appearance.cs
jnery: I believe this was centered on all platforms to support canceling an identify for both starting points and barriers.

PR #846, opened 2026-08-21
Bookmarks test fix

SamhithaBurugupalli:
This PR fixes the BookmarksView_style UI test failure on iOS and Catalyst and aligns automation-name behavior across all supported platforms.

The failure occurred when a custom bookmark template was applied. The template rendered the bookmark text using a custom format, such as Custom: Red bookmark, but the generated list item still received its automation name from the raw bookmark name, Red bookmark. On iOS and Catalyst, Appium uses the MAUI semantic description when locating the element, so the mismatch was observable and caused the test to fail.

The same underlying issue existed on WPF and WinUI, but it was less visible there. On those platforms, the top-level list item exposed the raw name, Red bookmark, while the text rendered beneath it displayed the custom value, Custom: Red bookmark. Because both values were present in the automation tree, the discrepancy did not cause the same visible test failure.

So, changed the previously suggested implementation (to override the template). If a user now tries to customize the template, then they should be setting the Automation name as well.


gdestigter (approved):
:+1:

imalcolm1 (approved):
This seems like a good approach, it's good for users to have more customizability

PR #849, opened 2026-08-25
Improves WPF SearchView keyboard and screen-reader accessibility

SamhithaBurugupalli:
Creating small feature/ platform wise PRs so that it is easier to review.

- Enables complete Tab/Shift+Tab navigation across source, suggestion, and result popups.
- Moves focus to results after keyboard suggestion selection and completed repeat searches.
- Corrects Narrator announcements for source selector, All sources, and selected state.

### DIFF
--- a/src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
+++ b/src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
@@ -164,6 +164,7 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
         private Button? _searchButton;
         private IInputElement? _focusTargetAfterSuggestions;
         private bool _focusResultsWhenAvailable;
+        private bool _sourceSelectionByKeyboard;
 
         /// <inheritdoc/>
         public override void OnApplyTemplate()
@@ -182,11 +183,15 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             if (_allSourcesButton != null)
             {
                 _allSourcesButton.PreviewKeyDown -= AllSourcesButton_PreviewKeyDown;
+                _allSourcesButton.Click -= AllSourcesButton_Click;
+                _allSourcesButton.PreviewMouseDown -= SourceSelection_PreviewMouseDown;
             }
 
             if (_sourceList != null)
             {
                 _sourceList.PreviewKeyDown -= SourceList_PreviewKeyDown;
+                _sourceList.SelectionChanged -= SourceList_SelectionChanged;
+                _sourceList.PreviewMouseDown -= SourceSelection_PreviewMouseDown;
             }
 
             if (_searchButton != null)
@@ -224,11 +229,15 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             if (_allSourcesButton != null)
             {
                 _allSourcesButton.PreviewKeyDown += AllSourcesButton_PreviewKeyDown;
+                _allSourcesButton.Click += AllSourcesButton_Click;
+                _allSourcesButton.PreviewMouseDown += SourceSelection_PreviewMouseDown;
             }
 
             if (_sourceList != null)
             {
                 _sourceList.PreviewKeyDown += SourceList_PreviewKeyDown;
+                _sourceList.SelectionChanged += SourceList_SelectionChanged;
+                _sourceList.PreviewMouseDown += SourceSelection_PreviewMouseDown;
             }
 
             if (_searchButton != null)
@@ -257,6 +266,11 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
 
         private void AllSourcesButton_PreviewKeyDown(object sender, KeyEventArgs e)
         {
+            if (e.Key == Key.Enter || e.Key == Key.Space)
+            {
+                _sourceSelectionByKeyboard = true;
+            }
+
             // Enter the list by focusing the first item container so keyboard navigation starts on an actual option.
             if (e.Key == Key.Tab && (Keyboard.Modifiers & ModifierKeys.Shift) == 0 && FocusListItem(_sourceList, 0))
             {
@@ -264,22 +278,51 @@ namespace Esri.ArcGISRuntime.Toolkit.UI.Controls
             }
         }
 
-        private void SourceList_PreviewKeyDown(object sender, KeyEventArgs e)
+        private void AllSourcesButton_Click(object sender, RoutedEventArgs e)
         {
-            if (e.Key != Key.Tab)
+            if (!_sourceSelectionByKeyboard)
             {
                 return;
             }
 
-            // Keep Tab traversal inside the popup region: Shift+Tab returns to All sources; Tab exits to query.
-            var focused = (Keyboard.Modifiers & ModifierKeys.Shift) != 0
-                ? _allSourcesButton?.Focus() == true
-                : CloseSourcePopupAndFocusQuery();
+            _sourceSelectionByKeyboard = false;
+            _ = Dispatcher.BeginInvoke(new Action(() => _queryEntry?.Focus()), System.Windows.Threading.DispatcherPriority.Loaded);
+        }
 
-            if (focused)
+        private void SourceList_PreviewKeyDown(object sender, KeyEventArgs e)
+        {
+            if (e.Key == Key.Tab)
             {
-                e.Handled = true;
+                // Keep Tab traversal inside the popup region: Shift+Tab returns to All sources; Tab exits to query.
+                var focused = (Keyboard.Modifiers & ModifierKeys.Shift) != 0
+                    ? _allSourcesButton?.Focus() == true
+                    : CloseSourcePopupAndFocusQuery();
+
+                if (focused)
+                {
+                    e.Handled = true;
+                }
+
+                return;
             }
+
+            _sourceSelectionByKeyboard = true;
+        }
+
+        private void SourceList_SelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            if (e.AddedItems.Count == 0 || !_sourceSelectionByKeyboard)
+            {
+                return;
+            }
+
+            _sourceSelectionByKeyboard = false;
+            _ = Dispatcher.BeginInvoke(new Action(() => _queryEntry?.Focus()), System.Windows.Threading.DispatcherPriority.Loaded);
+        }
+
+        private void SourceSelection_PreviewMouseDown(object sender, MouseButtonEventArgs e)
+        {
+            _sourceSelectionByKeyboard = false;
         }
 
         private bool CloseSourcePopupAndFocusQuery()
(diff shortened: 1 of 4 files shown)


src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -139,6 +152,243 @@ private void SuggestionList_SelectionChanged(object? sender, SelectionChangedEve
...
+            {
+                return;
+            }
+
+            _focusResultsWhenAvailable = false;
+            // Defer until layout completes so the first result container exists and can receive focus.
+            _ = Dispatcher.BeginInvoke(new Action(() => FocusListItem(_resultList, 0)), System.Windows.Threading.DispatcherPriority.Loaded);
+        }
+
+        {
```
gdestigter, line 405: Something happened here, we seem to be missing the method signature. `RepeatSearchAndFocusResults` perhaps?

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -139,6 +152,243 @@ private void SuggestionList_SelectionChanged(object? sender, SelectionChangedEve
...
+                ReferenceEquals(e.OldFocus, _focusTargetAfterSuggestions) &&
+                FocusListItem(_suggestionList, 0))
+            {
+                e.Handled = true;
+            }
+        }
+
+        private void SuggestionList_PreviewKeyDown(object sender, KeyEventArgs e)
+        {
+            if (e.Key == Key.Enter || e.Key == Key.Space)
```
gdestigter, line 356: Should this be accompanied by setting the `AcceptsReturn` flag in the WPF suggestion list? Or does this work differently?

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -139,6 +152,243 @@ private void SuggestionList_SelectionChanged(object? sender, SelectionChangedEve
...
+            var item = listView.Items[index];
+            listView.ScrollIntoView(item);
+            // Realize virtualized item containers before focusing, otherwise ContainerFromItem can return null.
+            listView.UpdateLayout();
+            return (listView.ItemContainerGenerator.ContainerFromItem(item) as ListViewItem)?.Focus() == true;
+        }
+
+        private void FocusResultsWhenAvailable()
+        {
+            if (!_focusResultsWhenAvailable || SearchViewModel?.Results?.Count < 1)
```
gdestigter: Do we need to clear `_focusResultsWhenAvailable` when the result count is 0?

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Oops missed to push a local commit. Thanks for catching that

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: I think the `AcceptsReturn` is only available for `TextBox` ([ref](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.textbox.acceptsreturn?view=windowsdesktop-10.0)) and it is serving a different purpose which is to create a new line. Is this the one which you are referring to?

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Updated

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
gdestigter: I was thinking: `KeyboardNavigation.AcceptsReturn`, but if what you have works and selects the `Item`, then I'm good with that.

gdestigter (approved):
Looks good :+1:

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -139,6 +152,249 @@ private void SuggestionList_SelectionChanged(object? sender, SelectionChangedEve
...
+            {
+                return;
+            }
+
+            e.Handled = (Keyboard.Modifiers & ModifierKeys.Shift) != 0
+                ? _searchButton?.Focus() == true
+                : MoveFocusPastSearchView();
+        }
+
+        private static bool FocusListItem(ListView? listView, int index)
```
imalcolm1, line 373: This function seems to trigger item selected which immediately navigates to the location and collapses the suggestions list. Hitting enter on the search button immediately searches the first result, and you can't use the arrow keys to navigate the search results before selecting one

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
imalcolm1: On my machine return does not select, but space does, so I think Greg may be right

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -139,6 +152,249 @@ private void SuggestionList_SelectionChanged(object? sender, SelectionChangedEve
...
+            if (focused)
+            {
+                e.Handled = true;
+            }
+        }
+
+        private bool CloseSourcePopupAndFocusQuery()
+        {
+            IsSourceSelectOpen = false;
+            return _queryEntry?.Focus() == true;
```
imalcolm1, line 331: I'm not sure why but this doesn't seem to be working. It seems to reset my tab navigation to the default root for the page

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -139,6 +152,249 @@ private void SuggestionList_SelectionChanged(object? sender, SelectionChangedEve
...
+        {
+            if (e.Key != Key.Tab)
+            {
+                return;
+            }
+
+            // Keep Tab traversal inside the popup region: Shift+Tab returns to All sources; Tab exits to query.
+            var focused = (Keyboard.Modifiers & ModifierKeys.Shift) != 0
+                ? _allSourcesButton?.Focus() == true
+                : CloseSourcePopupAndFocusQuery();
```
imalcolm1: This only applies when not on the AllSources button, which means I tab down to the first individual search source but none of the others which could be a bit confusing since to an end user they all look like they're part of the same list

imalcolm1 (commented):
Noticed a couple of small things. Also, this didn't fit anywhere specifically, but I get slightly strange behavior when hitting enter on the clear button after doing a search. It focuses what might be the collapsed button or the collapsed search list and when I shift tab again, it takes me to the Esri logo (though that last part might be intended)

https://github.com/user-attachments/assets/9bc37312-5f12-4b93-a541-361e3aa81008

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
imalcolm1: Never mind, I keep forgetting WPF does CTRL+Arrows to navigate these

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
imalcolm1: I can still ctrl+arrows so this probably isn't necessary

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -139,6 +152,249 @@ private void SuggestionList_SelectionChanged(object? sender, SelectionChangedEve
...
+            {
+                e.Handled = true;
+            }
+        }
+
+        private void SearchButton_PreviewGotKeyboardFocus(object sender, KeyboardFocusChangedEventArgs e)
+        {
+            // If Shift+Tab comes from the element after SearchView, re-enter suggestions instead of skipping the popup.
+            if ((Keyboard.Modifiers & ModifierKeys.Shift) != 0 &&
+                ReferenceEquals(e.OldFocus, _focusTargetAfterSuggestions) &&
```
OAzmitia, line 347: Does _focusTargetAfterSuggestion ever clear. What if the user continues using the app clicking around and then uses tab/shift+tab?

SamhithaBurugupalli: I am not clearing `_focusTargetAfterSuggestions` intentionally when the user clicks elsewhere. It remembers which control comes after the suggestions in the tab order.

If the user later focuses on that same control (which was after the suggestions list) and presses Shift+Tab, focus returns to the suggestions. This is needed because the suggestions are in a separate WPF popup, so normal tab navigation cannot reliably return to them. If the suggestions are no longer available, normal tab navigation is used.

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Addressed in commit - https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/849/commits/e09f75d6c02a0d618eaeff501fb866329ddbefee

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Thanks for catching that!!
This method is being used when I navigate using tab (without changing the selection). 
So, when a user sets the source (ex: to either All Sources/ World Geocoder/ etc) then isn't being called. Addressed this issue in commit -  https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/849/commits/56e0f9a354e485cb29f1496056a272f6eb317122

SamhithaBurugupalli:
> Noticed a couple of small things. Also, this didn't fit anywhere specifically, but I get slightly strange behavior when hitting enter on the clear button after doing a search. It focuses what might be the collapsed button or the collapsed search list and when I shift tab again, it takes me to the Esri logo (though that last part might be intended)
> 
>  20260825-1958-20.3819755.mp4

I tried and dug a bit more into this issue. On WPF, when an element which was focused by keyboard disappears (by some action, in this case clear button was hidden by clicking on it), the focus redirects to the parent element. If the parent is not focusable it will redirect to it's parent/ application window. So, here it is going to the SearchView control itself. I tried setting the `Focusable` to `false` but that is redirecting the focus to app window (which is way worse than what we have now). So, it should be handled again in the events (which is doable and it works), but I feel we could leave it as is. 

Also, I have set `IsTabStop` property on the searchView control to `False` in commit - https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/849/commits/68c6d003c279197927d92ce37741eb9b433de599. This would avoid an extra tabstop while navigating normally. (But the focus lands on the control when the clear search button is clicked)

imalcolm1 (approved):
Looks great, thank you!

PR #850, opened 2026-09-01
Improves search view keyboard navigation in WinUI

SamhithaBurugupalli:
1. Enables complete Tab/ Shift+Tab/ arrow keys navigation across source, suggestion, and result popups.
2. Moves focus to results after keyboard suggestion selection and completed repeat searches.
3. Corrects Narrator announcements for source selector, All sources, and selected state.
4. Did some code refactoring extracting the common code of WPF and WinUI


src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
imalcolm1, line 1: optional: The WinUI and WPF ifdefs for `OnApplyTemplate()` + event listeners are both long enough it might be a good idea to consider moving those sections to separate SearchView.WinUI.cs and SearchView.WPF.cs files.

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchSourceButtonAutomation.cs
```diff
@@ -76,16 +98,13 @@ private string GetAccessibleName(bool isSelected)
...
 
-            // Appending "selected" to the Name property produces consistent speech across screen readers.
             var format = Properties.Resources.GetString("SearchViewSelectedAutomationName") ?? "{0}, selected";
             return string.Format(CultureInfo.CurrentCulture, format, name);
         }
     }
 
-    // The source button opens/closes popup content, so automation should expose Expand/Collapse semantics.
-    public sealed class SourceSelectToggleButton : ToggleButton
+    public sealed partial class SourceSelectToggleButton : ToggleButton
```
imalcolm1, line 106: Not sure this is a real problem, but I realized I can't shift+tab out of the sources suggestion list. It's not a full trap: I can still tab out of it to collapse the list then shift+tab back

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -102,35 +124,321 @@ private void InitializeLocalizedStrings()
...
+            if (e.Key == VirtualKey.Tab)
+            {
+                var focused = IsShiftPressed()
+                    ? _allSourcesButton?.Focus(FocusState.Keyboard) == true
+                    : CloseSourcePopupAndFocusQuery();
+                e.Handled = focused;
+            }
+            else
+            {
+                _sourceSelectionByKeyboard = true;
```
imalcolm1, line 301: This line never seems to get hit when selecting using keyboard (only the SelectionChanged event handler below gets called), so selecting a source results in highlighting the drop down button rather than the query entry

imalcolm1 (commented):
Looking good mostly!

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -102,35 +124,321 @@ private void InitializeLocalizedStrings()
...
+        }
+
+        private void SourcePopup_Opened(object? sender, object e)
+        {
+            _ = DispatcherQueue.TryEnqueue(() => _allSourcesButton?.Focus(FocusState.Keyboard));
+        }
+
+        private void AllSourcesButton_KeyDown(object sender, KeyRoutedEventArgs e)
+        {
+            if (e.Key == VirtualKey.Tab && !IsShiftPressed() && FocusFirstVisibleItem(_sourceList))
```
gdestigter: Should we try to handle Shift+Tab as well?

gdestigter (commented):
I noticed the same Shift-Tab behavior that Ian saw. Otherwise, this looks good to me.

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchSourceButtonAutomation.cs
SamhithaBurugupalli: Thanks for catching that. I added that part now. Also, on WinUI when we are on search sources list, if we press "Esc" it would close the sources. For some reason, I thought it was the same case for WPF, Apparently not. so, adding this feature as well to just be consistent. 
addressed in commit: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/850/commits/cbd3657c607bcf54144da4a9cd88f5075cbe2b7c

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Thanks for catching that. I added that part now. Also, on WinUI when we are on search sources list, if we press "Esc" it would close the sources. For some reason, I thought it was the same case for WPF, Apparently not. so, adding this feature as well to just be consistent.
addressed in commit: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/commit/cbd3657c607bcf54144da4a9cd88f5075cbe2b7c

gdestigter (approved):

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Addressed in commit: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/850/commits/f074ce5b5e8c1906b11ebb1e731199083558baac
Also noted another minor issue in WinUI where the All sources button is receiving the keyboard focus even when a mouse is used to click on it. So fixed that as well in commit: https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/850/commits/cdf974eab91aae35f40c75bf8c435a9da2c009ed

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Sure, I will add this clean up in a separate PR.

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -178,6 +178,7 @@ protected override void OnApplyTemplate()
 
             GetCommonTemplateParts();
             _ungroupedSuggestionList = GetTemplateChild("PART_SuggestionListUnGrouped") as ListView;
+            _sourceSelectToggle = GetTemplateChild("PART_SourceSelectToggle") as ToggleButton;
```
imalcolm1: minor: This should probably be part of `GetCommonTemplateParts()` now

imalcolm1 (approved):
Looks good!

PR #851, opened 2026-09-01
[a11y] [SearchView] The screen reader should announce the No results message when there aren't any search suggestions/ results available

SamhithaBurugupalli:


src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
```diff
@@ -1040,6 +1074,7 @@ private async Task HandleResultsCollectionChanged()
 
             TemplateSettings.OnResultViewVisibilityChanged();
             TemplateSettings.OnResultMessageVisibilityChanged();
+            AnnounceNoResults();
```
gdestigter, line 1136: How is cancellation handled for these? Should we get announcement when a search is cancelled?

imalcolm1 (approved):

gdestigter (approved):

OAzmitia (approved):

src/Toolkit/Toolkit.UI.Controls/SearchView/SearchView.cs
SamhithaBurugupalli: Ideally with screen readers, any changes on the screen without the user interference should be announced. As clear search action would be done by the user, it is not necessary to announce that. But now as I think about it, it would be better to announce that there are suggestions available.

PR #854, opened 2026-09-02
Announce suggestions/ results availability when displayed

SamhithaBurugupalli:
Follow up based on the PR comment [here ](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/851)
- Added logic so that the automation events are raised and the screen reader announces that suggestions/ results are available
- Did not add it for sources list intentionally because the screen would say, expanded/ collapsed state so it would be repetitive
- Noticed that No results is not being announced in WPF because the Textblock is not able raise the right automation event, so changed it to label.

Note: The behavior may slightly vary based on the screen reader which is used. For the video below I am using NVDA. The default windows narrator announces it a bit differently. 

WPF:
 

https://github.com/user-attachments/assets/657b5c3f-6b4b-4f89-b219-1aed2616db4c


OAzmitia (approved):

gdestigter (approved):
LGTM

PR #856, opened 2026-09-09
[a11y] [SearchView] Improves the keyboard and narrator accessibility in Maui WinUI

SamhithaBurugupalli:
I tried to keep the behavior of Maui Winui consistent with WinUI. Will be raising a stacked PR for android

- Enables complete Tab/ Shift+Tab/ arrow keys navigation across source, suggestion, and result popups.
- Moves focus to results after keyboard suggestion selection and completed repeat searches.
- Enables the screen reader to announce the availability of suggestions, results and conveys no results message.


imalcolm1 (approved):
Works like a charm!

OAzmitia (approved):

PR #861, opened 2026-09-10
[a11y] [SearchView] Improves the keyboard and narrator accessibility in Android

SamhithaBurugupalli:
Aimed for similar behavior as winui and maui winui. (skipped minor enhancement changes for android)

Some explicit handling of arrow keys, tab, shift + tab had to be done in android because of default platform behavior which are described as follows

The arrow keys in android navigate to the next element by default. So, when in any of the list view items (ex: search sources/ suggestions/ results) the downward arrow was navigating fine till the last list item, and after that, the keyboard focus is reaching the next focusable items (in the samples page, it was map). This would not be semantically correct. So, had to make changes, to handle this flow. Other issue observed in grouped suggestions list with max suggestions in the 1st group (when 6 list items are present), after reaching the last item of first group, the downward arrow was setting the focus on mapview next (as the rest of the items were not visible at that point). So, had to handle this flow as well. Because of all these challenges, I had the custom flow for tab/ swit+tab/ arrowkeys (specifically for listviews)
Would appreciate any thoughts and feedback on this approach


src/Toolkit/Toolkit.Maui/SearchView/SearchView.Android.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        public void OnChildViewDetachedFromWindow(NativeView view) => _searchView.SetKeyHandler(view, false);
+    }
+
+    private static void FocusListItem(RecyclerView listView, int currentPosition, int direction)
+    {
+        var adapter = listView.GetAdapter();
+        for (var position = currentPosition + direction;
+            position >= 0 && position < (adapter?.ItemCount ?? 0);
+            position += direction)
```
imalcolm1: minor: this for loop always returns immediately, it could probably just be an if statement which would make the recursive setup of the function more immediately obvious

imalcolm1 (commented):
Looks good save for what we talked about with adding a comment for that one check, plus the workaround for the sources not disappearing also being a bug when tapping

src/Toolkit/Toolkit.Maui/SearchView/SearchView.Android.cs
```diff
@@ -198,6 +182,44 @@ focusedItem.BindingAdapterPosition is int position &&
         }
     }
 
+    private void SetItemHandlers(NativeView? itemView, bool attach)
```
imalcolm1, line 185: Not at all a blocker, but if the entry not being selected is affecting all platforms, is there a way we could fix that in a cross-platform maui way? (Potentially in another PR)

imalcolm1 (approved):
Looks good, thanks!

gdestigter (approved):

OAzmitia (approved):

PR #863, opened 2026-09-12
[a11y] [WPF] [WinUI]: Search View text scaling

SamhithaBurugupalli:
According to [WCAG guidelines](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html) any text displayed should be scalable up to 200% without any loss/ truncation of information.

Note: Winui automatically scales up the fontsize even if it is set to a hard coded value like 12px. WPF scales up, if nothing is set for fontsize or if we are using system resources. Explicit scaling is required for WPF, if none of the above scenarios meet.

Observed that some of the listviews in both wpf and winui were not scrollable, so fixed them as well

WPF:

https://github.com/user-attachments/assets/ae8af13d-e21c-46f4-ad5e-5f140b009790

WinUI:

https://github.com/user-attachments/assets/1796a4bf-cdc5-46c1-80d7-5dac2b7ea50e

### DIFF
--- a/src/Toolkit/Toolkit.WPF/Internal/TextScaledFontSizeExtension.cs
+++ b/src/Toolkit/Toolkit.WPF/Internal/TextScaledFontSizeExtension.cs
@@ -1,0 +1,22 @@
+using System;
+using System.Windows.Markup;
+using Windows.UI.ViewManagement;
+
+namespace Esri.ArcGISRuntime.Toolkit.Internal
+{
+    /// <summary>
+    /// Provides a font size adjusted by the Windows text scale factor.
+    /// </summary>
+    [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+    public sealed class TextScaledFontSizeExtension : MarkupExtension
+    {
+        /// <summary>
+        /// Gets or sets the unscaled font size.
+        /// </summary>
+        public double FontSize { get; set; }
+
+        /// <inheritdoc/>
+        public override object ProvideValue(IServiceProvider serviceProvider) =>
+            FontSize * new UISettings().TextScaleFactor;
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Toolkit/Toolkit.WPF/Internal/TextScaledFontSizeExtension.cs
```diff
@@ -0,0 +1,22 @@
...
+    public sealed class TextScaledFontSizeExtension : MarkupExtension
+    {
+        /// <summary>
+        /// Gets or sets the unscaled font size.
+        /// </summary>
+        public double FontSize { get; set; }
+
+        /// <inheritdoc/>
+        public override object ProvideValue(IServiceProvider serviceProvider) =>
+            FontSize * new UISettings().TextScaleFactor;
```
gdestigter, line 20: De we need `UISettings.TextScaleFactorChanged` support for this as well? Or is the markup extension good enough?

imalcolm1: It does seem that the entry bar and the result headings aren't scaling. They need a full app restart to update currently, not just reopening the sample

imalcolm1 (commented):
Looks good except for what Greg already mentioned

src/Toolkit/Toolkit.WPF/Internal/TextScaledFontSizeExtension.cs
SamhithaBurugupalli: Addressed in https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/pull/863/commits/0e75d8ab7f02e64447a63474276a83ded34cb616

gdestigter (approved):

OAzmitia (approved):

PR #864, opened 2026-09-14
a11y - PopupViewer for WPF and WinUI

OAzmitia:
# Summary
Implements the accessibility checkpoints for PopupViewer: UIA name/type/role for the control and its sub-elements, a real Grid/Table pattern for FieldsPopupElement, full-text exposure for TextPopupElement, localized/accessible prev-next media buttons, and removal of several "empty"/decorative nodes cluttering Narrator's navigation. WPF first, then ported to WinUI with the platform-specific workarounds its sealed controls and automation pipeline required.

### DIFF
--- a/src/Tests/UITests/Toolkit.UITests.Shared/Tests/PopupViewer/PopupViewerTests.cs
+++ b/src/Tests/UITests/Toolkit.UITests.Shared/Tests/PopupViewer/PopupViewerTests.cs
@@ -1,0 +1,112 @@
+namespace Toolkit.UITest.Shared.PopupViewer;
+
+[TestClass]
+public class PopupViewerTests : AppiumTestBase
+{
+    private const string PopupViewerFieldsPage = "PopupViewerFields";
+
+    private static readonly TimeSpan DefaultTimeout = TimeSpan.FromSeconds(5);
+
+    // Mirrors the constants in Toolkit.UITests.App.TestPages.PopupViewerFields (the shared test-page
+    // code-behind), which builds the same fully-offline Popup on every platform.
+    private const string PopupTitle = "Ridgeline Trailhead";
+    private const string FieldsElementTitle = "Fields";
+    private const string MediaElementTitle = "Media";
+    private const string TextElementFirstParagraph = "This trailhead connects to the Ridgeline Loop and Summit Spur trails.";
+    private const string TextElementSecondParagraph = "Parking is available year-round, but the upper trail closes seasonally.";
+    private const string ImageMediaAlternativeText = "Illustrated map showing three looping trails around the trailhead";
+    private const string ChartMediaTitle = "Monthly Visitors";
+    private const string ChartMediaCaption = "Visitor counts by month";
+
+    [TestMethod]
+    public async Task PopupViewerText_FullTextIsExposed()
+    {
+        OpenSample(PopupViewerFieldsPage);
+
+        var textArea = FindElement("TextArea", DefaultTimeout);
+        var name = GetAutomationName(textArea);
+        Assert.IsTrue(name.Contains(TextElementFirstParagraph), "Expected the text element's accessible name to include the first paragraph.");
+        Assert.IsTrue(name.Contains(TextElementSecondParagraph), "Expected the text element's accessible name to include the second paragraph, not just the first line.");
+    }
+
+#if WPF_TEST
+    [TestMethod]
+    public async Task PopupViewerFields_ImplementsTableControlPattern()
+    {
+        OpenSample(PopupViewerFieldsPage);
+
+        var fieldsElement = FindElementByName(FieldsElementTitle, DefaultTimeout);
+        var controlType = GetControlType(fieldsElement);
+        var localizedControlType = GetLocalizedControlType(fieldsElement);
+        TestContext.WriteLine($"Fields element ControlType=\"{controlType}\" LocalizedControlType=\"{localizedControlType}\"");
+        Assert.IsTrue(
+            controlType.Contains("Table", StringComparison.OrdinalIgnoreCase) || localizedControlType.Contains("table", StringComparison.OrdinalIgnoreCase),
+            $"Expected the fields element to be exposed as a table/grid to UIA, but ControlType was \"{controlType}\" and LocalizedControlType was \"{localizedControlType}\".");
+    }
+
+    [TestMethod]
+    public async Task PopupViewerMedia_PrevNextButtonsHaveLocalizedNames()
+    {
+        OpenSample(PopupViewerFieldsPage);
+
+        var previousButton = FindElement("PreviousButton", DefaultTimeout);
+        var nextButton = FindElement("NextButton", DefaultTimeout);
+        Assert.IsFalse(string.IsNullOrWhiteSpace(GetAutomationName(previousButton)), "Expected the previous-media button to have a non-empty accessible name.");
+        Assert.IsFalse(string.IsNullOrWhiteSpace(GetAutomationName(nextButton)), "Expected the next-media button to have a non-empty accessible name.");
+    }
+
+    [TestMethod]
+    public async Task PopupViewerMedia_AnnouncesPositionInSet()
+    {
+        OpenSample(PopupViewerFieldsPage);
+
+        var nextButton = FindElement("NextButton", DefaultTimeout);
+        Click(nextButton);
+
+        var currentItemHost = FindElement("CurrentMediaView", DefaultTimeout);
+        var positionInSet = currentItemHost.GetAttribute("PositionInSet");
+        var sizeOfSet = currentItemHost.GetAttribute("SizeOfSet");
+        TestContext.WriteLine($"Media host PositionInSet=\"{positionInSet}\" SizeOfSet=\"{sizeOfSet}\"");
+        Assert.AreEqual("2", positionInSet, "Expected the second media item to report position 2 in the set after clicking Next.");
+        Assert.AreEqual("2", sizeOfSet, "Expected the media set size to be reported as 2.");
+    }
+
+    [TestMethod]
+    public async Task PopupViewerMedia_KeyboardNavigationWorks()
+    {
+        OpenSample(PopupViewerFieldsPage);
+
+        var previousButton = FindElement("PreviousButton", DefaultTimeout);
+        Click(previousButton); // Focuses the button and wraps around to the last (second) media item.
+
+        var currentItemHost = FindElement("CurrentMediaView", DefaultTimeout);
+        Assert.AreEqual("2", currentItemHost.GetAttribute("PositionInSet"), "Expected clicking Previous from the first item to wrap around to the last item.");
+
+        const int VK_LEFT = 0x25;
+        const int VK_RIGHT = 0x27;
+        PressKey(VK_LEFT);
+        Assert.AreEqual("1", currentItemHost.GetAttribute("PositionInSet"), "Expected the Left arrow key to navigate to the previous media item.");
+
+        PressKey(VK_RIGHT);
+        Assert.AreEqual("2", currentItemHost.GetAttribute("PositionInSet"), "Expected the Right arrow key to navigate to the next media item.");
+    }
+#endif
+
+    [TestMethod]
+    public async Task PopupViewerMedia_AccessibleDescriptionForDiagrams()
+    {
+        OpenSample(PopupViewerFieldsPage);
+
+        // The image media item is shown first and has an explicit AlternativeText.
+        Assert.IsTrue(ElementExistsByName(ImageMediaAlternativeText, DefaultTimeout), "Expected the image media's accessible name/description to be its AlternativeText.");
+
+#if WPF_TEST
+        // Advance to the chart media item, which has no AlternativeText, to exercise the Title/Caption fallback.
+        var nextButton = FindElement("NextButton", DefaultTimeout);
+        Click(nextButton);
+        Assert.IsTrue(
+            ElementExistsByName(ChartMediaTitle, DefaultTimeout) || ElementExistsByName(ChartMediaCaption, DefaultTimeout),
+            "Expected the chart media (which has no AlternativeText) to fall back to its Title or Caption for its accessible description.");
+#endif
+    }
+}
--- a/src/Tests/UITests/Toolkit.UITests.Shared/Tests/PopupViewer/PopupViewerTests.cs
+++ b/src/Tests/UITests/Toolkit.UITests.Shared/Tests/PopupViewer/PopupViewerTests.cs
@@ -19,7 +19,7 @@ public class PopupViewerTests : AppiumTestBase
     private const string ChartMediaCaption = "Visitor counts by month";
 
     [TestMethod]
-    public async Task PopupViewerText_FullTextIsExposed()
+    public async Task PopupViewer_Text_FullTextIsExposed()
     {
         OpenSample(PopupViewerFieldsPage);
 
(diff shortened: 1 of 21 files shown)


src/Toolkit/Toolkit/UI/Controls/PopupViewer/FieldsPopupElementView.cs
```diff
@@ -190,6 +220,15 @@ private ChildElement CreateTextCell(string? text, bool wrap)
                 t.TextWrapping = TextWrapping.Wrap;
 #endif
             }
+
+#if !WINUI
+            // Microsoft.UI.Xaml.Controls.TextBlock has no Focusable property (only Control-derived elements
+            // get IsTabStop in WinUI), so there's nothing to set here for the FieldsTableCellTextBlock wrapper.
+            t.Focusable = true;
```
gdestigter, line 227: What happens on MAUI here? Or will MAUI never get to this point?

src/Tests/UITests/Toolkit.UITests.Shared/Appium/AppiumTestBase.Utils.cs
```diff
@@ -103,6 +103,24 @@ protected AppiumElement FindElementByName(string name, TimeSpan? timeout = null)
...
+    protected AppiumElement FindElementByClassName(string className, TimeSpan? timeout = null)
+    {
+#if WINDOWS_TEST
+        var action = () => Driver.FindElement(MobileBy.ClassName(className));
+#else
+        throw new NotImplementedException("FindElementByClassName(string,TimeSpan?) is not implemented for this platform.");
+#endif
+        try
+        {
+            return OptionalWaitCall(action, timeout);
```
gdestigter, line 115: Does this compile in MAUI without the `action` variable?

src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.WinUI.cs
```diff
@@ -50,6 +52,10 @@ private void OnElementPropertyChanged()
             {
                 rtb.Content = HtmlToView.ToUIElement(Element.Text, (s, e) => PopupViewer.GetPopupViewerParent(s as DependencyObject)?.OnHyperlinkClicked(e));
                 rtb.Visibility = string.IsNullOrEmpty(Element?.Text) ? Visibility.Collapsed : Visibility.Visible;
+
+                // A ContentControl hosting an arbitrary element tree isn't a text control, so Narrator has
+                // nothing to read from it by default; the full text is set explicitly as its accessible name.
+                AutomationProperties.SetName(rtb, Element.Text.ToPlainText());
```
gdestigter, line 58: Does this give a nullability warning? Maybe we need something similar to the Visibility setting above?

gdestigter (commented):
Accessibility code looks good, just comments on potential build breaks.

src/Tests/UITests/Toolkit.UITests.Shared/Tests/PopupViewer/PopupViewerTests.cs
```diff
@@ -0,0 +1,114 @@
+namespace Toolkit.UITest.Shared.PopupViewer;
```
imalcolm1, line 1: When I initially made this issue, I was mostly thinking about checking functionality. Preeti added on the accessibility requirements later (which was needed), but we still probably want functionality tests like the following: 
- Check link are clickable (I believe this control has an event for this that you could subscribe to in the test pages so that you don't actually open a browser)
- Check text highlights properly as in the popup from the sample app (some screenshots [here](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Popup/Popup.md#textpopupelement)). This test probably doesn't need to be super precise, just a screenshot of the sentence/paragraph then check that some amount of blue and red is showing
- Check charts render properly. Same idea as the text highlighting, just do a screenshot and look for a certain amount of each of the expected chart colors. You might check the ratio of different colors on a bar chart to make sure it's accurate too
- Check the [edit summary](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Popup/Popup.md#edit-summary) text. I think Samhitha and Morten implemented that quite recently, the PR(s) shouldn't be too hard to find)
- Check that [attachments](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Popup/Popup.md#attachmentspopupelement) appear and can be downloaded
- Check that the [fields](https://devtopia.esri.com/runtime/common-toolkit/blob/main/designs/Popup/Popup.md#fieldspopupelement) display appears and has correct colors in light and dark mode
- Check that images show in a carousel (This one might be more trouble than it's worth because of needing to scroll. If it is too difficult, just checking that at least one image renders with a quick screenshot may be sufficient. And actually on second look, the `PopupViewer_Media_KeyboardNavigationWorks()` test here probably satisfies the checking the carousel works part, so just the screenshot check should be fine.) I think there's actually currently a bug on WPF where images aren't initially rendering, so this is a good check

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
```diff
@@ -28,10 +28,11 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="{x:Type primitives:TextPopupElementView}">
-                    <RichTextBox x:Name="TextArea" Visibility="{Binding Text, Converter={StaticResource PopupViewerVisibilityConverter}}" IsReadOnly="True" IsDocumentEnabled="True" BorderThickness="0" />
+                    <RichTextBox x:Name="TextArea" Visibility="{Binding Text, Converter={StaticResource PopupViewerVisibilityConverter}}" IsReadOnly="True" IsDocumentEnabled="True" BorderThickness="0" IsTabStop="False" />
```
imalcolm1: I'm not sure these tab stop changes make sense, don't users need to tab over the text blocks so they can use the narrator?

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
imalcolm1, line 168: The media elements and media titles+descriptions for this ControlTemplate should have tab stops. The elements themselves can act as links, so keyboard-only users need to be able to focus them to activate them

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
```diff
@@ -28,10 +28,11 @@
...
                 </ControlTemplate>
             </Setter.Value>
         </Setter>
+        <Setter Property="IsTabStop" Value="False"/>
```
imalcolm1, line 35: I'm not sure these tab stop changes make sense, don't users need to tab over the text blocks so they can use the narrator?

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
```diff
@@ -483,15 +487,15 @@
...
                                                     <primitives:PopupElementItemsControl.TextPopupElementTemplate>
                                                         <DataTemplate>
-                                                            <primitives:TextPopupElementView Element="{Binding}" Margin="0,10" />
+                                                            <primitives:TextPopupElementView Element="{Binding}" Margin="0,10" IsTabStop="False" Focusable="False"/>
                                                         </DataTemplate>
                                                     </primitives:PopupElementItemsControl.TextPopupElementTemplate>
                                                     <primitives:PopupElementItemsControl.MediaPopupElementTemplate>
                                                         <DataTemplate>
-                                                            <primitives:MediaPopupElementView Element="{Binding}" Margin="0,10" />
+                                                            <primitives:MediaPopupElementView Element="{Binding}" Margin="0,10" IsTabStop="False" />
```
imalcolm1, line 498: Again, not sure we want to get rid of tab stops here. These ones might not be so necessary as the individual text and media elements they contain, but they still might be useful for narrator users to know which section they're in

src/Toolkit/Toolkit/UI/Controls/PopupViewer/AccessibleTextBlock.cs
imalcolm1, line 1: Nice solution! I had to work around this in the WPF test app [here](https://github.com/Esri/arcgis-maps-sdk-dotnet-toolkit/blob/v.next/src/Tests/UITests/Toolkit.UITests.Wpf.App/ControlPatcher.cs) to get the scale line tests to work. That may have been a mistake since that workaround applies to *all* text blocks in the WPF test app, so we may get a false positives when checking that textblocks show up to the accessibility framework in WPF. I'll put up an issue to potentially change that.

Also, since it's strictly for WPF it might want to go in the same folder as the WPF PopupViewer.Theme file, or alternatively be renamed AccessibleTextBlock.WPF.cs or something.

imalcolm1 (commented):
I'm suspicious about the tab stops here on WPF, and I think WinUI might need some improvement in that regard too. Otherwise seems fine, though I haven't done completely exhaustive testing with narrator, etc

src/Tests/UITests/Toolkit.UITests.Shared/Tests/PopupViewer/PopupViewerTests.cs
imalcolm1: This is quite a lot, especially accounting for all platforms. It can go in a separate PR, and maybe even a separate issue as well so you can close this one out sooner

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
SamhithaBurugupalli: Ideally, all the non-interactive elements(like text) should not have a tabstop. It would make the navigation to an actionable control (like button, checkbox) extremely difficult. To answer the screen reader compatibility, assistive technology like windows narrator, NVDA have a scan mode option. So, using that scan mode, narrator will be able to announce it (if it is focusable)

src/Toolkit/Toolkit/UI/Controls/PopupViewer/FieldsPopupElementView.cs
OAzmitia: I have some Maui code that will fix this that I will push.

src/Toolkit/Toolkit/UI/Controls/PopupViewer/TextPopupElementView.WinUI.cs
OAzmitia: Since Element is a UI, Text for TextBox is never null. But either way, I will add the check.

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
OAzmitia: So based on the Narrator best practices, I found that tab should prioritize interactive controls such as buttons, textboxes, checkboxes, etc...
The Narrator has a "Screen Reader" function that instead of tab it can use ```Caps Lock + Left/Right``` and that works like a tab but only for controls that have some sort of Accessibility in it.

It is a good solution because regular users don't have to tab through read-only elements and users using accessibility can navigate those read-only elements.

imalcolm1: Ah makes sense then, hadn't known the scan mode was a thing, but it does seem to work well now that I try using it

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
imalcolm1: n/a given scan mode

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
imalcolm1: This still seems like it might be worth considering even with scan mode given the links

SamhithaBurugupalli (commented):
As discussed, I still see some additional tabstops at non interactive elements like( panes/ text etc) which needs to be fixed for it to be fully compliant. Also, there should a way to click and convey the info that the images are a hyperlink. Still continuing on my thorough review

SamhithaBurugupalli (approved):
Great work in bringing this to a very good shape. Looks really good overall. 
There are some minor issues, I don't want to block the merging of this PR with mostly working changes. Here is a list of issues with respect to a11y that I observed with WPF in different modes.

**Using only the keyboard:**

https://github.com/user-attachments/assets/edd0dece-18ba-4516-bce3-ced98cf33c8e

-	1st tab stop is reaching the rich text element (which is non interactive by itself) even if considering that is for hyperlink, I am still not able to click the hyperlink. Everything else is good here. 

**Using keyboard + narrator scan mode:**

https://github.com/user-attachments/assets/d14697e6-a1b3-4724-8f1a-c609f519f7b4



-	The reading of the rich text stops after 1st para. After that I am not able to reach the next lines with just the downward arrow key. When I press the upward arrow key, I get to the hyperlink “More info” which is weird. After that I had use “Ctrl+ downward arrow” to get to the next element.
-	There some controls with no automation name are getting the focus.
 Scenarios: 
1.	After the “An Image of Thunderbolt peak” is said, the downward arrow takes me to the image and then the narrator says text. 
2.	After the previous or next media button the focus goes into it’s content.
3.	After the “Chart for thunderbolt peak” the narrator again says text
4.	After elevation vs prominence it again says “text”
5.	Next it says just “table” – either an automation name should be said on this one or the view should be set to "Raw" the scan mode does not get to this.
6.	After reading out the table heading it says text again.
7.	At the end of the table it again says text. Based on the position of it, I think, it is the “Edit Summary” block. 

**Using Keyboard + narrator (non-scan mode):**

https://github.com/user-attachments/assets/302ed718-e4bc-47d4-8298-809b6a634326

-	Again the 1st tab stop reaches rich text even if its non-interactive by itself. (still not able to click the hyperlink here)
-	The image with hyperlink is just said as “image of white mountain peak”, Which should technically say that it is a hyperlink. (I am still fine with the current behavior as well, as the redirect url is always going to take you to the image itself) – So it’s your call, you can leave it as is if you want.
-	The hyperlinks in the table (example: “view” in this sample) are not announced by the narrator.

**Additional thoughts:** 
- Would it be useful to set [heading level](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.automationproperties.headinglevel?view=winrt-28000) property on the Titles of all the popup elements?
- Can we announce the alternate texts/ automation names of the images, when the user clicks on previous/ next buttons, so that the user would have a context on what changed

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
```diff
@@ -25,10 +26,11 @@
         <Setter Property="Foreground" Value="Gray" />
     </Style>
     <Style TargetType="{x:Type primitives:TextPopupElementView}">
+        <Setter Property="IsTabStop" Value="False" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="{x:Type primitives:TextPopupElementView}">
-                    <RichTextBox x:Name="TextArea" Visibility="{Binding Text, Converter={StaticResource PopupViewerVisibilityConverter}}" IsReadOnly="True" IsDocumentEnabled="True" BorderThickness="0" />
+                    <RichTextBox x:Name="TextArea" ScrollViewer.VerticalScrollBarVisibility="Auto" Visibility="{Binding Text, Converter={StaticResource PopupViewerVisibilityConverter}}" IsReadOnly="True" BorderThickness="0" Background="Transparent" IsTabStop="False" Focusable="False" />
```
gdestigter: I remember `IsDocumentEnabled` affecting embedded hyperlinks. Can you check whether removing `IsDocumentEnabled` makes hyperlinks unclickable?

src/Tests/UITests/Toolkit.UITests.TestPages.Shared/PopupViewer/PopupViewerFields.xaml.cs
```diff
@@ -0,0 +1,115 @@
...
+
+    public const string TextElementFirstParagraph = "This trailhead connects to the Ridgeline Loop and Summit Spur trails.";
+    public const string TextElementSecondParagraph = "Parking is available year-round, but the upper trail closes seasonally.";
+
+    // A minimal valid 1x1 PNG, embedded so image media renders with no network access.
+    private const string OnePixelPngDataUri = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII=";
+
+    public PopupViewerFields()
+    {
+        InitializeComponent();
```
gdestigter, line 41: What about MAUI? I assume that's coming, but does not having the xaml break the MAUI test app build?

gdestigter (commented):
Looks good, just a couple things to check before merging.

src/Toolkit/Toolkit.WPF/UI/Controls/PopupViewer/PopupViewer.Theme.xaml
OAzmitia: Correct. I overlooked that.

SamhithaBurugupalli:
Updated my previous comment on what's fixed for WPF. All the accessibility changes look good as well. There are some additional tab stops in WinUI and not able to navigate to some elements in scan mode. These WinUI fixes can go in a follow-up PR.

PR #865, opened 2026-09-15
[a11y] [ios] : Search view keyboard and screen reader compatibility

SamhithaBurugupalli:
From my observations, ios is acting a bit different than android and winui. It has it's own way of navigating. So, I couldn't find a better way to handle the change of focus (ex: like focusing on the 1st result when a suggestion like "school" is selected/ focusing the "All sources button" when the user clicks the drop down using keyboard). So, such enhancements will be missing for ios, but with these changes, I feel it will still be acceptable according to WCAG guidelines


imalcolm1 (commented):
Looks good, I can get everywhere eventually, but I'm not sure the grouping is fully working. It seems like the query bar is separate from all the other search elements for some reason and takes a lot of extra tapping around different page elements to reach it or get from it to the suggestions. It could be I'm missing something particular to iOS navigation though

SamhithaBurugupalli:
> Looks good, I can get everywhere eventually, but I'm not sure the grouping is fully working. It seems like the query bar is separate from all the other search elements for some reason and takes a lot of extra tapping around different page elements to reach it or get from it to the suggestions. It could be I'm missing something particular to iOS navigation though

One thing I noted while testing was, I am able to type in the query entry even if my keyboard focus is on an other control (this could be the reason that focus reaches to it at the last as the user is still able to type into it, without actually tabbing to it). It is because of this `isfirstresponder` property of ios. https://developer.apple.com/documentation/uikit/uiresponder/isfirstresponder . Also to get to the suggestions, Type something in the query entry and then press enter and then e downward arrow to get to the suggestion immediately. Again this navigation order will be different when using voice over. I guess it's mostly the ios way of handling things

imalcolm1 (approved):

gdestigter (approved):
Looks good :+1:

PR #866, opened 2026-09-15
[a11y]: search view maui high contrast theme changes

SamhithaBurugupalli:


src/Toolkit/Toolkit.Maui/SearchView/SearchView.Appearance.cs
```diff
@@ -91,14 +92,14 @@ static SearchView()
             titleLabel.SetBinding(Label.TextProperty, static (SearchSuggestion suggestion) => suggestion.DisplayTitle);
             titleLabel.VerticalOptions = LayoutOptions.End;
             titleLabel.VerticalTextAlignment = TextAlignment.End;
-            titleLabel.SetAppThemeColor(Label.TextColorProperty, Color.FromArgb(FOREGROUND_LIGHT), Color.FromArgb(FOREGROUND_DARK));
+            titleLabel.SetDynamicResource(Label.TextColorProperty, ForegroundColorResourceKey);
```
gdestigter, line 95: Does `SetDynamicResource` only handle resource changes, not the item’s hover or selected state? In high contrast, could these labels keep `WindowText` when `HighlightText` is expected on hover / selection?

src/Toolkit/Toolkit.Maui/SearchView/SearchView.cs
```diff
@@ -103,17 +108,39 @@ public SearchView()
...
+        OnSearchViewUnloaded();
+
+        if (Application.Current != null)
+        {
+            Application.Current.RequestedThemeChanged -= Application_RequestedThemeChanged;
+        }
+    }
+
+    private void Application_RequestedThemeChanged(object? sender, AppThemeChangedEventArgs e)
+        => UpdateThemeColors();
```
gdestigter, line 142: If high contrast is already on, could changing the app’s light/dark theme replace the high-contrast colors with normal colors? Should this check high contrast before updating them?

SamhithaBurugupalli: It is not possible to shift to light/ dark directly when in high contrast mode in windows. User has to disable the high contrast mode, which then sets the light/ dark mode

src/Toolkit/Toolkit.Maui/SearchView/SearchView.Appearance.cs
SamhithaBurugupalli: I tested this in the Windows high-contrast theme (even with customization of colors), including hover and selection states. The SearchView behavior matches the default WinUI control, and the text remains correctly readable using the system high-contrast colors. Since SetDynamicResource is only responsible for the normal foreground color and the native item container might be handling the hover/selection visuals, I don’t think a code change is needed here.


https://github.com/user-attachments/assets/a141be4e-4e52-4509-8cb2-1e01f3fda777

gdestigter (approved):
:+1:

OAzmitia (approved):

